PR #6, opened 2015-02-04
setting up for automatic publication on TR

dontcallmedom:
W3C is setting up a new [automated publication workflow](http://www.w3.org/2014/08/pubworkflow.html) for Working Drafts; I would like to use this document as one of the first few to be published that way, and this pull request brings some of the necessary metadata to that end.

For now, this doesn't mean that editor draft releases will be published automatically as TR Working Drafts; just that the next WD publication will be using the new workflow.

### DIFF
--- a/W3CTRMANIFEST
+++ b/W3CTRMANIFEST
@@ -1,0 +1,1 @@
+index.html?specStatus=WD respec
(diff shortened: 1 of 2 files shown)


vr000m:
LGTM, should I merge it?

dontcallmedom:
please :)

PR #14, opened 2015-09-30
Remove protocol from src link

KaptenJansson:
Chrome complains about mixed content when visiting this page over HTTPS, should always use protocol agnostic links.

PR #15, opened 2015-10-11
include respec via https

fippo:
https://w3c.github.io/webrtc-stats/ refuses the insecure include of respec.


fippo:
#14 beat me to this

PR #30, opened 2016-04-21
Fix .travis.yml YAML syntax

vivienlacourba:
Please also make sure this change gets integrated in the gh-pages branch as well when you will be publishing a new ED.


alvestrand:
Assuming that you know what you're doing. Merged.

PR #31, opened 2016-05-02

aboba:

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -538,6 +538,14 @@ stats [
           <dt>double availableIncomingBitrate</dt>
           <dd><p>Measured in Bits per second, and is implementation dependent.
             It may be calculated by the underlying congestion control.</p></dd>
+          <dt>double totalRoundTripTime</dt>
+          <dd><p>Represents the total round trip time in ms.</p></dd>
+          <dt>unsigned long long requests</dt>
+          <dd><p>Total number of initial consent requests sent.</p></dd>
+          <dt>unsigned long long responses</dt>
+          <dd><p>Total number of consent responses received.</p></dd>
+          <dt>unsigned long long retransmissions</dt>
+          <dd><p>Total number of consent request retransmissions.</p></dd>
         </dl>
         <div class="note">
             OPEN ISSUE: do the bitrate metrics need refs?
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -539,13 +539,17 @@ stats [
           <dd><p>Measured in Bits per second, and is implementation dependent.
             It may be calculated by the underlying congestion control.</p></dd>
           <dt>double totalRoundTripTime</dt>
-          <dd><p>Represents the total round trip time in ms.</p></dd>
-          <dt>unsigned long long requests</dt>
-          <dd><p>Total number of initial consent requests sent.</p></dd>
+          <dd><p>Represents the total round trip time in ms for connectivity checks.</p></dd>
           <dt>unsigned long long responses</dt>
-          <dd><p>Total number of consent responses received.</p></dd>
+          <dd><p>Total number of connectivity check responses received.</p></dd>
           <dt>unsigned long long retransmissions</dt>
-          <dd><p>Total number of consent request retransmissions.</p></dd>
+          <dd><p>Total number of connectivity check retransmissions.</p></dd>
+          <dt>double bindingTotalRoundTripTime</dt>
+          <dd><p>Represents the total round trip time in ms for binding request/responses.</p></dd>
+          <dt>unsigned long long bindingRequests</dt>
+          <dd><p>Total number of binding requests sent.</p></dd>
+          <dt>unsigned long long bindingResponses</dt>
+          <dd><p>Total number of binding responses received.</p></dd>
         </dl>
         <div class="note">
             OPEN ISSUE: do the bitrate metrics need refs?
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -539,17 +539,17 @@ stats [
           <dd><p>Measured in Bits per second, and is implementation dependent.
             It may be calculated by the underlying congestion control.</p></dd>
           <dt>double totalRoundTripTime</dt>
-          <dd><p>Represents the total round trip time in ms for connectivity checks.</p></dd>
+          <dd><p>Represents the total round trip time in ms (for non-retransmitted connectivity checks only).</p></dd>
           <dt>unsigned long long responses</dt>
           <dd><p>Total number of connectivity check responses received.</p></dd>
           <dt>unsigned long long retransmissions</dt>
           <dd><p>Total number of connectivity check retransmissions.</p></dd>
-          <dt>double bindingTotalRoundTripTime</dt>
-          <dd><p>Represents the total round trip time in ms for binding request/responses.</p></dd>
-          <dt>unsigned long long bindingRequests</dt>
-          <dd><p>Total number of binding requests sent.</p></dd>
-          <dt>unsigned long long bindingResponses</dt>
-          <dd><p>Total number of binding responses received.</p></dd>
+          <dt>double consentTotalRoundTripTime</dt>
+          <dd><p>Represents the total round trip time in ms for consent request/responses.</p></dd>
+          <dt>unsigned long long consentRequests</dt>
+          <dd><p>Total number of consent requests sent.</p></dd>
+          <dt>unsigned long long consentResponses</dt>
+          <dd><p>Total number of consent responses received.</p></dd>
         </dl>
         <div class="note">
             OPEN ISSUE: do the bitrate metrics need refs?
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -540,18 +540,26 @@ stats [
             It may be calculated by the underlying congestion control.</p></dd>
           <dt>double totalRoundTripTime</dt>
           <dd><p>Represents the total round trip time in ms (for non-retransmitted connectivity checks only).</p></dd>
-          <dt>unsigned long long requests</dt>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -538,6 +538,18 @@
           <dt>double availableIncomingBitrate</dt>
           <dd><p>Measured in Bits per second, and is implementation dependent.
             It may be calculated by the underlying congestion control.</p></dd>
+          <dt>double totalRoundTripTime</dt>
+          <dd><p>Represents the total round trip time in ms (for non-retransmitted connectivity checks only).</p></dd>
+          <dt>unsigned long long responses</dt>
+          <dd><p>Total number of connectivity check responses received.</p></dd>
+          <dt>unsigned long long retransmissions</dt>
```
vr000m: Does this number exclude the first request?

aboba: Yes.  So requests = retransmissions + 1

webrtc-stats.html
```diff
@@ -538,6 +538,18 @@
           <dt>double availableIncomingBitrate</dt>
           <dd><p>Measured in Bits per second, and is implementation dependent.
             It may be calculated by the underlying congestion control.</p></dd>
+          <dt>double totalRoundTripTime</dt>
+          <dd><p>Represents the total round trip time in ms (for non-retransmitted connectivity checks only).</p></dd>
+          <dt>unsigned long long responses</dt>
+          <dd><p>Total number of connectivity check responses received.</p></dd>
+          <dt>unsigned long long retransmissions</dt>
+          <dd><p>Total number of connectivity check retransmissions (does not include initial check).</p></dd>
+          <dt>double consentTotalRoundTripTime</dt>
```
juberti: How would one distinguish consent checks from normal checks, and why is this useful?

webrtc-stats.html
juberti: Is that always the case? What about aggressive nomination?

juberti: Seems like we should have requests, responses, and retransmissions - and just avoid the consent distinction altogether, unless there's something specific to consent that we think apps need to know about.

robin-raymond:
@juberti I think the ice connectivity vs consent stats being combined could be misleading. For example, if the connectivity check dropped 5 packets prior to the remote firewall pinhole opening and then you sent 3 successful consent checks, you would see 4 success packets 5 lost packets; But that is misleading because all the failures happened during the set-up phase where dropped packets are likely and no failures occurred after that point. You'd not be able to tell by just looking at the stats if you have an intermittent problem vs a normal condition.

webrtc-stats.html
aboba: The transport behavior of consent checks (https://tools.ietf.org/html/rfc7675) is different from connectivity checks. Consent checks are not re-transmitted (transaction id is changed each time), while connectivity checks re-transmit (with backoff).  Also, you can have "suicide" connectivity checks.

juberti:
That's fair. But limiting these post-ICE pings to consent checks seems limiting. One might send binding requests for a variety of reasons, besides consent, and we'd want to be able to count those properly.

Perhaps this is just a naming issue.

webrtc-stats.html
juberti: That makes sense (see below) - but I'm not sure that requests = retransmissions + 1, e.g. aggressive nomination.

aboba:
@juberti   For post-ICE pings, question is whether they use the RFC 7675 transport profile (new transaction id for each request, no re-transmissions).  If so, they can use the 'consent' counters.  Do you have a suggestion for a better name than 'consent'?

Added connectivity check request counter.

juberti:
They will, necessarily, use the new transaction id so that RTTs are computed correctly. However, they may be sent over pairs other than the selected pair, so they are not consent checks.

Will think regarding the name question - it may well be the case that 'consent' is least ambiguous.

webrtc-stats.html
```diff
@@ -538,6 +538,20 @@
           <dt>double availableIncomingBitrate</dt>
           <dd><p>Measured in Bits per second, and is implementation dependent.
             It may be calculated by the underlying congestion control.</p></dd>
+          <dt>double totalRoundTripTime</dt>
+          <dd><p>Represents the total round trip time in ms (for non-retransmitted connectivity checks only).</p></dd>
```
vr000m: Isn't the totalRTT the same the RTT defined above (in STUN-PATH-CHAR)?

pthatcherg:
I have a few issues with the current state of this PR:

1.  I agree with Justin that we should just have one kind of check, not separate connectivity and consent checks.  Checks may be sent for both purposes, so which bucket does it go in?  And I don't really see much of a use for the stats being separate.  What useful information does that add?

2.  I like being able to have "requests sent" and "requests received" (and "responses sent" and "responses received").  I've heard it argued that such is redundant, but that's only true if you control both sides of the call.  If you only control one side, then having the info for requests received and responses sent is valuable.

3.  If we use a separate transaction ID for each connectivity check, what should the retransmission count be?  0?

robin-raymond:
@pthatcherg because firewall pinholes are closed, initial set of checks are likely lost (and therefor appear to show failures); however, after pinholes are open, checks failing are indicative of true loss; mixing the stats skews the ability to determine if packets lost were because of normal pinhole closure vs true loss.

aboba:
@pthatcherg  @juberti  I agree that it would be desirable to have a single set of counters, but the different transport behavior of connectivity checks and post-connectivity checks makes it difficult to use the same counters for both. 

Since you can't obtain an RTT observation from a re-transmitted connectivity check, a single connectivity check transaction will only provide a single RTT value (if successful on the first try) or none (if it fails or requires re-transmissions to succeed).  So not clear to me how useful connectivity check statistics really are. 

However, since RFC 7675 uses a distinct transaction id for each request, an RTT value is obtained from each response post-connectivity, and therefore post-connectivity check counters should provide some useful data.

aboba:
@pthatcherg  For post-connectivity checks done according to RFC 7675, the retransmission count is always 0.

pthatcherg:
It's the concept of "pre-connectivity" and "post-connectivity" checks that I have a problem with.   "Connected" is not binary, and if it were, it would not have a single event in time when it would flip the bit.  It's much more complex than that.  The ICE transport can have many candidate pairs, each with their own connectivity state, and be switching between them.  And even a given candidate pair may be in a "weakly connected" state where it was connected, but now it starts sending connectivity checks not just for consent, but to get more "strongly connected".

In such a dynamic world, how can you determine what is "pre-connectivity" and what is "post-connectivity"?   I don't think you can. 

And I don't see how this can help know why packets are lost (because of pinholes being open or not).  There's no way to know why a check is lost.  You can't know  if it's because of lack of pinhole or not.  You just know if it got there or not.  I don't think splitting up the stats is going to help anyone know why a check might have been lost.  

As for RTTs, we can use separate transaction IDs for each ICE check, regardless of "pre-connectivity" vs "post-connectivity".  Then RTT values are always useful.

aboba:
@pthatcherg  The distinction between "pre-connectivity" and "post-connectivity" checks was introduced by RFC 7675, which prescribed different transport behavior for "consent" checks.  I didn't understand why this was necessary at the time and I still don't:
https://www.ietf.org/mail-archive/web/rtcweb/current/msg14631.html  

If the transport behavior is rationalized (e.g. by using separate transaction IDs for each ICE check) then it would be easier to have a single set of counters.  However, the connectivity checks would still need backoff (unlike consent) so some transport differences probably would remain.

pthatcherg:
When I read RFC 7675, I don't see anything about "pre-connectivity" and
"post-connectivity".  I only see pre-consent and post-consent, which is not
exactly the same thing.  Or am I missing it?

If we're modeling this as "pre-consent" and "post-consent", then I can see
this making sense.  But then the names of and descriptions of the stats
should make that much more clear.  For example, by calling them
consentRequestsSent/consentResponsesReceived and
consentRefreshRequestsSent/consentRefreshResponsesReceived.   "connectivity
check" vs "consent check" makes it sound like a check can't be both.  But
"consent request" vs "consent refresh request" makes it clear one is before
having consent and the other is after.

But even if it makes sense and it is easy to implement, is it useful?
Well, I do think it would be useful to know, for a given candidate pair,
how many checks we sent before getting a first response.  That would could
be nice to know.


Ok, I guess I've convinced myself that having separate stats are useful, as
long as it's named correctly with clear semantics (request consent vs.
refresh consent).

vr000m:
Help me understand why we need 3 RTT metrics, I am not sure if we need all three RTT metrics.
+ double roundTripTime
+ double totalRoundTripTime
+ double consentTotalRoundTripTime

We could live with two, one during connectivity checks and the other after connectivity is established. To clarify, I am mainly curious to know, if the RTT does in fact vary wildly when connectivity checks take place compared to when connectivity is established.

vr000m:
About the pre-connectivity and post-connectivity, I'd have just cached the ICE request and response counters at the point of connectivity establishment. And just compared to that cached object later on in the session. But if the connectivity on a particular candidate pair would often switch from connected to disconnected and back to connected, then caching would be a lot of work. 

In mobile networks, I see value in having the sent and receive counters for consentRequest and consentRefresh metrics.

pthatcherg:
On Wed, May 11, 2016 at 3:23 AM, Varun Singh <notifications@github.com>
wrote:

> About the pre-connectivity and post-connectivity, I'd have just cached the
> ICE request and response counters at the point of connectivity
> establishment.
>
​Ah, but you don't know when a particular candidate pair gets connectivity,
only when the first candidate pair does.  Unless, of course, we add the
connectivity/​consent state of a given candidate pair to the stats of the
candidate pair.  And even then you might miss something because you're only
polling so frequently.

 And just compared to that cached object later on in the session. But if
the connectivity on a particular candidate pair would often switch from
connected to disconnected and back to connected, then caching would be a
lot of work.

PR #32, opened 2016-05-03
Aligned RTCIceCandidate with webrtc-pc

### LINKED ISSUE #12: serverUrl for RTCIceCandidate?
@alvestrand just noticed that serverUrl as described in https://lists.w3.org/Archives/Public/public-webrtc/2014May/0115.html is not part of the spec.

It's still more useful than just being able to figure out the type of the connection from the local type preference / priority.

### LINKED ISSUE #21: RTCIceCandidateAttributeStats needs sync with RTCPeerConnection object
The spec needs to say (in text) that the AttributeStats is representing everythig in the non-Stats object + some more, and have the field names & types be in sync.

alvestrand:
Names and definitions changed to conform with the corresponding
webrtc-pc type.

Fixes #12 
Fixes #21 

Tagging @vr000m @juberti for review

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -450,47 +450,36 @@ stats [
       </section>
 
       <section id="icecandidate-dict*">
-        <h3>RTCIceCandidateAttributes dictionary</h3>
+        <h3>RTCIceCandidateStats dictionary</h3>
         <p>
-          <code>RTCIceCandidateAttributes</code> reflects the properties of a
-          <code>candidate</code> in Section 15.1 of [[!RFC5245]].
+          <code>RTCIceCandidateStats</code> reflects the properties of a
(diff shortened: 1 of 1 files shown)


alvestrand:
Two sources of alignment: webrtc-pc and 5245. They don't align with each other.

vr000m:
I noted: 

+ ip used to be ipAddress, in 5245 it is: `connection-address`. Although, RFC5245 allows for an FQDN, which seems is not allowed.

+ port used to portNumber. I am okay with this as it aligns with the ABNF in 5245.

+ protocols used to be transport. In 5245 this is called the transport.

alvestrand:
replaced by PR #36 (this was easier than rebasing).

PR #33, opened 2016-05-03
Adds the "implementation" variable to RTCCodec.

### LINKED ISSUE #16: Codec should have implementation info
Called out by @pbos: In certain contexts, it's important to know which particular encoder is being used by a particular platform (for instance when both hardware and software codecs are available, and one or the other may be in use at any particular moment).

This is really only useful when one has intimate knowledge of the platform in use, so an implementation-defined string seems natural.

Suggestion: Add a new attribute to the RTCCodec dictionary.
Name: implementationName
Value: DOMString
Description: An implementation-specific identifier for the codec in use. This will be the encoder for outgoing streams, and the decoder for incoming streams; they may be different.

alvestrand:
Also pulls in the "fingerprint" implementation.

Fixes #16


vr000m:
LGTM. Closes #16

PR #34, opened 2016-05-13
Converted to WebIDL contiguous mode

dontcallmedom:
See https://lists.w3.org/Archives/Public/public-webrtc/2016May/0073.html for context


vr000m:
If there is nothing blocking this, I would suggest merging this.

alvestrand:
It's a whole-file replacement. Our outstanding CLs will need rebase.

PR #36, opened 2016-05-23
New update of alignment of RTCIceCandidate

### LINKED ISSUE #12: serverUrl for RTCIceCandidate?

### LINKED ISSUE #21: RTCIceCandidateAttributeStats needs sync with RTCPeerConnection object

alvestrand:
Fixes #12
Fixes #21

Replaces PR #32

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -100,8 +100,8 @@
       <section id="rtcstatstype-str*">
         <h3>
           RTCStatsType DOMString
-        </h3><!-- 
-        As per discussions: Move enum to a string that is in a registry 
+        </h3><!--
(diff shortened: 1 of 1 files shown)


vr000m:
LGTM.

vr000m:
LGTM.

PR #38, opened 2016-05-23
adding bitrate definitions to the spec

### LINKED ISSUE #13: Bitrate should have a definition
http://w3c.github.io/webrtc-stats/#dictionary-rtcoutboundrtpstreamstats-members has "targetBitrate" without a definition.

This needs to call on some definition - preferably a preexisting one. At least we should be able to tell whether it's with or without RTP headers, and it would also be nice to say what measurement interval is expected to produce this bitrate. Not sure one exists.

vr000m:
Using the definition of bitrate from RFC3890. Fixes #13.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1337,7 +1337,7 @@ stats [
               </dt>
               <dd>
                 <p>
-                  t is calculated by the underlying congestion control by combining the available
+                  It is calculated by the underlying congestion control by combining the available
                   bitrate for all the incoming media streams. The bitrate measurement does not
                   count the size of the IP or other transport layers like TCP or UDP. It is
                   similar to the TIAS defined in [[!RFC3890]], i.e., it is measured in bits per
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1330,16 +1334,15 @@
               </dt>
               <dd>
                 <p>
-                  Measured in Bits per second, and is implementation dependent. It may be
-                  calculated by the underlying congestion control.
+                  Measured in bits per second, and is implementation dependent. It may be
+                  calculated by the underlying congestion control by combining the available
+                  bitrate for all the incoming media. The bitrate is calculated over a
+                  1 second window [[!RFC3890]].
```
alvestrand: Add explicit TIAS reference (so that nobody thinks it's AS)

vr000m:
@alvestrand PTAL. added reference to TIAS. feedback appreciated.

alvestrand:
LGTM - nervous about all the "it is similar to" rather than "it is defined as equal to", but let's go with this.

Issue #28, opened 2016-03-18
Broken links and fragments


vivienlacourba:
Result of: https://validator.w3.org/checklink?url=https://w3c.github.io/webrtc-stats/webrtc-stats.html

```
 Line: 186 https://w3c.github.io/webrtc-stats/webrtc-stats.html
    Status: 200 OK

    Some of the links to this resource point to broken URI fragments (such as index.html#fragment). 
    Broken fragments:

        https://w3c.github.io/webrtc-stats/webrtc-stats.html#widl-RTCCodec-codec (line 186)

warning Lines: 70, 77, 80, 84 http://dev.w3.org/html5/spec/webappapis.html redirected to http://w3c.github.io/html/webappapis.html
    Status: 301 -> 200 OK

    This is a permanent redirect. The link should be updated to point to the more recent URI. 
    Broken fragments:

        http://dev.w3.org/html5/spec/webappapis.html#event-handlers (line 84)
        http://dev.w3.org/html5/spec/webappapis.html#fire-a-simple-event (line 80)
        http://dev.w3.org/html5/spec/webappapis.html#queue-a-task (line 77)
        http://dev.w3.org/html5/spec/webappapis.html#eventhandler (line 70)
```

vivienlacourba:
http://dev.w3.org/html5/spec/webappapis.html is now a redirect to https://w3c.github.io/html/webappapis.html which is section "7. Web application APIs" of the **HTML 5.1** Editor's Draft.

I believe our dependency is on **HTML5** not **HTML 5.1**. If true then we should link to the HTML5 REC https://www.w3.org/TR/2014/REC-html5-20141028/webappapis.html

Also [`Eventhandler`](https://www.w3.org/TR/2014/REC-html5-20141028/webappapis.html#eventhandler) is a callback function not an interface.

vr000m:
I think https://w3c.github.io/webrtc-stats/webrtc-stats.html#widl-RTCCodec-codec works now. Thanks to @dontcallmedom's update in #34

alvestrand:
@vivienlacourba can you look again and verify that it's fixed? Closing issue for now.

PR #42, opened 2016-05-23
updating HTML5 URIs and EventHandler definition

vr000m:
Fixes the validator warnings posted by @vivienlacourba in #28


alvestrand:
LGTM

PR #43, opened 2016-05-23
Added procedures for new stats

### LINKED ISSUE #25: Extension mechanism needs to be clear
For WebRTC-1.0, stats needs to be a stable reference (with all the stats that are in the MTI set).

Adding stats needs to be a process that's clear to people who want to do it.

alvestrand:
The approach sketched is "living document".

Fixes #25

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -200,14 +200,27 @@
       </p>
       <p>
         If a need for a new stats dictionary or dictionary member is found,
-        an issue should be raised against this spec on Github, and the W3C
-        consensus process will decide on whether the stat should be added
+        an issue should be raised against this spec on Github, and a
+        review process will decide on whether the stat should be added
         to the specification or not.
       </p>
       <p>
         A pull request for a change to this document may serve as guidance
-        for the discussion, but the eventual merge is dependent on the usual
-        W3C consensus process.
+        for the discussion, but the eventual merge is dependent on the
+        review process.
+      </p>
+      <p>
+        While the WebRTC WG exist, it will serve as the review body;
+        once it has disbanded, the W3C will have to establish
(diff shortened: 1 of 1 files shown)


alvestrand:
@fluffy @dontcallmedom @juberti please comment on approach

webrtc-stats.html
```diff
@@ -88,6 +88,28 @@
...
     </section><!-- VS: getStats() remains in base spec! -->
+    <section>
+      <h2>Maintenance procedures for stats types</h2>
+      <p>This document, in its editors' draft form, serves as the repository
+        for the currenly defined set of stats types.
+      </p>
+      <p>
+        If a need for a new stats dictionary or dictionary member is found,
+        an issue should be raised against this spec on Github, and the W3C
+        consensus process will decide on whether the stat should be added
```
fluffy: What is the W3C consensus process for a closed WG ?

fluffy:
Needless to say, I would prefer a process where an someone could get a stat name and start implementing that in parallel with it becoming a standards and know that there would be no name conflicts.

vr000m:
The process for getting a stat name for a well documented metric should indeed be documented.

alvestrand:
The experience I have with adding stats so far is that there are two controversial things:
- The name
- The definition (usually the units, sometimes how the value is computed)
Declaring either to be frozen by the time we get a proposal seems not likely to be acceptable.

vr000m:
The bar is set as: Expert review, similar to IANA process

+ citable reference
+ a good name

there was more content on how the actual process might work. need input from @fluffy

alvestrand:
Current thought: Add "bar is at Expert Review" and "after WG close, an expert will have to be appointed", and merge.

Issue #37, opened 2016-05-23
RTT metrics in milliseconds or seconds?


vr000m:
currently the spec is not clear on what the unit for all RTT measurements are. In one place it says (seconds), and for the other no units are presented. 

Should the RTT metrics be reported in milliseconds or seconds. Since the type is double, I feel the milliseconds for the integer part may be the right resolution.

jesup:
I vote for (and have assumed) ms

jan-ivar:
> Since the type is double, I feel the milliseconds for the integer part may be the right resolution. 

That seems counter-intuitive to me. Had you said "Since the type is integer", I would have bought ms.

But with floating point, why not just assume the base unit?

FWIW we've implemented `jitter` and `mozRtt` since Firefox 30 (both in seconds).

jan-ivar:
> currently the spec is not clear on what the unit for all RTT measurements are. In one place it says (seconds), and for the other no units are presented. 

@vr000m The place where it says "seconds" was committed by you on [Aug 29, 2014](https://github.com/w3c/webrtc-stats/commit/7c40ead6a6e0f2fe75092def27dc75418c5ef339).

The place with no mention of unit was committed by you on [Feb 2, 2015](https://github.com/w3c/webrtc-stats/commit/212576fdae0fc027d59c4d8d5487404a54811a0b).

So for at least 6 months it was consistent. ;-)

jesup:
I'll revise -- with floating point, seconds makes sense.

vr000m:
@jan-ivar, my implementation of STUN-PATH-CHAR draft uses milliseconds, although the spec itself does not say what it should be measured in. 

If we have converged on seconds, I'll fix my PR to seconds and move ahead.

alvestrand:
PR #44 (wrong title) will set it to seconds.

PR #44, opened 2016-05-23
setting units to seconds

vr000m:
possible way forward on issue #37.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -100,8 +100,8 @@
       <section id="rtcstatstype-str*">
         <h3>
           RTCStatsType DOMString
-        </h3><!-- 
-        As per discussions: Move enum to a string that is in a registry 
+        </h3><!--
+        As per discussions: Move enum to a string that is in a registry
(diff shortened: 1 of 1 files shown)


alvestrand:
@dontcallmedom @jan-ivar @adam-be do we have a preference for milliseconds vs seconds?
I think we're clear that we prefer double as the datatype.

jan-ivar:
I thought we had previously standardized on the base unit, because double. Why change now?

Would this affect [jitter](http://w3c.github.io/webrtc-stats/#dom-rtcinboundrtpstreamstats-jitter) as well?

We've implemented `jitter` since Firefox 30 (`mozRtt` is still prefixed).

vr000m:
I will change the PR to consistently use seconds as units.

alvestrand:
Using seconds throughout seems good to me.

vr000m:
@alvestrand this is ready to go. PTAL

alvestrand:
LGTM

PR #45, opened 2016-05-23

aboba:

Rebase of PR https://github.com/w3c/webrtc-stats/pull/31

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1223,7 +1223,6 @@ stats [
              unsigned long long            responsesReceived;
              unsigned long long            responsesSent;
              unsigned long long            retransmissionsSent;
-             double                        consentTotalRoundTripTime;
              unsigned long long            consentRequestsReceived;
              unsigned long long            consentRequestsSent;
              unsigned long long            consentResponsesReceived;
@@ -1340,7 +1339,8 @@ stats [
               </dt>
               <dd>
                 <p>
-                  Represents the RTT computed by the STUN connectivity checks [[!STUN-PATH-CHAR]].
+                  Represents the RTT computed by the STUN connectivity checks
+                  [[!STUN-PATH-CHAR]] as well as consent checks [[!RFC7675]].
                 </p>
               </dd>
               <dt>
@@ -1406,15 +1406,6 @@ stats [
               <dd>
                 <p>
                   Represents the total number of connectivity check retransmissions sent.
-                </p>
-              </dd>
-              <dt>
-                <dfn><code>consentRoundTripTime</code></dfn> of type <span class=
-                "idlMemberType"><a>double</a></span>
-              </dt>
-              <dd>
-                <p>
-                  Represents the round trip time in ms for consent requests/responses [[!RFC7675]].
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1218,6 +1218,17 @@ stats [
              double                        roundTripTime;
              double                        availableOutgoingBitrate;
              double                        availableIncomingBitrate;
+             double                        totalRoundTripTime;
+             unsigned long long            requestsReceived;
+             unsigned long long            requestsSent;
+             unsigned long long            responsesReceived;
+             unsigned long long            responsesSent;
+             unsigned long long            retransmissionsSent;
+             double                        consentTotalRoundTripTime;
+             unsigned long long            consentRequestsReceived;
+             unsigned long long            consentRequestsSent;
+             unsigned long long            consentResponsesReceived;
+             unsigned long long            consentResponsesSent;
 };</pre>
           <section>
             <h2>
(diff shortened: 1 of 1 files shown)


aboba:
@vr000m We only need 2 RTT metrics (pre-consent and post-consent). The PR has been changed to reflect that.

@pthatcherg Yes, the distinction is "pre-consent" and "post-consent".  This determines whether the transport behavior of RFC 5389 Section 7.2.1 (STUN connectivity check) or RFC 7675 section 5.1 (consent) is used.

webrtc-stats.html
```diff
@@ -1218,6 +1218,16 @@
...
+             unsigned long long            requestsSent;
+             unsigned long long            responsesReceived;
+             unsigned long long            responsesSent;
+             unsigned long long            retransmissionsSent;
+             double                        consentTotalRoundTripTime;
+             unsigned long long            consentRequestsReceived;
+             unsigned long long            consentRequestsSent;
+             unsigned long long            consentResponsesReceived;
+             unsigned long long            consentResponsesSent;
 };</pre>
```
pthatcherg: Are consentRequests the ones before or after consent is received?  I hope before, but it should be clear.

webrtc-stats.html
```diff
@@ -1218,6 +1218,16 @@
              double                        roundTripTime;
              double                        availableOutgoingBitrate;
              double                        availableIncomingBitrate;
+             unsigned long long            requestsReceived;
+             unsigned long long            requestsSent;
+             unsigned long long            responsesReceived;
+             unsigned long long            responsesSent;
```
pthatcherg: Are these the checks sent before or after consent is received?  It sounds like it's the total of both before and after, which I think makes a lot of sense.  But it's not clear.

webrtc-stats.html
```diff
@@ -1218,6 +1218,16 @@
...
+             unsigned long long            retransmissionsSent;
+             double                        consentTotalRoundTripTime;
+             unsigned long long            consentRequestsReceived;
+             unsigned long long            consentRequestsSent;
+             unsigned long long            consentResponsesReceived;
+             unsigned long long            consentResponsesSent;
```
pthatcherg: Are these the checks before or after consent is received?  It sounds like it's for the checks before, but it needs to be more clear.

webrtc-stats.html
```diff
@@ -1218,6 +1218,16 @@
...
              double                        availableIncomingBitrate;
+             unsigned long long            requestsReceived;
+             unsigned long long            requestsSent;
+             unsigned long long            responsesReceived;
+             unsigned long long            responsesSent;
+             unsigned long long            retransmissionsSent;
+             double                        consentTotalRoundTripTime;
+             unsigned long long            consentRequestsReceived;
+             unsigned long long            consentRequestsSent;
+             unsigned long long            consentResponsesReceived;
```
pthatcherg: If this is consent responses received before consent is received, won't this number always be zero?  You never receive any consent responses before you receive consent responses, by definition.

webrtc-stats.html
```diff
@@ -1218,6 +1218,16 @@
...
              double                        availableOutgoingBitrate;
              double                        availableIncomingBitrate;
+             unsigned long long            requestsReceived;
+             unsigned long long            requestsSent;
+             unsigned long long            responsesReceived;
+             unsigned long long            responsesSent;
+             unsigned long long            retransmissionsSent;
+             double                        consentTotalRoundTripTime;
+             unsigned long long            consentRequestsReceived;
+             unsigned long long            consentRequestsSent;
```
pthatcherg: Are these the checks before or after consent is received? It sounds like it's for the checks before, but it needs to be more clear.

webrtc-stats.html
pthatcherg: Is this before or after consent is received?  Because I don't think it really matters how many consent requests are received and responses sent before we receive consent.  I think we could just remove those two because total requestsReceived and responsesSent is enough.

webrtc-stats.html
```diff
@@ -1218,6 +1218,16 @@
              double                        roundTripTime;
              double                        availableOutgoingBitrate;
              double                        availableIncomingBitrate;
+             unsigned long long            requestsReceived;
+             unsigned long long            requestsSent;
+             unsigned long long            responsesReceived;
+             unsigned long long            responsesSent;
+             unsigned long long            retransmissionsSent;
+             double                        consentTotalRoundTripTime;
```
pthatcherg: If this is for the roundtrip before we receive consent, isn't there by definition one RTT sample (since once we receive one response, we have consent)?  How useful is that really?  I think we should just have one RTT value, not pre- and post-consent values.

webrtc-stats.html
```diff
@@ -1218,6 +1218,16 @@
              double                        roundTripTime;
              double                        availableOutgoingBitrate;
              double                        availableIncomingBitrate;
+             unsigned long long            requestsReceived;
+             unsigned long long            requestsSent;
+             unsigned long long            responsesReceived;
+             unsigned long long            responsesSent;
+             unsigned long long            retransmissionsSent;
```
pthatcherg: If an implementation never uses retransmissions (and just schedules another ICE check at the next time one should be scheduled), then will this always be zero?  That would be the case with our ICE implementation.

webrtc-stats.html
aboba: The intent was for requestsReceived/Sent and responsesReceived/Sent to represent checks sent before consent (when retransmission can occur).

webrtc-stats.html
aboba: consentRequestsReceived/RequestsSent, etc. represent counts after initial consent is given.  Looking at this, it might make more sense to hvae consentRoundTripTime instead of consentTotalRoundTripTime.

webrtc-stats.html
aboba: If the implementation doesn't retransmit, then retransmissionsSent will always be zero.  For such an implementaiton, the pre and post-consent transport behavior is the same.

webrtc-stats.html
aboba: Pre-consent there can only be a single RTT sample, so might as well use all data in the roundTripTime estimate.

webrtc-stats.html
```diff
@@ -1340,7 +1339,8 @@
               </dt>
               <dd>
                 <p>
-                  Represents the RTT computed by the STUN connectivity checks [[!STUN-PATH-CHAR]].
+                  Represents the RTT computed by the STUN connectivity checks
+                  [[!STUN-PATH-CHAR]] as well as consent checks [[!RFC7675]].
```
vr000m: Is this an average, i.e., sum of the RTTs computed divided by the number of checks (connectivity and consent)?

aboba: Yes.

vr000m: I wanted to confirm that it's the intention. Because this value will be different from the value exposed by the RTP stack (the combination of RTCP SR and RR). Even though the media maybe flowing on that particular candidate pair.

alvestrand:
LGTM - note: there will be a merge conflict with #44.

alvestrand:
@aboba can you respond to the open comments and rebase? (we generated a merge conflict)

Issue #50, opened 2016-05-31
Dashes needed in webrtc-stats enum


alvestrand:
The enum of stats types needs to follow W3C conventions as stated in #5 and use dashes.

Issue #48, opened 2016-05-26
Names of dictionaries need to explicitly link to the type they represent.


alvestrand:
For instance, "transport" needs to say that it is represented by an RTCTransportStats object (and hyperlinked).
Inboundrtp and outboundrtp have it already.

PR #51, opened 2016-06-07
added dashed enums and crosslinking to stats objects

vr000m:
related to #50 and #48.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -114,7 +114,7 @@
         </p>
         <dl>
           <dt>
-            <code>"inboundrtp"</code>
+            <code>"inbound-rtp"</code>
           </dt>
           <dd>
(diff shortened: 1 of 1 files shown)


alvestrand:
LGTM

webrtc-stats.html
```diff
@@ -142,11 +142,12 @@
             </p>
```
alvestrand: If session exists, it should point to RTCPeerConnectionStats. But should we rename it to peer-connection?

vr000m: Will fix this.

vr000m:
@alvestrand PTAL, changes the "session" key to "peer-connection", makes it consistent with "data-channel".

alvestrand:
LGTM!

PR #53, opened 2016-06-08
Clean up table of content

alvestrand:
This PR balances a number of unbalanced &lt;section> directives, and moves
one example section.


vr000m:
LGTM.

PR #54, opened 2016-06-09

aboba:

Rebase of PR https://github.com/w3c/webrtc-stats/pull/31

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1267,7 +1267,7 @@
               </dt>
               <dd>
                 <p>
-                  Represents the RTT measured in seconds, computed from both STUN
+                  Represents the average RTT measured in seconds, computed from both STUN
                   connectivity checks [[!STUN-PATH-CHAR]] and consent responses [[!RFC7675]].
                 </p>
               </dd>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1150,6 +1150,7 @@
              unsigned long long            requestsSent;
              unsigned long long            responsesReceived;
              unsigned long long            responsesSent;
+             unsigned long long            retransmissionsReceived;
              unsigned long long            retransmissionsSent;
              unsigned long long            consentRequestsReceived;
              unsigned long long            consentRequestsSent;
@@ -1331,6 +1332,15 @@
               <dd>
                 <p>
                   Represents the total number of connectivity check responses sent.
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>retransmissionsReceived</code></dfn> of type <span class=
+                "idlMemberType"><a>unsigned long long</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the total number of connectivity check retransmissions received.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1319,7 +1319,7 @@
               </dt>
               <dd>
                 <p>
-                  Represents the total number of connectivity check requests received (includes retransmissions).
+                  Represents the total number of connectivity check requests received (including retransmissions).
                 </p>
               </dd>
               <dt>
@@ -1328,7 +1328,7 @@
               </dt>
               <dd>
                 <p>
-                  Represents the total number of connectivity check requests sent (does not include retransmissions).
+                  Represents the total number of connectivity check requests sent (not including retransmissions).
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1269,9 +1269,12 @@
               </dt>
               <dd>
                 <p>
-                  Represents the total RTT measured in seconds since the beginning of the session,
-                  based on both STUN connectivity check [[!STUN-PATH-CHAR]] responses
-                  (responsesReceived) and consent [[!RFC7675]] responses (consentResponsesReceived).
+                  Represents the sum of all RTT measurements in seconds since the
+                  beginning of the session, based on both STUN connectivity check
+                  [[!STUN-PATH-CHAR]] responses (responsesReceived) and consent
+                  [[!RFC7675]] responses (consentResponsesReceived). The average
+                  RTT can be computed from <code>totalRtt</code> by dividing it
+                  by (<code>responsesReceived</code> + <code>consentResponsesReceived</code>).
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1144,6 +1144,7 @@
              unsigned long long            bytesSent;
              unsigned long long            bytesReceived;
              double                        roundTripTime;
+             double                        currentRtt;
              double                        availableOutgoingBitrate;
              double                        availableIncomingBitrate;
              unsigned long long            requestsReceived;
@@ -1270,6 +1271,15 @@
                 <p>
                   Represents the average RTT measured in seconds, computed from both STUN
                   connectivity checks [[!STUN-PATH-CHAR]] and consent responses [[!RFC7675]].
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>currentRtt</code></dfn> of type <span class=
+                "idlMemberType"><a>double</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the last RTT measured in seconds.
                 </p>
               </dd>
               <dt>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1258,8 +1267,8 @@
               </dt>
               <dd>
                 <p>
-                  Represents the RTT computed by the STUN connectivity checks and measured
-                  in seconds [[!STUN-PATH-CHAR]].
+                  Represents the RTT measured in seconds, computed from both STUN
+                  connectivity checks [[!STUN-PATH-CHAR]] and consent responses [[!RFC7675]].
```
vr000m: @aboba. Is this the average RTT for the ICE candidate pair across both mechanisms? 

In RTP we report the last RTT measurement.

aboba:
@vr00m Yes, it is intended as an average.

webrtc-stats.html
```diff
@@ -1288,6 +1297,87 @@
...
+                  Represents the total number of connectivity check responses sent.
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>retransmissionsSent</code></dfn> of type <span class=
+                "idlMemberType"><a>unsigned long long</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the total number of connectivity check retransmissions sent.
```
vr000m: Do these retx apply to connectivity check requests and not to consent requests?

aboba:
@vr000m  Yes, because consent requests use a distinct transaction id, as specified in RFC 7675.

vr000m:
@aboba, Probably need retransmissionReceived as well.

aboba:
Added retransmissionsReceived counter.

vr000m:
Mostly, LGTM. 

Circling back to the RTT, I think adding a currentRtt metric may also be useful, especially when there are large delay spikes and the average might not be a good estimate. By having currentRtt the app may know if it can rely on the average RTT or not. 

What do you think?

alvestrand:
What is "current"? Last measured single value, or average over a shorter base?

Den 19. juni 2016 09:10:47 GMT+01:00, skrev Varun Singh <notifications@github.com>:
>Mostly, LGTM. 
>
>Circling back to the RTT, I think adding a currentRtt metric may also
>be useful, especially when there are large delay spikes and the average
>might not be a good estimate. By having currentRtt the app may know if
>it can rely on the average RTT or not. 
>
>What do you think? 
>
>---
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/w3c/webrtc-stats/pull/54#issuecomment-226985167

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

vr000m:
"Current" would mean the latest value. Although a small sliding window might work as well, say 5 consent packets. 

Is the typical consent checking interval 1s, with exponential back off.

aboba:
@vr000m   Added a currentRtt metric.

webrtc-stats.html
```diff
@@ -1258,8 +1269,17 @@
...
+                  connectivity checks [[!STUN-PATH-CHAR]] and consent responses [[!RFC7675]].
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>currentRtt</code></dfn> of type <span class=
+                "idlMemberType"><a>double</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the last RTT measured in seconds.
```
vr000m: Computed from consent or connectivity checks?

vr000m: Latest instead of last.

vr000m:
Minor nits, my proposal to tweak the language a bit:

averageRtt:

Represents the average RTT measured in seconds since the beginning of the session, 
computed from both STUN connectivity checks [[!STUN-PATH-CHAR]] and consent responses [[!RFC7675]].

currentRtt:

Represents the latest RTT measured in seconds, computed from both STUN connectivity checks [[!STUN-PATH-CHAR]] and consent responses [[!RFC7675]].

aboba:
@vr000m Updated the definitions per your suggestion.

vr000m:
LGTM. Does anyone else need to review this?

aboba:
@vr00m @taylor-b Taylor - can you take a look?

webrtc-stats.html
```diff
@@ -1143,9 +1143,20 @@
              boolean                       readable;
              unsigned long long            bytesSent;
              unsigned long long            bytesReceived;
-             double                        roundTripTime;
+             double                        averageRtt;
+             double                        currentRtt;
```
taylor-b: At the interim where we discussed this, I thought we decided on having a "totalRtt" that was the sum of all round trip times? That would allow you to compute the overall average over the whole session by dividing by responsesReceived, but also allows you to compute the average over a specific time interval. Did that decision change?

webrtc-stats.html
```diff
@@ -1288,6 +1310,96 @@
...
+                  Represents the total number of connectivity check retransmissions sent.
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>consentRequestsReceived</code></dfn> of type <span class=
+                "idlMemberType"><a>unsigned long long</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the total number of consent requests received.
```
taylor-b: Does this (and the stat below) include retransmissions?

webrtc-stats.html
vr000m: A total of Rtt/total responses is a good suggestion. Depending on the sampling rate, an endpoint would be able to measure average interval RTTs.

How often are the requests expected to be sent?

taylor-b: Well, the value of "Ta" (the time between connectivity checks) is up to the implementation. And it also depends on how many pairs you have and other factors. So it would be difficult to predict when connectivity checks are sent.

Consent checks, however, are recommended to be sent and random time intervals around 5 seconds, and the interval MUST be no shorter than 4 seconds.

vr000m: Thanks for the clarification. I think currentRtt still makes sense in case
the intervals are polling interval is shorter than Ta or seconds.

aboba:
@taylor-b @vr000m  Included totalRtt as Taylor suggested.  Since averageRtt can be computed from totalRtt/(responsesReceived + consentResponsesReceived), we no longer need averageRtt, correct?

webrtc-stats.html
```diff
@@ -1264,14 +1264,14 @@
...
-                <dfn><code>averageRtt</code></dfn> of type <span class=
+                <dfn><code>totalRtt</code></dfn> of type <span class=
                 "idlMemberType"><a>double</a></span>
               </dt>
               <dd>
                 <p>
-                  Represents the average RTT measured in seconds since the beginning of the session,
-                  computed from both STUN connectivity checks [[!STUN-PATH-CHAR]] and consent
-                  responses [[!RFC7675]].
+                  Represents the total RTT measured in seconds since the beginning of the session,
```
taylor-b: May be valuable to elaborate a bit more, for clarity. We could replace "the total RTT" with "the sum of all RTT measurements", and explain how this can be used to derive the average, since that will probably be the most common use.

taylor-b:
I agree. LGTM, with one comment.

webrtc-stats.html
vr000m: I agree.

webrtc-stats.html
aboba: Yes. Consent requests are not retransmitted (the next consent request is send with a different transaction id).

aboba:
@vroom @taylor-b  Have cleanup up the definition of totalRtt.  Hopefully it is clear now.

webrtc-stats.html
vr000m: Would it make sense to add that bit in. To elaborate that consent requests are not retransmitted

vr000m:
Yes. this looks good to me. 

I have a question related to the "pacing" discussion in ICE WG at IETF96. Consent Checks Requests that are dropped at the sender because the ICE pacing algorithm cannot schedule the packets at the time.  This might be premature to add in, however, wanted to make a note of it. Perhaps capture this as a new Github issue?

https://www.ietf.org/proceedings/96/slides/slides-96-ice-4.pdf

vr000m:
Is this ready to merge @aboba @alvestrand?

alvestrand:
lgtm. There are a couple of comments from @vr000m that look like they haven't been responded to - but if @aboba thinks these can be addressed later, I think we should merge.

aboba:
@vr000m - let's handle pacing in a separate issue.

aboba:
@alvestrand I believe it is now ready to merge.

Issue #26, opened 2015-11-05
Behavior of stats on closed, stopped and never-started tracks/SSRCs and so on needs to be clear


alvestrand:
The behavior agred on is that stats for an object, once initialized, will be available for the lifetime of the PC. When it's stopped, the clock stops advancing - the stats are the stats at the moment of stoppage, but the stats object does not disappear even if the object does.

This applies to all stats objects.

This needs to be clear.

alvestrand:
Also see #8

alvestrand:
#8 is the attempted fix.

alvestrand:
Need some generic language here too.

alvestrand:
There is an issue (mentioned by @jan-ivar ) about the case where a track moves between senders.

https://github.com/w3c/webrtc-pc/issues/572

vr000m:
Proposal accepted: 
+ State is frozen on stop. Including timestamp.
+ The last state lasts until the PC it belongs to is gone.

alvestrand:
Seems resolved by #68

PR #68, opened 2016-10-12
Design considerations section

### LINKED ISSUE #20: What to do when a function is not supported should be clear
There are a number of counters (for example sliCount) that are likely to be:
1) Only implemented for some track types (video in this case)
2) Only implemented by some platforms
It should be clear what the value of this should be when it's not going to be handled:
- zero always for a platform that does not count them?
- missing for a platform that doesn't count them?
- zero or missing for a track where it's not appliciable? (I think missing)

alvestrand:
Adds design guidance on adding stats, on stats values when not relevant,
and on stats lifetimes.

Closes #20

Note: #20 says that we need to note that zero is a valid value for jitter,
so can't be used as an initial value for "not set yet". This isn't added.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -102,6 +102,60 @@
     DOMString           id;
 	};
       </pre>
+      <section>
+	<h3>Guidelines for design of stats objects</h3>
+	<p>
+	  When introducing a new stats type, the following principles
+	  should be followed:
+	  <ul>
+	    <li>An RTCStats object should correspond to an object
+	      defined in the specification it supports.
+	    </li>
+	    <li>All members of the new object need to have definitions
+	      that make them consistently implementable. References to
+	      other specifications are a good way of doing this.
+	    </li>
+	    <li>All members need to have defined behavior for what
+	      happens before the thing it counts happens, or when the
+	      information it's supposed to show is not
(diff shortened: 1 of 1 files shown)


alvestrand:
Closes #26

webrtc-stats.html
```diff
@@ -102,6 +102,60 @@
...
+	<h3>Lifetime of stats objects</h3>
+	<p>The general rule is that stats objects, once created, exist
+	  for the lifetime of the PeerConnection that contains
+	  them, even when the underlying object they report on is
+	  stopped or deleted. This is important in order to report
+	  consistently on short-lived objects and to be able to report
+	  totals over the lifetime of a PeerConnection.</p>
+	<p>When an object is closed or deleted,
+	  the <code>timestamp</code> member of the stats object stops
+	  updating; it is frozen at the time when the object stopped.
```
vr000m: With the frozen timestamps, an app can filter out the older candidates, assuming the app has some idea about how often it queries or when it last queried the stats.

Although one concern I do have (and perhaps could be solved in a different way) is if the ice candidate is currently inactive and was active in the past, we do not really know when it was last active. We perhaps need a lastPacketSentTimestamp, lastPacketReceivedTimestamp or similar

PR #70, opened 2016-10-12
Added QP statistics

### LINKED ISSUE #57: Add QP statistics to video tracks
From Sami Kalliomäki sakal@google.com:

What we need is a way to get average QP over some period of time. I don't think adding this to the receiver side is needed by us but maybe it should be added as well. I think the approach of adding a sum of QP would be a good idea. It probably should be added to the RTCMediaStreamTrackStats structure. Maybe something like this:

qpSum of type unsigned long
Only valid for video. QP (quantization parameter) describes how much spatial detail is included a frame. Low value corresponds with good quality. The range of the value per frame is defined by the codec being used. This parameter represents the sum of all QPs for framesDecoded on remote streams and framesSent on local streams.

The average QP has been implemented before as a goog stat but it never landed. The relevant CLs are here:
https://codereview.webrtc.org/1264693003/
https://codereview.webrtc.org/1420963005/

alvestrand:
This Cl also includes the result of an "untabify", which makes review
difficult. Apologies.

Fixes #57

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -473,14 +473,19 @@ enum RTCStatsType {
                   The sum of the QP values of frames passed. The count
                   of frames is in <a>framesDecoded</a> for inbound
                   stream stats, and in <a>framesEncoded</a> for
-                  outbound stream stats. If QP varies across the frame,
-		  an average or typical QP value is recorded.
+                  outbound stream stats.
                 </p>
                 <p>
                   The definition of QP value depends on
-                  the <a>codec</a>; for VP8, QP values are defined in
-                  [[RFC6386]]. <!-- Need appropriate references for
-                  VP8, VP9 and H.264 - RFC 6386 doesn't use the term -->
+                  the <a>codec</a>; for VP8, the QP value is the value carried
+                  in the frame header as the syntax element "y_ac_qi", and
+                  defined in [[RFC6386]] section 19.2. Its range is 0..127.
+                  <!-- Need appropriate references for VP9 and H.264 -->
+                </p>
+                <p>
+                  Note that the QP value is only
+                  an indication of quantizer values used; many formats have
+                  ways to vary the quantizer value within the frame.
                 </p>
                 <p>Only valid for video.</p>
               </dd>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -114,16 +114,16 @@
     <section class="informative">
       <h2>Basic concepts</h2>
       <p>The stats API is defined in [[!WEBRTC]]. It is defined to return
-	a set of objects that are a subclass of the RTCStats dictionary. This is
-	normatively defined in [[!WEBRTC]], but is reproduced here
-	for ease of reference.
+        a set of objects that are a subclass of the RTCStats dictionary. This is
+        normatively defined in [[!WEBRTC]], but is reproduced here
+        for ease of reference.
       </p>
       <pre class="idl">
-	dictionary RTCStats {
+        dictionary RTCStats {
     DOMHighResTimeStamp timestamp;
     RTCStatsType        type;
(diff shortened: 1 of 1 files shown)


alvestrand:
Note - this includes changes to the MediaStreamTrack statistics to remove references to "ssrc", since a MediaStreamTrack can have multiple ssrcs (in the case of simulcast, layered decoding or other complexities).

alvestrand:
See discussion in bug - RTCMediaStreamTrackStats has no codec, and codec is necessary for interpreting QP.
In cases of simulcast, where a frame from one MediaStreamTrack can be encoded on multiple RTP streams, often with different frame rates, I think it's better to keep the counters for outgoing on the RTP stream - and once we have that, putting the counters for incoming on the MST doesn't make sense.

Scintillo:
I noticed that we actually have the QP available before decoding the frame. So for receiving streams, maybe we should have framesReceived instead of framesDecoded? The other option is just to postpone increasing the qp sum in the code.

alvestrand:
Is the QP available before decoding the frame for all of {H.264, VP8, VP9}?
I'm a bit worried about rumbles that QP might be variable across a frame in AOM, but let's burn that bridge when AOM's finished.

Scintillo:
> Is the QP available before decoding the frame for all of {H.264, VP8, VP9}?

It looks like it currently decoded only for VP8. It would be quite easy to add it for H.264. I don't know how hard it would be to add it for VP9. However, getting it during decoding is also tricky because many hardware decoders might not export the information.

Scintillo:
To be noted though is that decoding the QP value separately adds some duplicate processing and some implementations might prefer to get it from the decoder. So maybe it is better for the standard to allow greater flexibility at the cost of small implementation overhead on our side.

stefhak:
Regarding https://github.com/w3c/webrtc-stats/pull/70#issuecomment-254515763, it's my understanding that for most implementations QP varies over a picture also for H.264 and 265 (and can be made to do so also for VP9 by using segments). However, perhaps reporting the average over the picture is good enough.

Scintillo:
Yeah, I have also heard that frames have multiple QP values for each "block". I have also heard that the way we parse QP right now might be broken. My understanding is that we just take the "base QP" of the frame the block QPs are offset of. This might or might not be the average of the block QPs. However, I haven't read the spec/code and this is mostly hearsay.

Getting the average of the block QPs might be more expensive. Therefore, I think it would be the right thing to just keeps framesDecoded.

webrtc-stats.html
```diff
@@ -251,6 +251,7 @@
              unsigned long pliCount;
              unsigned long nackCount;
              unsigned long sliCount;
+             unsigned long qpSum;
```
alvestrand: Changed to "long long".

webrtc-stats.html
```diff
@@ -360,10 +361,28 @@
...
+                <p>
+                  The sum of the QP values of frames passed. The count
+                  of frames is in <a>framesDecoded</a> for inbound
+                  stream stats, and in <a>framesEncoded</a> for
+                  outbound stream stats.
+                </p>
+                <p>
+                  The definition of QP value depends on
+                  the <a>codec</a>; for VP8, QP values are defined in
+                  [[RFC6386]]. <!-- Need appropriate references for
```
alvestrand: Noted in text comment. Should be checked before merge.

alvestrand:
Sorry for the noise, I rebased on the wrong version of master.
Some comments are adddressed, I still miss references for QP values. Working on it.

alvestrand:
A suitable reference for QP in VP8 seems to be y_ac_qi (the luma quantizer index) in section 19.2 of RFC 6386.

alvestrand:
@vr000m @Scintillo are we done?

PR #71, opened 2016-10-12
Introduced the term "RTP stream", with definition.

### LINKED ISSUE #58: Need to define "RTP stream" in terminology section
The term "media stream" is heavily overloaded, and should not be used unqualified.
The term "RTP stream" is used in the document, and needs a reference to the correct RFC in the terminology section.

All occurences of "media stream" without qualifier should be updated to reference the term (with links).

alvestrand:
Went through the document and removed all occurences of "media stream",
and did some other cleanups.

Fixes #58


alvestrand:
@henbos you might want to check that I got it right.

Issue #72, opened 2016-10-17
Not clear if RTCTransportStats combines RTP+RTCP or not.


taylor-b:
The summary states:

> A Transport carries a part of an SDP session, consisting of RTP and RTCP. When Bundle is in use, an SDP session will have only one Transport per Bundle group. When Bundle is not in use, there is one Transport per m-line.

This heavily implies that an RTCTransportStats combines RTP and RTCP for an m= section.

But later, we have rtcpTransportStatsId, defined as:

> If RTP and RTCP are not multiplexed, this is the id of the transport that gives stats for the RTCP component, and this record has only the RTP component stats.

Which implies that there are separate RTCTransportStats for RTP and RTCP. This seems correct, since that would allow a 1:1 mapping from an RTCIceTransport to an RTCTransportStats. Also, it's the only way there could be a single "selected candidate pair" per RTCTransportStats.

alvestrand:
The intent of the spec (I think this was hashed out in the Washington interim, a few years back) is that in the normal case (RTCP-mux), there would be only one transport object; in the exceptional case of unbundled RTCP, there would be two (has to be, given that selectedCandidateId will be different), but the only way to find the RTCP one would be through the RTP one.

taylor-b:
That's what I'd assumed. In that case I think the summary just needs to be updated to make this clear. I'll make a PR.

alvestrand:
PR is #73

alvestrand:
Fixed in #73

PR #73, opened 2016-10-18
Fixing the summary of RTCTransportStats (issue #72).

taylor-b:
The new summary should make it clear that there is a 1:1 mapping between
an RTCTransportStats and an RTCDtlsTransport/RTCIceTransport pair.
Meaning, when RTCP muxing is _not_ used, you get separate
RTCTransportStats for RTP and RTCP. Also adding links to the appropriate
terms, and to WebRTC (to describe RTCP muxing and bundling in more
detail).

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -105,8 +105,9 @@
         defined in [[!GETUSERMEDIA]].
       </p>
       <p>
-        The terms <dfn>RTCPeerConnection</dfn>, <dfn>RTCDataChannel</dfn> are defined in
-        [[!WEBRTC]].
+        The terms <dfn>RTCPeerConnection</dfn>, <dfn>RTCDataChannel</dfn>,
+        <dfn>RTCDtlsTransport</dfn> and <dfn>RTCIceTransport</dfn> are defined
+        in [[!WEBRTC]].
       </p>
       <p>
 	The term <dfn>RTP stream</dfn> is defined in [[RFC7656]] section 2.1.10.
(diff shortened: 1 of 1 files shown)


taylor-b:
I put this in terms of "MediaStreamTrack"s, since that's what the rest of the document does, but is that really correct? In reality, each transport is linked to a transceiver, not a track.

webrtc-stats.html
```diff
@@ -1128,9 +1129,18 @@
...
+          underlying <code><a>RTCIceTransport</a></code>. When RTCP multiplexing
+          is used, one transport is used for both RTP and RTCP. Otherwise, RTP
+          and RTCP will be sent on separate transports, and
+          <code>rtcpTransportStatsId</code> can be used to pair the resulting
+          <code>RTCTransportStats</code> objects. Additionally, when bundling
+          is used, a single transport will be used for all
+          <code><a>MediaStreamTrack</a></code>s in the bundle group. If
+          bundling is not used, different <code><a>MediaStreamTrack</a></code>
+          will use different transports. RTCP multiplexing and bundling are
+          described in [[!WEBRTC]].
```
vr000m: Since there is DtlsTransport and IceTransport, I am not sure if it is clear to a novice reader which "transports" are changing or multiplexed.

Do we need to clarify that somewhere?

alvestrand: I may have been reading this stuff too long, but I'm finding the text suggested to be clear enough for me.

vr000m:
related to #73

PR #76, opened 2016-11-01
Adding missing "codec" RTCStatsType.

taylor-b:
Assuming "codec" is what we want for RTCCodecStats.
Also adding a missing description for "codecId".


alvestrand:
I take this reordering request back. This corresponds to the order in which they are listed in the section below. Which order they *should* be in is a matter the editors need to dispose of. Merging.

Issue #74, opened 2016-10-24
Need consistent name for RTT estimate on RTCRTPStreamStats and on ICECandidatePair


alvestrand:
We currently have an RTT stat on a candidate pair, but the RTP spec defines a different RTT metric (based on RTCP and timestamp correlations).

Consider whether we should expose this too, and if so, how.

vr000m:
@alvestrand we do have Rtt in OutboundStats:
```
dictionary RTCOutboundRTPStreamStats : RTCRTPStreamStats {
    unsigned long      packetsSent;
    unsigned long long bytesSent;
    double             targetBitrate;
    double             roundTripTime;
};
```

alvestrand:
Oops. Sorry. Closing as "missed it".

fippo:
which means (which caused this issue and confused me as well) that we now have one roundTripTime (on stream stats) and two "rtt" in https://w3c.github.io/webrtc-stats/#dom-rtcicecandidatepairstats-currentrtt

For the sake of consistency either both should be abbreviated or none. I have no opinion which one ;-)

vr000m:
we could change it to RTT. @alvestrand, @jan-ivar -- what do you think?

alvestrand:
Reopening and renaming issue.
Stylistically, I prefer the longer name. But I'd prefer to go with whatever's implemented first. @ShijunS @jan-ivar?

jan-ivar:
@vr000m  We [implement](https://dxr.mozilla.org/mozilla-central/rev/215f9686117673a2c914ed207bc7da9bb8d741ad/dom/webidl/RTCStatsReport.webidl#51) `mozRtt` in RTCInboundRTPStreamStats, since rtcp info is in the [opposite dictionary](http://stackoverflow.com/a/29840886/918910) . E.g.

```
{ "id": "outbound_rtcp_video_1",
 "timestamp": 1477343155549.51,
 "type": "inboundrtp", "isRemote": true,
 "mediaType": "video",
 "remoteId": "outbound_rtp_video_1",
 "ssrc": "892087744",
 "bytesReceived": 99076,
 "jitter": 0.289,
 "mozRtt": 1,
 "packetsLost": 0,
 "packetsReceived": 124 
}
```

How come you have it on RTCOutboundRTPStreamStats ?

You can get a dump in FF from https://jsfiddle.net/jib1/moz71L7h/

@alvestrand I have no strong preference on naming.

alvestrand:
[...]

Issue #78, opened 2016-11-01
Consider putting roundTripTime on RTCInboundStreamStats


alvestrand:
Per mailing list discussion: For a local RTCOutboundRTPStreamStats, the RTT is calculated based on received RTCP RR reports. The rest of the RTCP-derived data is placed in the remote RTCInboundRTPStreamStats, with an appropriate timestamp. RTT should go there too.

Consideration: Can RTT be calculated at all on local RTCInboundStreamStats?

alvestrand:
See #74 for background and discussion.

PR #95, opened 2016-12-08
Rename "RTT" to RoundTripTime

alvestrand:
Partial resolution of issue #74


alvestrand:
@vr000m @jan-ivar @fippo is thiis an acceptable resolution of #74?
I put "partial" in the description here, but now I can't figure out what else is needed for #74.

alvestrand:
Closes #74 

That should be the magic token :-)

vr000m:
Although we still need to move the rtt from the outbound to inbound remote

jan-ivar:
@vr000m Is there an issue for that?

fippo:
its #78

PR #114, opened 2016-12-16
Adding minor clarification regarding stats object lifetime.

taylor-b:
A stats object is created at the same time as the underlying object; NOT
when getStats is called. That means that if an underlying object is
created and destroyed in-between an application calling getStats, it
should still appear in the stats.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -183,11 +183,11 @@
           Lifetime of stats objects
         </h3>
         <p>
-          The general rule is that stats objects are created at the same time as the underlying
-          object they report on, and once created, they exist for the lifetime of the PeerConnection
-          that contains them, even when the underlying object is stopped or deleted. This is
-          important in order to report consistently on short-lived objects and to be able to report
-          totals over the lifetime of a PeerConnection.
+          The general rule is that stats objects are considered to be created at the same time as
+          the underlying object they report on, and once created, they exist for the lifetime of the
+          PeerConnection that contains them, even when the underlying object is stopped or deleted.
+          This is important in order to report consistently on short-lived objects and to be able to
+          report totals over the lifetime of a PeerConnection.
         </p>
         <p>
           When an object is closed or deleted, the <code>timestamp</code> member of the stats
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -183,10 +183,11 @@
           Lifetime of stats objects
         </h3>
         <p>
-          The general rule is that stats objects, once created, exist for the lifetime of the
-          PeerConnection that contains them, even when the underlying object they report on is
-          stopped or deleted. This is important in order to report consistently on short-lived
-          objects and to be able to report totals over the lifetime of a PeerConnection.
+          The general rule is that stats objects are created at the same time as the underlying
+          object they report on, and once created, they exist for the lifetime of the PeerConnection
+          that contains them, even when the underlying object is stopped or deleted. This is
+          important in order to report consistently on short-lived objects and to be able to report
+          totals over the lifetime of a PeerConnection.
         </p>
         <p>
           When an object is closed or deleted, the <code>timestamp</code> member of the stats


webrtc-stats.html
```diff
@@ -183,10 +183,11 @@
           Lifetime of stats objects
         </h3>
         <p>
-          The general rule is that stats objects, once created, exist for the lifetime of the
-          PeerConnection that contains them, even when the underlying object they report on is
-          stopped or deleted. This is important in order to report consistently on short-lived
-          objects and to be able to report totals over the lifetime of a PeerConnection.
+          The general rule is that stats objects are created at the same time as the underlying
```
alvestrand: I'd formulate this as "are considered to be created". It's a perfectly valid implementation strategy to create "frozen" stats objects when the underlying objects are destroyed, and creating the "live" stats objects when queried - in fact, that's an implementation strategy I've been meaning to suggest to @henbos.

The observable behavior should be indistinguishable from the text you suggest.

taylor-b: Is this covered by: "Conformance requirements phrased as algorithms or specific steps may be implemented in any manner, so long as the end result is equivalent"?

alvestrand: Yes, but we have a long tradition of people treating "are" as an implementation requirement without considering whether or not it's observable.

vr000m:
LGTM

jan-ivar:
I had trouble reading this, so I've opened https://github.com/w3c/webrtc-stats/issues/137.

Also, note that Firefox does not work this way. If you remove an object then stats for it go away. Are we sure we want stats reports to grow forever like this?

What are we supposed to return for objects that are gone? Currently we take a snapshot on `pc.close` only. This spec seems to require us to cache information all the time, just in case some object will be removed later.

alvestrand:
The implementation I think makes the most sense is that one attaches to the destructor of each object a call that "pickles" the stats and stores it in a table for later stats calls to return.
The alternative is that objects (that may consume resources or influence events) between two stats calls are permanently invisible to the stats API - this did not seem like a desirable feature.
ie if you have stats at 5-second intervals, and someone is doing 1-second datachannels, you will never see anything about them (except the counters in the peerconnection stats object increasing).

jan-ivar:
@alvestrand 1-second data channels? Best ignored IMHO. After an hour, how many stats objects would we need to return otherwise?

alvestrand:
@jan-ivar 3600. At 100 bytes each, that's 360 Kbytes. Or 1/3 of a VGA-sized raw video buffer, which is ~a megabyte.

henbos:
We don't want the caller to miss out on stats that may be of interest. But an hour old 1-second data channel does not sound very interesting. _If_ we want to stop it from growing indefinitely we could either:
1. After an object is deleted, return its stats object the next time getStats is called and then remove it from the cache.
2. Have an API for clearing deleted stats.
3. Have a time limit, say if an object was removed >1 minute ago we remove it from the stats cache. The caller(s) would have had plenty of time to get it if they were interested in it.

1+2 will not fly because we are agnostic to how many callers there are to getStats and these would interfer with other callers. 2 is also racey unless you explicitly state what to clear, this would be a bad API IMO. Also problem with 1+2 is if you don't call this the stats cache (memory) keeps growing indefinitely.

What about 3?

If someone is interested in the complete history of stats objects they could cache all detached/removed stats themselves, I think that is an unusual usage case but it would be easy to do for anyone interested.

alvestrand:
My main worry with throwing away the stats is that we lose all sense of completeness. Currently, with "frozen" stats, we can guarantee that the data sent over the transport can be attributed to *some* client stats object - what's left is protocol overhead.
The same issue applies to candidates - when we keep them all, we can guarantee that adding up bytes for all candidates will account for the bytes sent on the transport. If we start throwing away candidates, we don't know that.

jan-ivar:
I worry this comes at the cost of getting an accurate snapshot of the current state. We're mixing models IMHO. A "snapshot" to me means just that, a snapshot of the *current state*. If every object that ever lived hangs around, then that's not a snapshot.

This pushes the problem of reading current state correctly onto the caller: Are people supposed to check timestamps on every object to deduce which objects are real and which are ghosts? Live rtcp timestamps may beat only every ~5 seconds, so getting an accurate snapshot can get tricky.

Do tracks, transports ever go away except in response to JS? If not, then it seems simple enough for JS to call `getStats` before `removeTrack` if it wants to collect everything that happened.

alvestrand:
If it calls getStats, 100 kbytes arrive, and it calls removeTrack, where are those 100 kbytes going to be visible?

wrt the problem of knowing whether the object is active or not: we added "deleted" to ICE candidates (https://w3c.github.io/webrtc-stats/webrtc-stats.html#dom-rtcicecandidatestats-deleted) - I have wondered afterwards if this was at the wrong level, and we should instead have it as a generic RTCStats dictionary member, since it seems to apply in many contexts.

henbos:
(And RTCMediaStreamTrackStats has [detached](https://w3c.github.io/webrtc-stats/#dom-rtcmediastreamtrackstats-detached)).

jan-ivar:
@alvestrand Same as what happens if it calls removeTrack, then 100kbytes arrive.

PR #122, opened 2017-01-02
Replace RTCTransportStats.active with .dtlsState

### LINKED ISSUE #67: Need definition for RTCTransportStats.activeConnection
The definition for RTCTransportStats.activeConnection is "Set to true when transport is active."

But there's no definition of what "active" means. It might be that the RTCIceTransport that it corresponds to has its state in {connecting, completed}, or it might mean something else. Clarification needed.

alvestrand:
"active" was not well defined. "state" is more specific, and
is well defined.
Closes #67

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -106,7 +106,7 @@
       </p>
       <p>
         The terms <dfn>RTCPeerConnection</dfn>, <dfn>RTCDataChannel</dfn>,
-        <dfn>RTCDtlsTransport</dfn>. <dfn>RTCDtlsTransportState</dfn> and
+        <dfn>RTCDtlsTransport</dfn>, <dfn>RTCDtlsTransportState</dfn> and
         <dfn>RTCIceTransport</dfn> are defined in [[!WEBRTC]].
       </p>
       <p>
@@ -1350,7 +1350,8 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Set to the current value of the "state" field of the underlying RTCDtlsTransport.
+                  Set to the current value of the "state" attribute of the underlying
+                  RTCDtlsTransport.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1204,7 +1204,7 @@ enum RTCStatsType {
              DOMString           label;
              DOMString           protocol;
              long                datachannelid;
-             RTCDataChannelState state;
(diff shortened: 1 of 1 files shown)


vr000m:
Shouldn't it be called dtlsState instead of state?

alvestrand:
Seems like a good idea, given that we've considered adding iceState too. I only added one state in this pass.

henbos:
Whether we call this dtlsState or state, LGTM. This stat is trivial to implement in Chromium's WebRTC.

alvestrand:
Changed to dtlsState, to make room for also reflecting the iceState.

henbos:
You just changed the name of RTCDataChannelStats.state to dtlsState, not RTCTransportStats

alvestrand:
Oops. Does this look better?

henbos:
LGTM.
Should you reference the webrtc-pc spec for the use of RTCDtlsTransportState?

alvestrand:
@henbos already referenced in section 3!

vr000m:
LGTM.

webrtc-stats.html
```diff
@@ -106,7 +106,8 @@
       </p>
       <p>
         The terms <dfn>RTCPeerConnection</dfn>, <dfn>RTCDataChannel</dfn>,
-        <dfn>RTCDtlsTransport</dfn> and <dfn>RTCIceTransport</dfn> are defined in [[!WEBRTC]].
+        <dfn>RTCDtlsTransport</dfn>. <dfn>RTCDtlsTransportState</dfn> and
```
taylor-b: Should be comma instead of period

webrtc-stats.html
```diff
@@ -1344,12 +1345,12 @@
...
               <dt>
-                <dfn><code>activeConnection</code></dfn> of type <span class=
-                "idlMemberType"><a>boolean</a></span>
+                <dfn><code>dtlsState</code></dfn> of type <span class=
+                "idlMemberType"><a>RTCDtlsTransportState</a></span>
               </dt>
               <dd>
                 <p>
-                  Set to <code>true</code> when transport is active.
+                  Set to the current value of the "state" field of the underlying RTCDtlsTransport.
```
taylor-b: Is "attribute" or "field" the right word in WebIDL terminology?

alvestrand:
Will fix.

webrtc-stats.html
alvestrand: I think "field" is usable, but "attribute" is better.

PR #123, opened 2017-01-02
Explain "sum and count" design paradigm

### LINKED ISSUE #69: Describe the reason for "sum & count" rather than "average"
In the design considerations, we should include explanations of why we use "sum & count" rather than "average" - the reason being that "average" implies averaging over some fixed time interval, which may or may not be what the app wants; sum & count allows us to do getStats() twice and calculate the average between them over an app-chosen time interval - which may be long or short.

alvestrand:
Closes #69


alvestrand:
@vr000m please review.

alvestrand:
Explaining the source of timestamps is a new issue.

PR #126, opened 2017-01-03
Fix RTCStatsType for "stream"

### LINKED ISSUE #80: Messed up "stream" and "track"
At the moment, the text relevant to this says:

"track"
Contains the sequence of tracks related to a specific media stream and the corresponding media-level metrics. It is accessed by the RTCMediaStreamStats.

This is obviously wrong. The RTCStatsType string for a MediaStream should be "stream", and "track" should point to RTCMediaStreamTrackStats. The term "media stream" shouldn't be used without a reference.

alvestrand:
This also untangles some definitions, and adds a link to the certificate
data type from the identifier spec.

Closes #80

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -250,6 +250,7 @@ enum RTCStatsType {
 "outbound-rtp",
 "peer-connection",
 "data-channel",
+"stream",
 "track",
 "transport",
 "candidate-pair",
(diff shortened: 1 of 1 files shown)


alvestrand:
@vr000m please review

vr000m:
LGTM

PR #127, opened 2017-01-03
Added "kind" to RTCMediaStreamTrackStats.

### LINKED ISSUE #84: RTCMediaStreamTrackStats does not have mediaType member.
The "RTC[In/Out]boundRTPStreamStats : RTCRTPStreamStats" dictionaries are valid for both audio and video. RTCRTPStreamStats has a DOMString mediaType that is either "audio" or "video" to distinguish the two cases.

RTCMediaStreamTrackStats is also valid for both audio and video, but it has no mediaType member. You'd have to examine the other members (some of which are only valid for audio or video) to deduce if its for audio or video.

We should add "DOMString mediaType" to RTCMediaStreamTrackStats.
- Side-question: Why is this a DOMString and not an enum? Should we change it to enum for both dictionaries?

Alternatively we could split it up into "RTCMediaStream[Audio/Video]TrackStats : RTCMediaStreamTrackStats". Its members are mostly audio- or video-specific so why are all of the members in the same dictionary?

alvestrand:
Also noted that the RTCRTPStreamStats MediaType MUST match the
track it's attached to.

Closes #84

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -431,7 +431,8 @@ enum RTCStatsType {
                 <p>
                   Either "<code>audio</code>" or "<code>video</code>". This MUST match the media
                   type part of the information in the corresponding <code><a>codec</a></code>
-                  member of <code>RTCCodecStats</code>.
+                  member of <code>RTCCodecStats</code>, and MUST match the "kind" attribute of the
+                  related MediaStreamTrack.
                 </p>
               </dd>
               <dt>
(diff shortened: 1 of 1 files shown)


henbos:
Is there not an enum that can be used for this?

vr000m:
LGTM

alvestrand:
It's DOMString in the base spec. There's an enum around too, but it's not the API value.

PR #128, opened 2017-01-04
Define audio level rigidly

### LINKED ISSUE #81: Volume needs a more complete definition.
Current text:

audioLevel of type double
Only valid for audio, and the value is between 0..1 (linear), where 1.0 represents 0 dBov. Calculated as defined in [RFC6464].

RFC 6464 defines a logartihmic scale with -127 pegged to silence. That's not what is meant here.
"Linear" comes close, but not quite enough, I think.

I think we want 0..1, where 0 is silence, 1 is 0 dBov, and 0.5 represents "approximately 6 dBSPL change in the sound pressure level from the maximum" (quoted from the mediacapture draft's "volume" spec).

alvestrand:
This definition should match the one in mediacapture-main for "volume",
and also shows how to relate it to RFC 6465 levels.

Fixes #81

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1203,15 +1203,17 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Only valid for audio, and the value is between 0..1 (linear), where 1.0
+                  Only valid for audio. The value is between 0..1 (linear), where 1.0
                   represents <code>0 dBov</code>, 0 represents silence, and 0.5 represents
                   approximately 6 dBSPL change in the sound pressure level from 0 dBov.
                 </p>
                 <p>
-                  The "audio level" value defined in [[!RFC6464]] (defined as 0..127, where 0
-                  represents 0 dBov, 126 reprsents -126 dBov and 127 represents silence) can be
-                  obtained by the calculation given in appendix A of [[RFC6465]]: level =
-                  -log10(audioLevel) * 20.
+                  The "audio level" value defined in [[RFC6464]] and used in the
+		  RTCRtpContributingSource.audioLevel of [[WEBRTC]] (defined as 0..127, where 0
+                  represents 0 dBov, 126 represents -126 dBov and 127 represents silence) is
+                  obtained by the calculation given in appendix A of [[!RFC6465]]: informally, level =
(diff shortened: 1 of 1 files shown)


alvestrand:
@aboba @fluffy please check.

webrtc-stats.html
```diff
@@ -1167,7 +1167,14 @@
               <dd>
                 <p>
                   Only valid for audio, and the value is between 0..1 (linear), where 1.0
-                  represents <code>0 dBov</code>. Calculated as defined in [[!RFC6464]].
+                  represents <code>0 dBov</code>, 0 represents silence, and 0.5 represents
+                  approximately 6 dBSPL change in the sound pressure level from 0 dBov.
+                </p>
+                <p>
+                  The "audio level" value defined in [[!RFC6464]] (defined as 0..127, where 0
+                  represents 0 dBov, 126 reprsents -126 dBov and 127 represents silence) can be
```
fippo: typo: reprsents -> represents

vr000m: lower case must?

webrtc-stats.html
```diff
@@ -1167,7 +1167,14 @@
               <dd>
                 <p>
                   Only valid for audio, and the value is between 0..1 (linear), where 1.0
-                  represents <code>0 dBov</code>. Calculated as defined in [[!RFC6464]].
+                  represents <code>0 dBov</code>, 0 represents silence, and 0.5 represents
+                  approximately 6 dBSPL change in the sound pressure level from 0 dBov.
+                </p>
+                <p>
+                  The "audio level" value defined in [[!RFC6464]] (defined as 0..127, where 0
```
alvestrand: The mapping is normative, and should be marked as such, even though it's not used here (it's used when mapping to RTP headers).

vr000m:
Minor comments, otherwise LGTM

webrtc-stats.html
alvestrand: Used "is obtained" rather than "can be obtained". Made 6465 normative and 6464 informative.

webrtc-stats.html
alvestrand: Can't find a must to change the case of.

PR #138, opened 2017-01-17
Make a RTCMediaStreamTrackStats object per track attachment

### LINKED ISSUE #130: Track stats: track or attachment?
In implementation discussions, it was asked what the stats for a track mean for the case where a track is attached to a PC multiple times.

This is not possible using AddTrack, but is possible using ReplaceTrack or AddTransceiver.
The WG has previously said that this is "no special treatment" - if you ask for a track to be sent twice, it will be sent twice, and surface at the receiver as two tracks.

The cleanest way out seems to be to say that the track stats track the attachment of the track to an RTCPeerConnection, so that the sequence

pc.addTransceiver(track=a)
pc.addTransceiver(track=a)
pc.getStats(a)

will return a stats object with two RTCMediaStreamTrackStats objects in it.
For receivers, there is no such discrepancy, because a track is always sourced from a single receiver, and can't be detached from it.

@taylor-b @henbos @jan-ivar - comments requested.

alvestrand:
This makes it clear that one track attachment exists for each time
the track is attached.

Closes #130

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1051,7 +1051,7 @@ enum RTCStatsType {
           <dfn>RTCMediaStreamTrackStats</dfn> dictionary
         </h3>
         <p>
-          An RTCMediaStreamTrackStats object represents the stats about one connection of a
+          An RTCMediaStreamTrackStats object represents the stats about one attachment of a
           MediaStreamTrack to the PeerConnection object for which one calls getStats.
         </p>
         <p>
@@ -1064,8 +1064,8 @@ enum RTCStatsType {
           "trackIdentifier" attribute, but different "id" attributes.
         </p>
         <p>
-          If the connection is terminated (via removeTrack or via replaceTrack), it continues to
-          appear, but with the "detached" member set to True.
+          If the track is detached from the PeerConnection (via removeTrack or via replaceTrack),
+          it continues to appear, but with the "detached" member set to True.
         </p>
         <div>
           <pre class="idl">dictionary RTCMediaStreamTrackStats : RTCStats {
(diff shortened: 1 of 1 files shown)


alvestrand:
@jan-ivar @henbos please review

alvestrand:
@vr000m can you review and merge?

henbos:
Bump, review and merge? This is now Chromium's behavior.

webrtc-stats.html
```diff
@@ -1093,7 +1100,7 @@
               </dt>
               <dd>
                 <p>
-                  Represents the <code>track.id</code> property.
+                  Represents the <code>track.id</code> property of the track.
```
jan-ivar: Maybe: *id* property of track? Just to not have track twice?

alvestrand: Syntactically better. We have a pretty heavy collision of "id" names here, but I guess it's clear enough.

webrtc-stats.html
```diff
@@ -1051,14 +1051,21 @@
...
-          the "detached" member set to True.
+          ReplaceTrack on an RTPSender object, or via being created on an RTPReceiver object).
+        </p>
+        <p>
+          If an outgoing track is attached twice (via addTransceiver or ReplaceTrack), there will
+          be two RTCMediaStreamTrackStats objects, one for each attachment. They will have the same
+          "trackIdentifier" attribute, but different "id" attributes.
+        </p>
+        <p>
+          If the connection is terminated (via removeTrack or via replaceTrack), it continues to
```
alvestrand: Or "if the track is detached".

alvestrand:
Harald will update and merge.

PR #139, opened 2017-01-23
Define terminology for "stats object" et al.

### LINKED ISSUE #136: Are the stats read-only snapshots or will they change kept around?
I was having trouble understanding whether stats objects can be changed after they are pulled via getStats. I assert that they should be well defined as either being snapshotted when getStats is called OR as being updated over time.

### LINKED ISSUE #137: "stats object" terminology is confusing
This document seems to use the term "stats object" to mean different things:
 1. a dictionary in the RTCStatsReport map returned from getStats, a.k.a. "RTCStats object",
 2. some internal "underlying object" presumably with a lifetime as long as the peer connection,
 3. A "stats type", or the full implementation of one [...](https://w3c.github.io/webrtc-stats/#guidelines-for-design-of-stats-objects), or
 4. some general term to mean all of these things.

E.g. This [nugget](https://w3c.github.io/webrtc-stats/#lifetime-of-stats-objects) seems to use all of the above in one paragraph, and I'm having trouble reading it. I think we need clearer terminology here.

alvestrand:
This defines a "stats object" as the invisible object that keeps
track of stats, and coins the term "stats value set" for the object
that is returned from the API.

Closes #137
Closes #136

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -82,11 +82,6 @@
         Terminology
       </h2>
       <p>
-        The <code><a href=
-        "https://www.w3.org/TR/2014/REC-html5-20141028/webappapis.html#eventhandler">EventHandler</a></code>
-        represents a callback used for event handlers as defined in [[!HTML5]].
-      </p>
-      <p>
         The concepts <dfn><a href=
         "https://www.w3.org/TR/2014/REC-html5-20141028/webappapis.html#queue-a-task">queue a
         task</a></dfn>, and <dfn><a href=
@@ -118,9 +113,56 @@
         Basic concepts
       </h2>
       <p>
-        The stats API is defined in [[!WEBRTC]]. It is defined to return a set of objects that are
-        a subclass of the RTCStats dictionary. This is normatively defined in [[!WEBRTC]], but is
-        reproduced here for ease of reference.
+        The basic object of the stats model is the <a>stats object</a>. The following terms are
+        defined to describe it:
+      </p>
+      <dl>
+        <dt>
+          <dfn>Stats object</dfn>
+        </dt>
+        <dd>
+          <p>
+            This represents an internal object that keeps a set of data values, together with a
+            time for when these values are current. Most stats objects are related one-on-one to an
+            object defined in the WebRTC API; they may be thought of as being internal properties
+            of those objects.
+          </p>
+        </dd>
+        <dt>
+          <dfn>Stats object reference</dfn>
+        </dt>
+        <dd>
+          <p>
+            A stats object has a stable identifier "id". Other stats objects may contain references
+            to this stats object. In a <a>stats value set</a>, these references are represented by
+            a DOMString containing "id" value of the referenced stats object.
+          </p>
+          <p>
+            The stats object is defined by describing the stats value set it returns.
+          </p>
+        </dd>
+        <dt>
+          <dfn>Stats value set</dfn>
+        </dt>
+        <dd>
+          This is a set of values, copied out from a stats object at a specific moment in time. It
+          is recturned as a WebIDL "dictionary" through the GetStats API.
+        </dd>
+        <dt>
+          <dfn>Stats value</dfn>
+        </dt>
+        <dd>
+          Refers to a single value within a stats value set.
+        </dd>
+      </dl>
+      <p>
+        A stats object changes the values it contains continuously over its lifetime, but is never
+        visible through the GetStats API. A stats value set, once returned, never changes.
+      </p>
+      <p>
+        The stats API is defined in [[!WEBRTC]]. It is defined to return a collection of <a>stats
+        value set</a>s that are a subclass of the RTCStats dictionary. This API is normatively
+        defined in [[!WEBRTC]], but is reproduced here for ease of reference.
       </p>
       <pre class="idl">
         dictionary RTCStats {
@@ -134,11 +176,15 @@
           Guidelines for design of stats objects
         </h3>
         <p>
-          When introducing a new stats type, the following principles should be followed:
+          When introducing a new stats object, the following principles should be followed:
         </p>
         <ul>
           <li>An RTCStats object should correspond to an object defined in the specification it
           supports.
+          </li>
+          <li>The object MUST define a new value in the "RTCStastType" enum, and MUST define the
+          syntax of the <a>stats value set</a> it returns either by reference to an existing
+          sub-dictionary of <a>RTCStats</a> or by defining a new sub-dictionary of <a>RTCStats</a>.
           </li>
           <li>All members of the new object need to have definitions that make them consistently
           implementable. References to other specifications are a good way of doing this.
@@ -153,10 +199,9 @@
           (camelCase).
         </p>
         <p>
-          Names ending in "Id" (such as "datachannelId") are always of type DOMString and are used
-          to refer to other stats objects by the value of their "id" field; names ending in "Ids"
-          (such as "trackIds") are always of type sequence&lt;DOMString&gt;, and are used to refer
-          to a list of other stats objects.
+          Names ending in "Id" (such as "datachannelId") are always a <a>stats object
+          reference</a>; names ending in "Ids" (such as "trackIds") are always of type
+          sequence&lt;DOMString&gt;, where each DOMString is a <a>stats object reference</a>.
         </p>
         <p>
           Stats are sampled by Javascript. In general, an application will not have overall control
@@ -216,16 +261,16 @@
     </section><!-- VS: getStats() remains in base spec! -->
     <section>
       <h2>
-        Maintenance procedures for stats types
+        Maintenance procedures for stats object types
       </h2>
       <p>
         This document, in its editors' draft form, serves as the repository for the currently
-        defined set of stats types.
+        defined set of stats object types.
       </p>
       <p>
-        If a need for a new stats dictionary or dictionary member is found, an issue should be
-        raised against this spec on Github, and a review process will decide on whether the stat
-        should be added to the specification or not.
+        If a need for a new stats object type or stats value within a stats object is found, an
+        issue should be raised against this spec on Github, and a review process will decide on
+        whether the stat should be added to the specification or not.
       </p>
       <p>
         A pull request for a change to this document may serve as guidance for the discussion, but
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -132,7 +132,7 @@
         </dt>
         <dd>
           This is a set of values, copied out from a monitored object at a specific moment in time.
-          It is recturned as a WebIDL "dictionary" through the GetStats API.
+          It is returned as a WebIDL dictionary through the getStats API call.
         </dd>
         <dt>
           <dfn>Stats object reference</dfn>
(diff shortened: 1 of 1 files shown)


alvestrand:
@dontcallmedom @toddreifsteck @jan-ivar please comment.

webrtc-stats.html
```diff
@@ -118,9 +113,56 @@
...
+          <p>
+            The stats object is defined by describing the stats value set it returns.
+          </p>
+        </dd>
+        <dt>
+          <dfn>Stats value set</dfn>
+        </dt>
+        <dd>
+          This is a set of values, copied out from a stats object at a specific moment in time. It
+          is recturned as a WebIDL "dictionary" through the GetStats API.
```
dontcallmedom: typo in 'recturned'

webrtc-stats.html
```diff
@@ -118,9 +113,56 @@
...
+        The basic object of the stats model is the <a>stats object</a>. The following terms are
+        defined to describe it:
+      </p>
+      <dl>
+        <dt>
+          <dfn>Stats object</dfn>
+        </dt>
+        <dd>
+          <p>
+            This represents an internal object that keeps a set of data values, together with a
```
jan-ivar: I think we can and should define this API without specifying an internal object that's not exposed anywhere, that may or may not exist (it doesn't in Firefox). Even if we can't, I find the webrtc terminology more intuitive (from [id](http://w3c.github.io/webrtc-pc/#dom-rtcstats-id)):

*"A unique id that is associated with the object that was inspected to produce this RTCStats object. Two RTCStats objects, extracted from two different RTCStatsReport objects, MUST have the same id if they were produced by inspecting the same underlying object. User agents are free to pick any format for the id as long as it meets the requirements above."*

I think most people when they hear "stats object" will think of the RTCStats object, not the "underlying object" which is only mentioned indirectly. I think we should define it that way, otherwise we risk confusing people.

webrtc-stats.html
```diff
@@ -118,9 +113,56 @@
...
+          </p>
+        </dd>
+        <dt>
+          <dfn>Stats object reference</dfn>
+        </dt>
+        <dd>
+          <p>
+            A stats object has a stable identifier "id". Other stats objects may contain references
+            to this stats object. In a <a>stats value set</a>, these references are represented by
+            a DOMString containing "id" value of the referenced stats object.
```
jan-ivar: Here's an example of where I think we get into trouble. Using the given definition, this sentence basically explains that (paraphrasing) one underlying object "references" a different underlying object using its id.

I find this an odd way to explain the act of referencing one RTCStats object using a reference found in a different RTCStats object. At no time does anything directly "reference" the internal object. In fact, the internal object appears redundant in the explanation.

webrtc-stats.html
jan-ivar: This to me is a "stats object".

webrtc-stats.html
```diff
@@ -134,12 +176,16 @@
           Guidelines for design of stats objects
         </h3>
         <p>
-          When introducing a new stats type, the following principles should be followed:
+          When introducing a new stats object, the following principles should be followed:
```
jan-ivar: I think this was better before. This part of the spec is really about designing new stats *types*, of which there may be multiple instances.

webrtc-stats.html
```diff
@@ -134,12 +176,16 @@
...
         </p>
         <ul>
           <li>An RTCStats object should correspond to an object defined in the specification it
           supports.
           </li>
+          <li>The object MUST define a new value in the "RTCStastType" enum, and MUST define the
```
jan-ivar: This is a clue that it is a type.

webrtc-stats.html
```diff
@@ -153,10 +199,9 @@
           (camelCase).
         </p>
         <p>
-          Names ending in "Id" (such as "datachannelId") are always of type DOMString and are used
-          to refer to other stats objects by the value of their "id" field; names ending in "Ids"
-          (such as "trackIds") are always of type sequence&lt;DOMString&gt;, and are used to refer
-          to a list of other stats objects.
+          Names ending in "Id" (such as "datachannelId") are always a <a>stats object
+          reference</a>; names ending in "Ids" (such as "trackIds") are always of type
```
jan-ivar: I think "stats object reference" would work with my definition, not with the one here.

webrtc-stats.html
```diff
@@ -216,16 +261,16 @@
     </section><!-- VS: getStats() remains in base spec! -->
     <section>
       <h2>
-        Maintenance procedures for stats types
+        Maintenance procedures for stats object types
```
jan-ivar: Better before.

webrtc-stats.html
```diff
@@ -216,16 +261,16 @@
...
       </h2>
       <p>
         This document, in its editors' draft form, serves as the repository for the currently
-        defined set of stats types.
+        defined set of stats object types.
```
jan-ivar: Better before.

webrtc-stats.html
```diff
@@ -216,16 +261,16 @@
...
       </p>
       <p>
-        If a need for a new stats dictionary or dictionary member is found, an issue should be
-        raised against this spec on Github, and a review process will decide on whether the stat
-        should be added to the specification or not.
+        If a need for a new stats object type or stats value within a stats object is found, an
```
jan-ivar: s/object type/type/ imho

webrtc-stats.html
```diff
@@ -216,16 +261,16 @@
...
+        issue should be raised against this spec on Github, and a review process will decide on
+        whether the stat should be added to the specification or not.
```
jan-ivar: Maybe we should define "stat"?

alvestrand:
I observe that you disagree with the overall direction.
The bugs filed came from the problem that we sometimes need to talk about a stats value changing - in a way that is a bit shorter than "the difference between the attributes of two Stats Objects that have the same "id" value". This led to the questions raised in the specified bugs.
Another approach would be:
- Say that each object in the object model of WebRTC has some associated stats values (some of the things that have stats values aren't in the object model, but if the approach makes sense, we can make our way around that)
- State that a stats object (of a specific stats type) is the snapshot of the stats values of the object at a specific moment in time (the timestamp)
I think this would be a smaller textual change, and might achieve the same goal of being able to talk about a value changing (on the stats-owning object) while still being clear that a returned value never changes.

Does that sound workable?

alvestrand:
@jan-ivar I tried for another approach - defining "monitored object" as the thing that the numbers come from, and "stats object" as the value returned. Does this work better?

webrtc-stats.html
```diff
@@ -118,51 +118,48 @@
...
+          <dfn>Stats object reference</dfn>
         </dt>
         <dd>
-          This is a set of values, copied out from a stats object at a specific moment in time. It
-          is recturned as a WebIDL "dictionary" through the GetStats API.
+          <p>
+            A monitored object has a stable identifier "id", which is reflected in all stats
+            objects produced from the monitored object. Stats objects may contain references to
+            other stats objects using this "id" value. In a <a>stats object</a>, these references
+            are represented by a DOMString containing "id" value of the referenced stats object.
```
jan-ivar: Is this maybe a good spot to mention the "xxxId" convention?

webrtc-stats.html
```diff
@@ -118,51 +118,48 @@
...
         <dd>
-          Refers to a single value within a stats value set.
+          Refers to a single value within a stats object.
         </dd>
       </dl>
       <p>
-        A stats object changes the values it contains continuously over its lifetime, but is never
-        visible through the GetStats API. A stats value set, once returned, never changes.
+        A monitored object changes the values it contains continuously over its lifetime, but is
+        never visible through the GetStats API. A stats object, once returned, never changes.
```
jan-ivar: getStats

webrtc-stats.html
```diff
@@ -118,51 +118,48 @@
...
       </p>
       <p>
         The stats API is defined in [[!WEBRTC]]. It is defined to return a collection of <a>stats
-        value set</a>s that are a subclass of the RTCStats dictionary. This API is normatively
-        defined in [[!WEBRTC]], but is reproduced here for ease of reference.
+        object</a>s, each of which is of a subclass of the RTCStats dictionary. This API is
```
jan-ivar: I think the [WebIDL terminology](https://heycam.github.io/webidl/#dfn-inherited-dictionaries) here would be "each of which is a dictionary inherited directly or indirectly from *RTCStats*."

webrtc-stats.html
```diff
@@ -244,18 +241,15 @@
...
-        </p>
-        <p>
-          When an object is closed or deleted, the <code>timestamp</code> member of the stats
-          object stops updating; it is frozen at the time when the object stopped.
+          When a monitored object is destroyed, a final stats object is produced, carrying the
+          values current at the time of destruction. This object will be returned, with a timestamp
+          reflecting the time of destruction, whenever getStats() is called, as long as the
+          PeerConnection exists. This is important in order to report consistently on short-lived
+          objects and to be able to consistently report totals over the lifetime of a
+          PeerConnection.
```
jan-ivar: @alvestrand Could you give me a link to consensus for this? I forget, sorry. FWIW I don't think this is a good idea, but for the purposes of this review, it looks like a clarification of something that was already implied, so lgtm.

webrtc-stats.html
alvestrand: Yes.

webrtc-stats.html
alvestrand: "Stats object type"?

webrtc-stats.html
alvestrand: I'll go on an archive search, but as you say, it's not necessary to resolve this to land this CL.

webrtc-stats.html
alvestrand: I like it better with "stats object types", because it's sufficiently clear that it's not a stats value type. I guess we differ.

webrtc-stats.html
alvestrand: In this particular instance, "proposal" is better than "stat".

vr000m: +1 for the use of "proposal"

webrtc-stats.html
vr000m: There are two proposals here:
1. that the getStats returns for an object that was destroyed
2. that returned object contains a snapshot of the values at the time of destruction.

If my memory serves me correctly, I believe that we did agree on 1. I cannot say we explicitly discussed 2. Although I would have expected that particular behavior.

webrtc-stats.html
```diff
@@ -118,51 +118,48 @@
...
-          <p>
-            A stats object has a stable identifier "id". Other stats objects may contain references
-            to this stats object. In a <a>stats value set</a>, these references are represented by
-            a DOMString containing "id" value of the referenced stats object.
-          </p>
-          <p>
-            The stats object is defined by describing the stats value set it returns.
-          </p>
+          This is a set of values, copied out from a monitored object at a specific moment in time.
+          It is recturned as a WebIDL "dictionary" through the GetStats API.
```
alvestrand: Oops.

alvestrand:
Harald will address Jan-Ivar's comments to the last version and then merge it.

webrtc-stats.html
jan-ivar: Ah, good point. But when would the distinction matter?

PR #142, opened 2017-01-27
Rename RTCRtpMediaStreamStats.trackId

### LINKED ISSUE #113: RTCRTPStreamStats.mediaTrackId, RTCMediaStreamStats.trackIds
We have two members referring to RTCMediaStreamTrackStats objects by ID, RTCRTPStreamStats.mediaTrackId and RTCMediaStreamStats.trackIds.

Do we want to be consistent (mediaTrackId + mediaTrackIds OR trackId + trackIds OR mediaStreamTrackId + mediaStreamTrackIds) or keep these names?

Doesn't matter much what we call these but it would be nice to avoid renaming after Chromium unflags getStats (https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/bG-3TFOZC3s).

alvestrand:
This makes the naming of the ID for tracks consistent between
MediaStreamStats and RtpMediaStreamStats.

Closes #113.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -414,7 +414,7 @@ enum RTCStatsType {
              DOMString     associateStatsId;
              boolean       isRemote = false;
              DOMString     mediaType;
-             DOMString     mediaTrackId;
+             DOMString     trackId;
              DOMString     transportId;
              DOMString     codecId;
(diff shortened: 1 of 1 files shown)


alvestrand:
@henbos @vr000m please review

PR #148, opened 2017-01-31
Explanation for "remoteSource"

### LINKED ISSUE #110: Need description for `remoteSource` member of `RTCMediaStreamTrackStats`
It's clear to me what this means, but a brief description for the uninformed would be nice.

alvestrand:
Closes #110

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1100,7 +1100,10 @@ enum RTCStatsType {
                 <dfn><code>remoteSource</code></dfn> of type <span class=
                 "idlMemberType"><a>boolean</a></span>
               </dt>
-              <dd></dd>
+              <dd>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1100,7 +1100,10 @@
                 <dfn><code>remoteSource</code></dfn> of type <span class=
                 "idlMemberType"><a>boolean</a></span>
               </dt>
-              <dd></dd>
+              <dd>
+                True if the source is remote, for instance if it is sourced from another host via
```
taylor-b: Just to confirm, does this mean that if a PeerConnection is re-transmitting a MediaStreamTrack received from *another* PeerConnection, `remoteSource` will be `true`? If so this sounds right.

vr000m: Yes, that was the idea.

PR #149, opened 2017-01-31
Remove references saying "defines an API"

### LINKED ISSUE #115: Where is the API?
https://w3c.github.io/webrtc-stats/ suggests the document defines a set of APIs, but no entry point is to be found. It sounds like some other document might define an API that ends up returning the objects described in this document, but it's all rather unclear.

alvestrand:
This should make it clear that this document only defines objects.

Closes #115

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -20,8 +20,11 @@
   <body>
     <section id="abstract">
       <p>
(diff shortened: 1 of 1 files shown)


alvestrand:
@annevk please check that this makes sense for you.

annevk:
Looks reasonable.

PR #156, opened 2017-02-02
frameWidth/frameHeight: use last decoded value

fippo:
The stats for the frameWidth and frameHeight should be for the last
decoded frame. This is for consistency with the MediaElement width/height
which does not change when there are temporarily no frames received.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1136,8 +1136,9 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Only valid for video MediaStreamTracks and represents the width of the video
-                  frame for this track.
+                  Only valid for video MediaStreamTracks and represents the width of the last
+                  processed video frame for this track. Before the first frame is processed this
+                  attribute is missing.
                 </p>
               </dd>
               <dt>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1136,8 +1136,8 @@
               </dt>
               <dd>
                 <p>
-                  Only valid for video MediaStreamTracks and represents the width of the video
-                  frame for this track.
+                  Only valid for video MediaStreamTracks and represents the width of the last
+                  decoded video frame for this track.
```
alvestrand: I think this should be "last processed video frame" - for locally sourced tracks, there may not be any decoding involved (at least for RGB cameras).

Should we also say that before the first frame is processed, this attribute is missing?

fippo:
Changed + squashed. WRT to missing attributes, don't we have the same issue elsewhere, e.g. with framesDropped and framesCorrupted? Arguably those should be set to 0 once the first frame is processed and not be missing

alvestrand:
framesDropped and framesCorrupted make sense to have as zero before the first frame is processed, but there's no sensible value we could use for widht and height.

Issue #105, opened 2016-12-15
Why is ssrc a DOMString?


taylor-b:
Every instance of `ssrc` in webrtc is an `unsigned long`. Is there a reason for it to be a DOMString in webrtc-stats? If so, a brief explanation would be helpful, with a description of what non-numeric values can be expected.

jesup:
SSRCs by definition in RTP are 32-bit unsigned values.  While you could translate them into hex (upper or lower?  What if people compare against SDP, and the case is different?  etc), that's kinda silly.

They should be 32 bits, unsigned.

alvestrand:
We should convert it in the spec to 32 bit unsigned integers.

alvestrand:
All the comments seem to indicate consensus here.

alvestrand:
Changed to unsigned int.

PR #157, opened 2017-02-02
Changes type of RTCRTPStreamStats.ssrc from string to unsigned long.

henbos:
Resolves issue #105. PTAL and merge @alvestrand

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -435,7 +435,8 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  32-bit integer value per [[RFC3550]].
+                  The 32-bit unsigned integer value per [[RFC3550]] used to identify the source of
+                  the stream of RTP packets that this stats object concerns.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -430,8 +430,8 @@ enum RTCStatsType {
             <dl data-link-for="RTCRTPStreamStats" data-dfn-for="RTCRTPStreamStats" class=
             "dictionary-members">
               <dt>
-                <dfn><code>ssrc</code></dfn> of type <span class=
-                "idlMemberType"><a>DOMString</a></span>
+                <dfn><code>ssrc</code></dfn> of type <span class="idlMemberType"><a>unsigned
+                long</a></span>
               </dt>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -430,8 +430,8 @@
             <dl data-link-for="RTCRTPStreamStats" data-dfn-for="RTCRTPStreamStats" class=
             "dictionary-members">
               <dt>
-                <dfn><code>ssrc</code></dfn> of type <span class=
-                "idlMemberType"><a>DOMString</a></span>
+                <dfn><code>ssrc</code></dfn> of type <span class="idlMemberType"><a>unsigned
+                long</a></span>
               </dt>
               <dd></dd>
```
alvestrand: While you're at it, can you fill in the explanation (the part between <dd> and </dd>) with "32-bit integer value per [[RFCxxxx]]"?

webrtc-stats.html
```diff
@@ -410,7 +410,7 @@
         </h3>
         <div>
           <pre class="idl">dictionary RTCRTPStreamStats : RTCStats {
-             DOMString     ssrc;
+             unsigned long ssrc;
```
alvestrand: Please align with the others, I don't think the formatting will re-align your webIDL.

henbos: Done

webrtc-stats.html
henbos: Done

webrtc-stats.html
```diff
@@ -430,10 +430,14 @@
...
               <dt>
-                <dfn><code>ssrc</code></dfn> of type <span class=
-                "idlMemberType"><a>DOMString</a></span>
+                <dfn><code>ssrc</code></dfn> of type <span class="idlMemberType"><a>unsigned long
+                </a></span>
               </dt>
-              <dd></dd>
+              <dd>
+                <p>
+                  32-bit integer value per [[RFC3550]].
```
jan-ivar: unsigned

henbos: Description updated.
(The spec says "identified by a 32-bit numeric SSRC identifier carried in the RTP header", based on it it is not clear to me whether it should be signed, unsigned or just a series of bits, but its appendix has sample code that use uint32_t :) )

PR #164, opened 2017-02-25
Adding remoteTimestamp to RTCRtpStreamStats.

taylor-b:
Since `timestamp` always refers to the locally measured timestamp,
as decided by the working group, this is where the remote timestamp can
be found.

See: https://github.com/w3c/webrtc-pc/issues/729

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -416,7 +416,6 @@ enum RTCStatsType {
              unsigned long       ssrc;
              DOMString           associateStatsId;
              boolean             isRemote = false;
-             DOMHighResTimeStamp remoteTimestamp;
              DOMString           mediaType;
              DOMString           trackId;
              DOMString           transportId;
@@ -462,26 +461,6 @@ enum RTCStatsType {
                   <code>false</code> indicates that the statistics are measured locally, while
                   <code>true</code> indicates that the measurements were done at the remote
                   endpoint and reported in an RTCP RR/XR.
-                </p>
-              </dd>
-              <dt>
-                <dfn><code>remoteTimestamp</code></dfn> of type <span class=
-                "idlMemberType"><a>DOMHighResTimeStamp</a></span>
-              </dt>
-              <dd>
-                <p>
-                  Present if <code>isRemote</code> is <code>true</code>,
-                  <code>remoteTimestamp</code>, of type
-                  <code>DOMHighResTimeStamp</code> [[!HIGHRES-TIME]],
-                  represents the remote timestamp at which these statistics
-                  were sent by the remote endpoint. This differs from
-                  <code>timestamp</code>, which represents the time at which
-                  the statistics were generated or received by the local
-                  endpoint. The <code>remoteTimestamp</code>, if present, is
-                  derived from the NTP timestamp in an RTCP SR packet, which
-                  reflects the remote endpoint's clock. That clock may not be
-                  synchronized with the local clock. The time is relative to
-                  the UNIX epoch (Jan 1, 1970, UTC).
                 </p>
               </dd>
               <dt>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -464,6 +465,23 @@
...
               <dt>
+                <dfn><code>remoteTimestamp</code></dfn> of type <span class=
+                "idlMemberType"><a>DOMHighResTimeStamp</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Present if <code>isRemote</code> is <code>true</code>,
+                  <code>remoteTimestamp</code>, of type
+                  <code>DOMHighResTimeStamp</code> [[!HIGHRES-TIME]],
+                  represents the remote timestamp at which these statistics
```
alvestrand: I'd suggest adding "This timestamp, if present, is derived from the NTP timestamp in an RTCP SR packet, which reflects the remote side's clock. That clock may not be synchronized with the local clock."

I think "derived from" gives enough wiggle room to allow for other computations if the report arrived in RTCP that didn't contain an SR (for instance in RR-only packets).

taylor-b: Sounds good, done.

alvestrand:
lgtm

webrtc-stats.html
vr000m: I think we should be explicit that this value only present on inbound streams when isRemote is true.

webrtc-stats.html
```diff
@@ -464,6 +465,26 @@
...
+                  were sent by the remote endpoint. This differs from
+                  <code>timestamp</code>, which represents the time at which
+                  the statistics were received by the local endpoint. This
+                  timestamp, if present, is derived from the NTP timestamp in
```
vr000m: s/This timestamp/The remoteTimestamp/

webrtc-stats.html
vr000m: Do we want the wiggle room for timestamps from RTCP RR/XRs? otherwise I would suggest moving this to RTCOutboundRTPStreamStats, i.e., being consistent with the data from the RTCP SRs.

taylor-b: > I think we should be explicit that this value only present on inbound streams when isRemote is true.

What wording would you suggest? Change "Present if" to "Present if and only if"?

> Do we want the wiggle room for timestamps from RTCP RR/XRs?

That's what was decided, I believe.

webrtc-stats.html
taylor-b: Done

alvestrand:
Varun will take another look at the definition of timestamp in SR - when it is present - and whether it makes sense to compute timestamp for other reports based on the SR, and if so how.

webrtc-stats.html
```diff
@@ -464,6 +465,26 @@
...
+              </dt>
+              <dd>
+                <p>
+                  Present if <code>isRemote</code> is <code>true</code>,
+                  <code>remoteTimestamp</code>, of type
+                  <code>DOMHighResTimeStamp</code> [[!HIGHRES-TIME]],
+                  represents the remote timestamp at which these statistics
+                  were sent by the remote endpoint. This differs from
+                  <code>timestamp</code>, which represents the time at which
+                  the statistics were received by the local endpoint. The
```
vr000m: the statistics were _generated_ or received by the local endpoint.

taylor-b: Done

alvestrand:
After discussion between Varun and Harald: We should only add RemoteTimestamp to remote outbound (which means moving it to RTPOutboundStreamStats), since that's the one where we are guaranteed to have an SR.
The extension to inbound should have its own bug, in case people see a strong need for that.
@vr000m  will file a bug on the inbound.

taylor-b:
Ok, moved to RTCOutboundRTPStreamStats

Issue #63, opened 2016-09-22
Missing consent interval in RTCIceCandidatePairStats


vr000m:
Currently, the consent interval is different for each browser, it would be nice to have a way to know what value is used by the browser.

@aboba @taylor-b comments? 

TPAC16 decision:
Go ahead and create PR.

taylor-b:
As Peter described in Lisbon, we don't really have a fixed consent interval, because for us, "consent checks" serve a double purpose as connectivity checks, and as a way of evaluating the RTT. So we may send them more or less frequently based on the state of the connection.

We still could probably expose some interval in the stats. But it would need to be variable.

vr000m:
That interval, although variable would be a sufficient. 

I'd probably query this when the ICE goes to the disconnected state, mainly to determine the duration of loss in connectivity.

alvestrand:
So the useful number would be the interval between the last two connectivity checks sent?

vr000m:
Yes. The interval between the last two checks. And perhaps a long term average, sum of intervals should be sufficient as we can divide that by the (total number of checks-1)

alvestrand:
Hm. If connectivity is lost, you won't get duration of loss in connectivity from the interval between the last two checks; both were sent before connectivity was lost.
Two interesting numbers are:
- Time of last successful check (which places an upper bound on how long connectivity has been lost)
- Sum and count of intervals

A third interesting number might be the time of the last check sent (which would be a failed one in the case of lost connectivity).

vr000m:
These two would work for me.

alvestrand:
will you write the PR to add lastCheck and sumCheckInterval?

vr000m:
Yes.

PR #165, opened 2017-02-27
Remove separation of received consent and connectivity requests.

### LINKED ISSUE #96: Not clear how to differentiate between received connectivity checks and consent requests
This issue applies to RTCIceCandidatePairStats.requestsReceived and consentRequestsReceived. The actual STUN packet looks the same, so how should an implementation differentiate between the two?

For *sent* requests, we can say that any binding request sent before a response is received is a connectivity check, and any request sent after is a consent request. But we can't do this for received binding requests, because we don't know if the remote peer received a response or not.

alvestrand:
It seems to be impossible for a receiver of such a request to tell
why it was sent, so we should not try to differentiate on either
incoming requests, outgoing responses or incoming responses.

Fixes #96

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -433,8 +433,8 @@ enum RTCStatsType {
             <dl data-link-for="RTCRTPStreamStats" data-dfn-for="RTCRTPStreamStats" class=
             "dictionary-members">
               <dt>
-                <dfn><code>ssrc</code></dfn> of type <span class="idlMemberType"><a>unsigned long
-                </a></span>
+                <dfn><code>ssrc</code></dfn> of type <span class="idlMemberType"><a>unsigned
+                long</a></span>
               </dt>
               <dd>
                 <p>
@@ -561,7 +561,7 @@ enum RTCStatsType {
                 <p>
                   The definition of QP value depends on the <a>codec</a>; for VP8, the QP value is
                   the value carried in the frame header as the syntax element "y_ac_qi", and
-                  defined in [[RFC6386]] section 19.2. Its range is 0..127.
+                  defined in [[RFC6386]] section 19.2. Its range is 0..127. 
                   <!-- Need appropriate references for VP9 and H.264 -->
                 </p>
                 <p>
@@ -1630,12 +1630,8 @@ enum RTCStatsType {
              unsigned long long            requestsSent;
              unsigned long long            responsesReceived;
              unsigned long long            responsesSent;
-             unsigned long long            retransmissionsReceived;
              unsigned long long            retransmissionsSent;
-             unsigned long long            consentRequestsReceived;
              unsigned long long            consentRequestsSent;
-             unsigned long long            consentResponsesReceived;
-             unsigned long long            consentResponsesSent;
 };</pre>
           <section>
             <h2>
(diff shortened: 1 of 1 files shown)


vr000m:
Aren't connectivity checks done before connectivity is established and consent after? is the issue that the endpoints do not know when the connectivity was setup?

i.e., after the first RTP packet is sent or received on a particular ICE Candidate Pair?

vr000m:
If we remove this does this affect issue: https://github.com/w3c/webrtc-stats/issues/63

alvestrand:
The one sending the requests knows why he sends it, so it makes sense to separate the counters on the sender. This PR only deals with removing the distinction at the responder.

alvestrand:
Pinging @taylor-b as well.

vr000m:
There was a request and response counters in RFC7982, https://tools.ietf.org/html/rfc7982

taylor-b:
@vr000m Ah. Then we should probably leave retransmissionsReceived, which would be defined as "number of received requests with TRANSACTION_TRANSMIT_COUNTER > 1".

vr000m:
@taylor-b yes and the transaction id will be retransmitted (increasing transaction transmit counters) until a response is received,

alvestrand:
@vr000m @taylor-b PTAL.
In processing this, I realized that the text didn't specify whether retransmission counters covered both requests and responses or just requests; I flipped a coin and made them count requests only.

webrtc-stats.html
```diff
@@ -1749,11 +1746,11 @@
...
                 <p>
                   Represents the sum of all round trip time measurements in seconds since the
-                  beginning of the session, based on both STUN connectivity check
-                  [[!STUN-PATH-CHAR]] responses (responsesReceived) and consent [[!RFC7675]]
-                  responses (consentResponsesReceived). The average round trip time can be computed
-                  from <code>totalRoundTripTime</code> by dividing it by
-                  (<code>responsesReceived</code> + <code>consentResponsesReceived</code>).
+                  beginning of the session, based on STUN connectivity check [[!STUN-PATH-CHAR]]
+                  responses (responsesReceived), including those that reply to requests that are
+                  sent in order to verify consent [[!RFC7675]] responses. The average round trip
```
taylor-b: nit: the word "responses" at the end of this sentence seems redundant.

webrtc-stats.html
```diff
@@ -1763,7 +1760,8 @@
               <dd>
                 <p>
                   Represents the latest round trip time measured in seconds, computed from both
```
taylor-b: Remove the word "both"?

webrtc-stats.html
```diff
@@ -1836,7 +1838,10 @@
               </dt>
               <dd>
                 <p>
-                  Represents the total number of connectivity check retransmissions received.
+                  Represents the total number of connectivity check request retransmissions
```
taylor-b: I'd suggest counting both requests and responses here. If there's a need to differentiate retransmitted requests and responses, this could be split into two fields, but I can't think of such a need.

webrtc-stats.html
alvestrand: Agree.

webrtc-stats.html
alvestrand: Hm. This is a retransmissions *received* count, which means that the number of retransmitted requests received will reflect the number of retransmissions that the sender thinks is lost, and is driven by the remote end's desire for information.
The number of retransmitted responses received (strictly: the number of responses received that are in response to a retransmitted requests - responder side doesn't ever initiate retransmissions) is received at the side that initiates the requests, which means that it is always strictly less than the number of retransmitted requests sent (unless duplication happens), a number the receiver doesn't have access to.
I'm not sure either is very useful for diagnostics, but I'm very sure they're different.

taylor-b: > responder side doesn't ever initiate retransmissions

Oh, I think you're right. I was confused about the "Resp" field in RFC7982. In that case this looks good to me.

PR #167, opened 2017-02-28
moving roundTripTime from outbound to inbound

vr000m:
resolves issue #74 and #78.


vr000m:
@alvestrand PTAL, merge if okay.

Issue #77, opened 2016-11-01
Discrepancies between RTCRtpCodecParameters and RTCCodecStats naming


taylor-b:
RTCCodecStats has "codec" and "parameters" whereas RTCRtpCodecParameters has "mimeType" and "sdpFmtpLine" which seem to mean the same things.

If these are indeed expected to contain the same information, it would be nice if the names aligned. I slightly prefer the RTCRtpCodecParameters names; they seem more descriptive.

alvestrand:
We think this was introduced at TPAC in Sapporo. Have to check the minutes.
The CodecParameter names seem very SDP-linked.

vr000m:
@jan-ivar, @aboba Has someone implemented this? and what did you use?

alvestrand:
We also need to check implementations.

jan-ivar:
@vr000m Firefox has implemented neither (there's some webidl in place, but nothing backing it).

PR #168, opened 2017-02-28
aligning codec types with webrtc-pc

vr000m:
Fixing issues #77

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -637,7 +637,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  The a=fmtp line in the SDP corresponding to the codec, as defined by [[!JSEP]] (section 5.7.).
+                  The a=fmtp line in the SDP corresponding to the codec, i.e., after the colon following the PT. This defined by [[!JSEP]] in Section 5.7.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -583,10 +583,10 @@ enum RTCStatsType {
         <div>
           <pre class="idl">dictionary RTCCodecStats : RTCStats {
              unsigned long payloadType;
-             DOMString     codec;
+             DOMString     mimeType;
              unsigned long clockRate;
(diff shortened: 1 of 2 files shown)


vr000m:
One unrelated change was pushed for a reference.

webrtc-stats.html
```diff
@@ -632,12 +632,12 @@
...
               </dd>
               <dt>
-                <dfn><code>parameters</code></dfn> of type <span class=
+                <dfn><code>sdpFmtpLine</code></dfn> of type <span class=
                 "idlMemberType"><a>DOMString</a></span>
               </dt>
               <dd>
                 <p>
-                  From the SDP description line.
+                  The a=fmtp line in the SDP corresponding to the codec, as defined by [[!JSEP]] (section 5.7.).
```
alvestrand: Need to say where in the a=fmtp line this starts (after the colon following the PT).

webrtc-stats.js
```diff
@@ -94,17 +94,13 @@ var respecConfig = {
...
-            href:     "https://tools.ietf.org/html/draft-ietf-xrblock-independent-burst-gap-discard",
-            authors:  [
-                "Varun Singh",
-                "Colin Perkins",
-                "Alan Clark",
-                "Rachel Huang"
-            ],
-            status:   "Internet Draft",
-            publisher:  "IETF"
+        "JSEP": {
```
alvestrand: Check how webrtc-pc references JSEP, and do the same thing.

alvestrand:
Varun to address comments and then merge.
Note: Renames will cause more trouble once Chrome has released, so this may be the last easy rename.

webrtc-stats.js
vr000m: the sentence and reference here is verbatim from webrtc-pc

webrtc-stats.html
taylor-b: Is that something we'd already decided? It's one of the virtual interim topics.

vr000m: Yes, agreed almost a year ago in Sapporo.

Issue #83, opened 2016-11-12
Expose protocol used to communicate with TURN server


taylor-b:
For TURN candidates, it would be useful to see the protocol used to communicate with the TURN server. Either UDP, TCP, or TLS over TCP. The `protocol` field that currently exists is just the protocol of the allocated candidate.

If we add this, it may be worth adding to `RTCIceCandidate` as well. Though it's in a different category from the other attributes, in that it's not something that appears in the candidate SDP.

alvestrand:
From https://tools.ietf.org/html/rfc5928#section-3 it seems that a TURN client-server link can indeed resolve to any protocol (if unspecified), so you can't tell from the URI scheme, which was my first thought.

I'm not clear on when in the process of resolving a candidate this field can be populated - can you help in describing that?

fippo:
this would basically have to expose the local type preference (which is priority >> 24) and then turn that into the transport protocol using a browser-specific mapping.

But wasn't the URL the candidate was gathered from for exactly that purpose? I agree though that this is a more direct solution.

alvestrand:
It turns out a turn url without a transport modifier can end up using any protocol depending on dns config and local capabilities so no, in general you cannot know from the url. Complex stuff!

fippo:
pinging @jan-ivar -- iirc you have a mozlocaltransport already?

jan-ivar:
Yes we have `mozLocalTransport`. You can see it in a dump [here](https://jsfiddle.net/moz71L7h/).

I found [this old comment](https://bugzilla.mozilla.org/show_bug.cgi?id=970690#c8) on it.

taylor-b:
@alvestrand 
> I'm not clear on when in the process of resolving a candidate this field can be populated - can you help in describing that?

Well, it needs to happen before the candidate is allocated ("gathered"), so there's never a situation where it could be unknown at the time of appearing in a stats report.

Of course, this field would only be present for local candidates, not remote candidates.

PR #169, opened 2017-02-28
fixes issue with TURN URL protocol

vr000m:
Fixes issue #83

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1500,6 +1500,7 @@ enum RTCStatsType {
              RTCIceCandidateType      candidateType;
              long                     priority;
              DOMString                url;
+             DOMString                turnUrlProtocol;
              boolean                  deleted = false;
 };</pre>
           <section>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1500,6 +1500,7 @@
              RTCIceCandidateType      candidateType;
              long                     priority;
              DOMString                url;
+             DOMString                turnUrlProtocol;
```
taylor-b: Even though this protocol is included in the URL, I don't think "url" needs to be in the variable name. Plus, as Harald pointed out, it may not necessarily be in the URL. I'd suggest just "turnProtocol" or "relayProtocol".

webrtc-stats.html
```diff
@@ -1558,6 +1559,15 @@
                 </p>
               </dd>
               <dt>
+                <dfn><code>turnUrlProtocol</code></dfn> of type <span class=
+                "idlMemberType"><a>DOMString</a></span>
+              </dt>
+              <dd>
+                <p>
+                  It is the protocol used by the endpoint to communicate with the TURN server, this is only present for local candidates. Valid values for the TURN URL protocol is one of <code>udp</code>, <code>tcp</code>, or <code>tls</code>.
```
taylor-b: > ", this is only present for..."

This is a separate sentence fragment; should separate with a semicolon or period.

Would also be good to break this into multiple lines.

webrtc-stats.html
alvestrand: relayProtocol sounds better.

vr000m: I wonder if the name will cause confusion with the protocol field above. I do not have a strong opinion.

PR #181, opened 2017-03-13
Add section on obsoleted stats.

### LINKED ISSUE #174: Need advice for handling obsolete stats
If stats are renamed or declared useless, we need to keep the names around for reference.
There should be advice in the spec on how to do this.

alvestrand:
Closes #174

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -273,35 +273,56 @@
       <h2>
         Maintenance procedures for stats object types
       </h2>
-      <p>
-        This document, in its editors' draft form, serves as the repository for the currently
-        defined set of stats object types.
-      </p>
-      <p>
-        If a need for a new stats object type or stats value within a stats object is found, an
-        issue should be raised against this spec on Github, and a review process will decide on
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -2096,6 +2117,36 @@
     </section>
     <section>
       <h2>
+        Obsolete stats
+      </h2>
+      <p>
+        This section's content is an example.
```
vr000m: We have a concrete example, rtt got deprecated for roundTripTime, handle that as a separate issue?

alvestrand:
Harald will switch example in the "obsoleted" section to use rtt (which was renamed and moved).
Moved in #167 - renamed in #95 
We should say that we don't keep track of stuff that was never implemented - if we were pedantic, we would report things that were in drafts, but we don't want to do that.

alvestrand:
@vr000m still LGTM?

vr000m:
The #167 is related to the RTP RTT, we need another entry for RTCInboundRTPStreamStats.

In the case of #167, we changed the units from ms to seconds.

PR #170, opened 2017-02-28
Related to outgoing and incoming bitrate estimates on a candidate pair

vr000m:
corresponds to issues reported in #86, #97, and #112. 

related to outgoing and incoming bitrate estimates on a candidate pair.

Issue #112, opened 2016-12-15
What should `availableOutgoingBitrate`/`availableOutgoingBitrate` be for candidate pairs not in use?


taylor-b:
If a candidate pair isn't currently selected by ICE, it's not clear what `availableOutgoingBitrate`/`availableIncomingBitrate` should be. Should they be the previous estimates, if the pair was used in the past? Or should they be left undefined? If it's the latter, then maybe these should be moved up to `RTCIceTransportStats`.

Also... What if audio/video use different DSCP values and have different bandwidth estimates? Which one should go in `availableOutgoingBitrate`?

alvestrand:
When I added this for candidate pairs, I was thinking that we could use this to detect the case where a low-bandwidth connection is used when a high-bandwidth connection was previously used. The only way I can think of where this makes sense is to have the last estimated bitrate when it was used in the past, and have the stat be absent for pairs that never were used for live traffic.

vr000m:
I agree with @alvestrand's suggestion for the updating the last estimate with candidate-pairs that were previously used. And the key missing for the pairs that were never used.

About the DSCP,  I think it is more important to know the overall estimate, hence, adding all the media estimates together would make sense. The tracks have relative priorities, and an implementation can compare the outgoing bitrates of the individual streams with the 

Although, I think we should expose the markings (codepoint value) in the appropriate inbound/outbound track.

alvestrand:
Commentary in offline discussion is that the bandwidth data can go stale pretty quickly (walk-out-the-door problem), so it may be wiser / less misleading to just omit the bandwidth estimate on pairs that aren't in use.

taylor-b:
In that case let's revisit the question: Should these move to RTCIceTransportStats, since there will only be one value per transport?

Issue #97, opened 2016-12-09
Remove availableIncomingBitrate?


alvestrand:
In the case of sender-side bitrate control, there is no way for the receiver to estimate the available bitrate.
Since it can't be implemented consistently across all bandwidth control algorithms, its value seems limited; should we delete it?

Ping @aboba

vr000m:
While this is currently true, do we want to limit future implementations from experimenting with receiver side estimation? 

Implementations that do not have a receive side estimate can skip filling in the key.

alvestrand:
We need to add to description that it's only relevant if r-side estimation is implemented.

aboba:
@alvestrand Having bandwidth availability reported only from the sender side is ok.

aboba:
@Alvestrand We are currently using the receiver bandwidth estimate, so I'd prefer it be kept.

vr000m:
@aboba see text in PR #182, does it work for you?

Issue #86, opened 2016-11-29
Design consideration for inactive SSRCs and inactive transports


vr000m:
In #68, it was raised that:

+ ICE restarts will cause new candidate pairs to become active and previously active ones to become inactive, if the call last long, the number of candidate pairs will accumulate.
+ Similarly if SSRCs change, previously active SSRCs will still be reported upon.

Is there something that we can do better?

vr000m:
Copying my comment from the issue #68:

With the frozen timestamps, an app can filter out the older candidates, assuming the app has some idea about how often it queries or when it last queried the stats.

Although one concern I do have (and perhaps could be solved in a different way) is if the ice candidate is currently inactive and was active in the past, we do not really know when it was last active. We perhaps need a lastPacketSentTimestamp, lastPacketReceivedTimestamp or similar

alvestrand:
Would this timestamp be valid for only local candidates, or would we also update the remote candidate of a pair? Or perhaps it's on candidate pairs? (a candidate can't be active without being part of a pair).

vr000m:
I will add the timestamps to the candidate pair stats.

PR #182, opened 2017-03-15
Bandwidth estimations again (Issue 97 redux)

vr000m:
corresponds to issues reported in #86, #97, and #112. related to outgoing and incoming bitrate estimates on a candidate pair.

Tries to undo the git rebase hassle in #170.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1862,7 +1862,6 @@ enum RTCStatsType {
                 <p>
                   Represents the timestamp at which the last packet
                   was sent on this particular candidate pair.
-                  It is of the type <code>DOMHighResTimeStamp</code> [[!HIGHRES-TIME]].
                 </p>
               </dd>
               <dt>
@@ -1873,7 +1872,6 @@ enum RTCStatsType {
                 <p>
                   Represents the timestamp at which the last packet
                   was received on this particular candidate pair.
-                  It is of the type <code>DOMHighResTimeStamp</code> [[!HIGHRES-TIME]].
                 </p>
               </dd>
               <dt>
@@ -1914,13 +1912,12 @@ enum RTCStatsType {
                   in bits per second and the bitrate is calculated over a 1 second window.
                 </p>
                 <p>
-                  Implementations that do not calculate a sender-side estimate should leave this undefined.
-                  Additionally, the value should be undefined for candidate pairs that were never used.
+                  Implementations that do not calculate a sender-side estimate MUST leave this undefined.
+                  Additionally, the value MUST be undefined for candidate pairs that were never used.
                   For pairs in use, the estimate is normally no lower than the bitrate for
-                  the packets sent at <dfn>lastPacketSentTimestamp</dfn>, but might be higher.
+                  the packets sent at <code>lastPacketSentTimestamp</code>, but might be higher.
                   For candidate pairs that are not currently in use but were used before,
-                  implementations can use the last calculated estimate provided that no further
-                  packets are sent.
+                  implementations MUST return undefined.
                 </p>
               </dd>
               <dt>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1889,6 +1913,15 @@
...
                   in bits per second and the bitrate is calculated over a 1 second window.
                 </p>
+                <p>
+                  Implementations that do not calculate a sender-side estimate should leave this undefined.
+                  Additionally, the value should be undefined for candidate pairs that were never used.
+                  For pairs in use, the estimate is normally no lower than the bitrate for
+                  the packets sent at <dfn>lastPacketSentTimestamp</dfn>, but might be higher.
+                  For candidate pairs that are not currently in use but were used before,
+                  implementations can use the last calculated estimate provided that no further
+                  packets are sent.
```
taylor-b: It's not clear to me what this last sentence means. Does "implementations can use the last calculated estimate" mean "when a candidate pair is not in use, implementations should return the last calculated estimate instead of returning nothing"? Also, what does the "provided that no further packets are sent" part mean? ICE consent checks may be sent even if the candidate pair is not "in use" (assuming "in use" means "in use for media packets").

webrtc-stats.html
```diff
@@ -1902,6 +1935,15 @@
                   or UDP. It is similar to the TIAS defined in [[!RFC3890]], i.e., it is measured
                   in bits per second and the bitrate is calculated over a 1 second window.
                 </p>
+                <p>
+                  Implementations that do not calculate a receiver-side estimate should leave this undefined.
+                  Additionally, the value should be undefined for candidate pairs that were never used.
+                  For pairs in use, the estimate is normally no lower than the bitrate for
+                  the packets sent at <dfn>lastPacketReceivedTimestamp</dfn>, but might be higher.
```
vr000m: nice catch! will fix.

webrtc-stats.html
vr000m: Assuming the case that the pair was active and now inactive, 
1.  no packets (including consent packet) being sent/received, use last estimate, we know when that last estimate was made because of the last packet timestamp.
2. no media packets are being sent but consent packets, return undefined.

question, would doing both make sense or just the case 2.

taylor-b: Doing both seems like it would be racy. If a candidate pair is immediately discarded, then `availableOutgoingBitrate` is defined, but if it's still used for a while to send consent checks, it's undefined?

Could we just add `lastMediaPacketSentTimestamp` to resolve this?

vr000m: Is it even possible to have an bitrate estimate if no media packets are sent. 

I suppose the best solution would be to say if  media or data packets are being sent then have an estimate, else undefined.

webrtc-stats.html
```diff
@@ -1853,6 +1855,28 @@
...
               <dt>
+                <dfn><code>lastPacketSentTimestamp</code></dfn> of type <span class=
+                "idlMemberType"><a>DOMHighResTimeStamp</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the timestamp at which the last packet
+                  was sent on this particular candidate pair.
+                  It is of the type <code>DOMHighResTimeStamp</code> [[!HIGHRES-TIME]].
+                </p>
```
alvestrand: The last sentence merely repeats the type, which is already stated above. Please remove it.

If you want a ref for DOMHighResTimeStamp, put it in the terminology section.

webrtc-stats.html
```diff
@@ -1889,6 +1913,15 @@
                   or UDP. It is similar to the TIAS defined in [[!RFC3890]], i.e., it is measured
                   in bits per second and the bitrate is calculated over a 1 second window.
                 </p>
+                <p>
+                  Implementations that do not calculate a sender-side estimate should leave this undefined.
```
alvestrand: I'd prefer MUST. There is no circumstance wherein it makes sense to give a value for this member if you don't calcluate it.

webrtc-stats.html
alvestrand: All estimation algorithms start off at some initial value, so internally, there is some estimate (likely made with no real world input). It doesn't make much sense to expose this number.

It makes sense to me to only expose this number after we have some faith in it (which would be after sending media for a while), but I'm not sure how to phrase that.

alvestrand:
Discussed at meeting: Make bandwidth estimate null for all not-currently-used pairs.

vr000m:
@alvestrand PTAL. Merge if ready.

PR #184, opened 2017-03-21
Add RTCOutboundRTPStreamStats.totalEncodeTime

### LINKED ISSUE #150: Stat for how much time it takes to encode video
The non-standardized Chromium getStats contains the following stats:
```
// The average encode time in ms (video).
ssrc.googAvgEncodeMs

ssrc.googEncodeUsagePercent
```
These can be used to estimate "how fast the fan is spinning" and be used as an indicator that you might want to to lower capture resolution.

Is this useful? Should we standardize it or something similar?
Averages are discouraged, and time is typically expressed as double seconds. How about a `double totalEncodingTime` in RTCMediaStreamTrackStats that is the total time, in seconds, that has been spent on encoding video for that track?

henbos:
Fixes #150


webrtc-stats.html
```diff
@@ -1047,6 +1048,20 @@
                   encoded for this RTP media stream.
                 </p>
               </dd>
+              <dt>
+                <dfn><code>totalEncodeTime</code></dfn> of type <span class="idlMemberType"><a>
+                long</a></span>
```
henbos: This is supposed to say double, will fix before merge.

webrtc-stats.html
```diff
@@ -976,7 +976,8 @@
              unsigned long long bytesSent;
              double             targetBitrate;
              unsigned long      framesEncoded;
-};</pre>
+             double             totalEncodeTime;
+          };</pre>
```
henbos: (I thought the }; not being aligned was a mistake but I find non-indented }; all over the place? Should I not fix the indentation?)

alvestrand:
Henbos will merge after fixing long -> double.

vr000m:
LGTM

webrtc-stats.html
vr000m: we can do the indentation fix in a later PR, otherwise merge conflicts with existing PRs.

webrtc-stats.html
alvestrand: @henbos please do!

PR #185, opened 2017-03-21
Added RTCInboundRTPStreamStats sample counters.

### LINKED ISSUE #152: Stat for how many audio stream packets are expanded when packets are lost (and lost and the user is speaking)
The non-standardized Chromium getStats contains the following stat:
```
// Fraction of packets that are expanded (synthesized) when we detect
// that the user is speaking.
ssrc.googSpeechExpandRate
```
What exactly does this mean? Is this useful? Should we standardize it or something similar? Sums and counters are preferred over rates.
RTCInboundRTPStreamStats already have various packet counters. Should we add another `unsigned long packetsSpeechExpanded`?

henbos:
Including
```
    unsigned long long totalSamples;
    unsigned long long concealedSamples;
    unsigned long long concealedAudibleSamples;
    unsigned long long concealmentEvents;
```

Fixes #152.


webrtc-stats.html
```diff
@@ -1202,6 +1202,10 @@
              double              audioLevel;
              double              echoReturnLoss;
              double              echoReturnLossEnhancement;
+             unsigned long long  totalSamples;
```
alvestrand: samplesReceived? (Also add samplesSent?)

webrtc-stats.html
```diff
@@ -1400,6 +1404,56 @@
...
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>concealedAudibleSamples</code></dfn> of type <span class=
+                "idlMemberType"><a>unsigned long long</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Only present for inbound audio tracks. The total number of concealed audio samples
+                  during audible portions of the stream. Audible means that the received audio is
```
alvestrand: No explanation of what "concealed" means here.
I think it should say "Number of samples that were played out based on synthesized data and not based on incoming data".

This shouldn't include comfort noise played out when CN is used, I think.

alvestrand:
@jesup @hlundin can you review this? We're particularly concerned if we have concealedAudibleSamples defined clearly enough for interoperability.

webrtc-stats.html
henbos: Done.

webrtc-stats.html
henbos: Updated definition of "concealed" and referencing that definition (conceleadSamples) from concealedAudibleSamples. Audibility still defined the same way.

webrtc-stats.html
```diff
@@ -1413,6 +1418,68 @@
...
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>concealedAudibleSamples</code></dfn> of type <span class=
+                "idlMemberType"><a>unsigned long long</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Only present for inbound audio tracks. The total number of concealed audio samples
+                  (see <a>concealedSamples</a>) that was played out during an audible portions of the
```
hlundin: portions -> portion

hlundin: portions -> portion

vr000m:
@jesup @jan-ivar: need your feedback on `concealedAudibleSamples`, is the definition clear enough? is this something that you could implement by reading the spec? since we do not have a citable reference.

vr000m:
@henbos Maybe another way forward is to make a new issue for concealedAudioSamples and a new PR, i.e., remove concealedAudioSamples from here and discuss this separately. 

Currently the totalConcealedSamples includes the concealedAudioSamples. So that we do not hold up the other metrics in the PR.

jan-ivar:
`concealedAudibleSamples` seems fuzzy. What's the use-case?

vr000m:
@henbos  will remove concealedAudibleSamples from PR and raise a separate issue for this.

vr000m:
@jan-ivar what about the other metrics, are you okay with those. So that we can move forward with the PR.

jan-ivar:
@padenot do these look good to you?

My only comment would be that it's not very clear that these are not present in the remote inbound/outbound dictionaries, but that should be solved by letting https://github.com/w3c/webrtc-stats/pull/191 merge first.

henbos:
I removed concealedAudibleSamples from this PR to add it to a follow-up (follow-up not created until this has merged).

vr000m:
@henbos this needs to be rebased because of the changes made by @jan-ivar

jan-ivar:
Yeah the subject confused me, saying these were in RTCInboundRTPStreamStats when they're all in RTCMediaStreamTrackStats, so no conflict after all (I seem unable to rename it).

PR #186, opened 2017-03-22
Change log, and make tidy

alvestrand:
Note that "make tidy" inserts an extra space for reasons unknown.
It is too tedious to remove it by hand, so it's left in.


alvestrand:
Merge after #182 and publish updated doc.

alvestrand:
TODO: Add 182 and 188, merge and publish.

PR #187, opened 2017-03-22
RTCMediaStreamTrackStats members to keep track of audio and video sync.

### LINKED ISSUE #158: Stats to keep track of sync between audio and video
It would be useful to be able to detect the audio and video of a stream becoming out-of-sync. There should be a current playout timestamp for remote audio and video tracks. They should be monotonically increasing and comparable between tracks from the same remote source.

Shall we add `DOMHighResTimeStamp playoutTimestamp` to `RTCMediaStreamTrackStats`?

henbos:
Add estimatedPlayoutTimestamp and estimatedCaptureTimestamp to RTCMediaStreamTrackStats.

Fixes #158


webrtc-stats.html
```diff
@@ -1189,6 +1189,8 @@
              boolean             ended;
              boolean             detached;
              DOMString           kind;
+             DOMHighResTimeStamp estimatedPlayoutTimestamp;
+             DOMHighResTimeStamp estimatedCaptureTimestamp;
```
alvestrand: I don't think we should have "estimated" in the name as long as it's not possible to get a better value.

webrtc-stats.html
```diff
@@ -1256,6 +1258,39 @@
...
+              <dt>
+                <dfn><code>estimatedCaptureTimestamp</code></dfn> of type <span class=
+                "idlMemberType"><a>DOMHighResTimeStamp</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Only valid for remote sources. This is the estimated capture time of this track.
+                  The capture time is the NTP timestamp of the last played out audio sample or video
+                  frame that has a known timestamp + half of what the estimated round trip time was
+                  when that audio sample or video frame arrived, extrapolated with the time elapsed
```
alvestrand: Which RTT estimate do we use?

webrtc-stats.html
```diff
@@ -1256,6 +1258,39 @@
...
               <dt>
+                <dfn><code>estimatedPlayoutTimestamp</code></dfn> of type <span class=
+                "idlMemberType"><a>DOMHighResTimeStamp</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Only valid for remote sources. This is the estimated playout time of this track.
+                  The playout time is the NTP timestamp of the last played out audio sample or video
+                  frame that has a known timestamp, extrapolated with the time elapsed since it was
+                  played out. This is the "current time" of the track and can be present even if
```
alvestrand: "current time" of the track relative to the NTP clock of the sender?

alvestrand:
@jesup @holmer @icydragons can you review this one?

webrtc-stats.html
henbos: Relative to the NTP clock of the sender sounds to me like if sender NTP time is 100 and rtt causes playout time to be at 90 this would be -10. I changed it to "This is the "current time" of the track in NTP clock time of the sender" to refer to the clock in terms of time units. I hope that wording makes sense.

webrtc-stats.html
henbos: Added "The estimated round trip time is computed from both STUN connectivity checks [[!STUN-PATH-CHAR]], including those that are sent for consent verification [[!RFC7675]]."

Open question: Do we use the latest value, a smoothed value, or up to user agent?

icydragons: I would say we should pick either latest or smoothed and use that. I am not sure which is better, but consistency on different implementations seems better then user agent choice.

webrtc-stats.html
```diff
@@ -1256,6 +1258,39 @@
...
+                  there is no audio or video currently playing.
```
icydragons: Do we want to say that is is expected to be able use estimatedPlayoutTimestamp from two tracks of the same source to calculate out of sync issues?

webrtc-stats.html
henbos: OK, latest it is.

vr000m: Do we really need an RTT estimate? 

Because the receiver gets a sync event (RTCP SR containing the RTP timestamp and the corresponding UNIX timestamp. We could essentially use the time elapsed in the local clock to estimate the captureTimestamp.

webrtc-stats.html
```diff
@@ -1269,6 +1271,41 @@
...
+                <dfn><code>playoutTimestamp</code></dfn> of type <span class="idlMemberType"><a>
+                DOMHighResTimeStamp</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Only valid for remote sources. This is the estimated playout time of this track.
+                  The playout time is the NTP timestamp of the last played out audio sample or video
+                  frame that has a known timestamp, extrapolated with the time elapsed since it was
+                  played out. This is the "current time" of the track in NTP clock time of the
+                  sender and can be present even if there is no audio or video currently playing.
```
vr000m: Would the AVSync still be estimated if the clockrates for audio and video are different?

Perhaps some explanation for calculating AV Sync would be helpful.

webrtc-stats.html
henbos: I updated it as current time + half RTT where half RTT is estimated as latest RTCP SR packet's NTP timestamp minus current time for when that packet arrived.

webrtc-stats.html
```diff
@@ -1202,6 +1202,8 @@
              boolean             ended;
              boolean             detached;
              DOMString           kind;
+             DOMHighResTimeStamp playoutTimestamp;
+             DOMHighResTimeStamp captureTimestamp;
```
icydragons: I still prefer these to be estimatedPlayoutTimestamp and estimatedCaptureTimestamp. I know we won't get a better estimate, but with the names as it it sounds like it is the last frame rather then what we estimate is the current one.

webrtc-stats.html
```diff
@@ -1269,6 +1271,44 @@
...
+                DOMHighResTimeStamp</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Only valid for remote sources. This is the estimated capture time of this track,
+                  i.e. what the "current time" is for the remote capturer. It is estimated by
+                  looking at the latest RTCP SR packet containing an NTP timestamp. The difference
+                  between the local clock when the packet arrived and the NTP timestamp gives an
+                  estimate of half the round trip time. The <code>captureTimestamp</code> is the
+                  current local time plus the latest half-RTT estimate and can be present even if
```
icydragons: Current "local time"? Does that mean in NTP time?

webrtc-stats.html
```diff
@@ -1269,6 +1271,44 @@
...
+                  The playout time is the NTP timestamp of the last played out audio sample or video
+                  frame that has a known timestamp (from an RTCP SR packet mapping RTP timestamps to
+                  NTP timestamps), extrapolated with the time elapsed since it was played out. This
+                  is the "current time" of the track in NTP clock time of the sender and can be
+                  present even if there is no audio or video currently playing.
+                </p>
+                  This can be useful for estimating how much audio and video is out of sync for two
+                  tracks from the same source, <code>audioTrackStats.playoutTimestamp -
+                  videoTrackStats.playoutTimestamp</code>.
+                <p>
```
vr000m: this section is outside the <p></p>

webrtc-stats.html
vr000m: I had this question before, there are no synchronized clocks, so the one-way delay cannot be calculated by taking a difference between NTP timestamps in the SR and the local time.

webrtc-stats.html
henbos: Done.

webrtc-stats.html
henbos: Done for estimatedPlayoutTimestamp. captureTimestamp removed.

PR #188, opened 2017-03-22
rtt undefined when no RTCP RR

### LINKED ISSUE #180: RoundTripTime not defined when the underlying stream can't calculate it
RoundTripTime can only be calculated after an exchange of RTCP packets with the relevant information.  Until then, it's not possible to calculate.  What should be RoundTripTime have before we know the time?  0?  undefined?  null?  And what will be the impact on existing applications if we choose something like undefined or null?

vr000m:
fixes #180,


vr000m:
@alvestrand this is a quick fix, if can be merged before the new release #186.

jesup:
I think that needs to be " If no RTCP Receiver Report is received with a DLSR value other than 0, the round trip time is left undefined."

vr000m:
@jesup correct, to cover the first RTCP RR when no SR is received.

Issue #189, opened 2017-03-22
Reuse of "inbound-rtp" and "outbound-rtp" for RTCP is confusing.


jan-ivar:
The `isRemote` overload of `"inbound-rtp"` and `"outbound-rtp"` is just confusing. It invites surprise and bugs to anyone trying to enumerate stats (scripts end up reading wrong data without knowing it!)

I can't think of a single use-case where enumerating by type alone is not wrong.

It's also false symmetry: both dictionaries have members which existence depend on `isRemote`:
 * [RTCOutboundRTPStreamStats](http://rawgit.com/w3c/webrtc-stats/master/webrtc-stats.html#outboundrtpstats-dict*) has `framesEncoded` (local-only), `remoteTimestamp` (remote-only),
 * [RTCInboundRTPStreamStats](http://rawgit.com/w3c/webrtc-stats/master/webrtc-stats.html#inboundrtpstats-dict*) has `framesDecoded` (local-only), `roundTripTime` (remote-only).

These absences break down easily by type, i.e. *never* appear in subsequent calls where missing.

## Proposal

(***Edited*** *to include feedback in comments below. Thanks Harald!*)

Remove `isRemote`, and round out the two rtp types with two more for rtcp instead:
 * "outbound-rtp"
 * "inbound-rtp"
 *  **"remote-inbound-rtp"** (formerly "inbound-rtp" with `isRemote == true`)
 *  **"remote-outbound-rtp"** (formerly "outbound-rtp" with `isRemote == true`)

Define discrete dictionaries for the four types (in inheritance rather than corrresponding order):
```js
dictionary RTCSentRTPStreamStats : RTCRTPStreamStats {
    unsigned long       packetsSent;
    unsigned long long  bytesSent;
};

dictionary RTCOutboundRTPStreamStats : RTCSentRTPStreamStats {
    double              targetBitrate;
    unsigned long       framesEncoded;
};

dictionary RTCRemoteOutboundRTPStreamStats : RTCSentRTPStreamStats {
    DOMHighResTimeStamp remoteTimestamp;
};
```

```js
dictionary RTCReceivedRTPStreamStats : RTCRTPStreamStats {
    unsigned long      packetsReceived;
    unsigned long long bytesReceived;
    unsigned long      packetsLost;
    double             jitter;
    double             fractionLost;
    unsigned long      packetsDiscarded;
    unsigned long      packetsRepaired;
    unsigned long      burstPacketsLost;
[...]

PR #191, opened 2017-04-03
Refactor out isRemote.

jan-ivar:
Fix for https://github.com/w3c/webrtc-stats/issues/189.

I found it affirming that this PR revealed to me that the stats examples in both specs have always been broken (they never tested for `!isRemote`)!

@alvestrand PTAL - I'll try to add a slide for tomorrow.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -73,8 +73,10 @@
         <li>It MUST support generating statistics for the type RTCOutboundRTPStreamStats, with
         attributes packetsSent, bytesSent.
         </li>
-        <li>For all subclasses of RTCRTPStreamStats, it MUST include ssrc, mediaType,
-        and either remoteId or localId as appropriate.
+        <li>For all subclasses of RTCRTPStreamStats, it MUST include ssrc and mediaType.
+        When stats exist for both sides of a connection, in the form of an
+        inbound-rtp / remote-outbound-rtp pair or an outbound-rtp /
+        remote-inbound-rtp pair, the members remoteId and localId MUST also be present.
         </li>
         <li>It MAY support generating other stats.
         </li>
@@ -408,7 +410,7 @@ enum RTCStatsType {
               corresponding to an outbound stream that is currently sent with
               this <code>RTCPeerConnection</code> object. It is measured at the
               remote endpoint and reported in an RTCP RR/XR. It is accessed by
-              the <code><a>RTCInboundRTPStreamStats</a></code>.
+              the <code><a>RTCRemoteInboundRTPStreamStats</a></code>.
             </p>
           </dd>
           <dt>
@@ -420,7 +422,7 @@ enum RTCStatsType {
               corresponding to an inbound stream that is currently received with
               this <code>RTCPeerConnection</code> object. It is measured at the
               remote endpoint and reported in an RTCP SR. It is accessed by the
-              <code><a>RTCOutboundRTPStreamStats</a></code>.
+              <code><a>RTCRemoteOutboundRTPStreamStats</a></code>.
             </p>
           </dd>
           <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -73,8 +73,8 @@
         <li>It MUST support generating statistics for the type RTCOutboundRTPStreamStats, with
         attributes packetsSent, bytesSent.
         </li>
-        <li>For all subclasses of RTCRTPStreamStats, it MUST include ssrc, associateStatsId,
-        mediaType.
+        <li>For all subclasses of RTCRTPStreamStats, it MUST include ssrc, remoteId,
+        localId, mediaType.
         </li>
         <li>It MAY support generating other stats.
         </li>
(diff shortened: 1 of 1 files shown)


jan-ivar:
Note that this PR also adds back `remoteId` instead of `associateStatsId`, because this is now possible. It also adds a symmetrical `localId`. I think this will help avoid bugs like the one in the stats example.

I found it also helps with readability in the spec, since we can now read the exact corresponding type that `remoteId`/`localId` point to, rather than the nondescript `RTCStats`. [Preview here](https://rawgit.com/jan-ivar/webrtc-stats/197e58083aa0d113ec84866f591ceb9933b2c0e0/webrtc-stats.html#inboundrtpstats-dict*).

jan-ivar:
In fact, the stats association example now seems redundant. Removed.

jan-ivar:
PR conflict with https://github.com/w3c/webrtc-stats/pull/192 but can be rebased later.

webrtc-stats.html
```diff
@@ -73,8 +73,8 @@
         <li>It MUST support generating statistics for the type RTCOutboundRTPStreamStats, with
         attributes packetsSent, bytesSent.
         </li>
-        <li>For all subclasses of RTCRtpStreamStats, it MUST include ssrc, associateStatsId,
-        isRemote, mediaType.
+        <li>For all subclasses of RTCRTPStreamStats, it MUST include ssrc, remoteId,
+        localId, mediaType.
```
alvestrand: This should be "either remoteId or localId, as  appropriate".

alvestrand:
I now think I like this one. @vr000m needs to review it.
It's a big enough change that we should push out a new dated draft once it's merged.

jan-ivar:
Ok I need to rebase this one now...

jan-ivar:
New [preview link](https://rawgit.com/jan-ivar/webrtc-stats/e4b498d324ce0bd4882d0ba449de5003c296a570/webrtc-stats.html#rtcstatstype-str*).

webrtc-stats.html
```diff
@@ -398,6 +400,30 @@
...
           <dt>
+            <dfn><code>remote-inbound-rtp</code></dfn>
+          </dt>
+          <dd>
+            <p>
+              Statistics for the remote endpoint's inbound RTP stream
+              corresponding to an outbound stream that is currently sent with
+              this <code>RTCPeerConnection</code> object. It is measured at the
+              remote endpoint and reported in an RTCP RR/XR. It is accessed by
+              the <code><a>RTCInboundRTPStreamStats</a></code>.
```
alvestrand: Oops - missing Remote in typename here.

webrtc-stats.html
```diff
@@ -398,6 +400,30 @@
...
+          <dt>
+            <dfn><code>remote-outbound-rtp</code></dfn>
+          </dt>
+          <dd>
+            <p>
+              Statistics for the remote endpoint's outbound RTP stream
+              corresponding to an inbound stream that is currently received with
+              this <code>RTCPeerConnection</code> object. It is measured at the
+              remote endpoint and reported in an RTCP SR. It is accessed by the
+              <code><a>RTCOutboundRTPStreamStats</a></code>.
```
alvestrand: Also missing "Remote".

webrtc-stats.html
```diff
@@ -73,8 +73,8 @@
         <li>It MUST support generating statistics for the type RTCOutboundRTPStreamStats, with
         attributes packetsSent, bytesSent.
         </li>
-        <li>For all subclasses of RTCRtpStreamStats, it MUST include ssrc, associateStatsId,
-        isRemote, mediaType.
+        <li>For all subclasses of RTCRTPStreamStats, it MUST include ssrc, mediaType,
+        and either remoteId or localId as appropriate.
```
alvestrand: Thinking about this a little more .... if the implementation doesn't support remote stats, or the first RR / SR hasn't arrived yet, it's not logical to have the remote stats, and then remoteId doesn't make sense.

Suggested rephrase: "MUST include ssrc and mediaType. When stats exist for both sides of a connection, in the form of an inbound-rtp / remote-outbound-rtp pair or an outbound-rtp / remote-inbound-rtp pair, the members "remoteId" and "localid" MUST be present".

webrtc-stats.html
jan-ivar: @alvestrand good catch, thanks!

jan-ivar:
@vr000m Anything else needed here?

vr000m:
@aboba are you okay with the changes.

webrtc-stats.html
```diff
@@ -1064,6 +1144,57 @@
...
+        </h3>
+        <p>
+          The <dfn>RTCRemoteOutboundRTPStreamStats</dfn> dictionary represents
+          the remote endpoint's measurement metrics for its outgoing RTP stream
+          (our incoming RTP stream).
+        </p>
+        <div>
+          <pre class="idl">dictionary RTCRemoteOutboundRTPStreamStats : RTCSentRTPStreamStats {
+             DOMString           localId;
+             DOMHighResTimeStamp remoteTimestamp;
```
henbos: This is a separate issue, but wouldn't it make sense to have a remoteTimestamp on RTCRemoteInboundRTPStreamStats as well?

jan-ivar: @henbos See https://github.com/w3c/webrtc-stats/pull/164#issuecomment-284976013.

PR #195, opened 2017-04-06
Adding "codec type" and transportId to RTCCodecStats.

### LINKED ISSUE #109: RTCCodecStats needs `transportId` and `isRemote` to give it context
An `RTCCodecStats` by itself (if not being actively used to send/receive a stream) doesn't have enough context to be useful. I'd suggest adding:

- `transportId` - a payload type only makes sense in the context of an RTP session, and there's one RTP session per media transport so I think this works.
- `isRemote` - is this a codec we're prepared to encode, or decode?

taylor-b:
Fixes #109.

The codec type indicates if it's a media format the implementation is
prepared to encode or decode, and the transportId identifies what RTP
session it's being used for.

An implementation of getStats may want to return RTCCodecStats objects
for codecs not being actively used, in which case these fields are
necessary in order for those objects to provide meaningful information.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -664,6 +664,8 @@ enum RTCStatsType {
         <div>
           <pre class="idl">dictionary RTCCodecStats : RTCStats {
              unsigned long payloadType;
+             RTCCodecType  codecType;
+             DOMString     transportId;
              DOMString     mimeType;
              unsigned long clockRate;
              unsigned long channels;
@@ -682,7 +684,29 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Payload type as used in RTP encoding.
+                  Payload type as used in RTP encoding or decoding.
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>codecType</code></dfn> of type <span class=
+                "idlMemberType"><a>RTCCodecType</a></span>
+              </dt>
+              <dd>
+                <p>
+                  <code>"encode"</code> or <code>"decode"</code>, depending on
+                  whether this object represents a media format that the
+                  implementation is prepared to encode or decode.
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>transportId</code></dfn> of type <span class=
+                "idlMemberType"><a>DOMString</a></span>
+              </dt>
+              <dd>
+                <p>
+                  The unique identifier of the transport on which this codec is
+                  being used, which can be used to look up the corresponding
+                  <code><a>RTCTransportStats</a></code> object.
                 </p>
               </dd>
               <dt>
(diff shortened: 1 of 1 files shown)


taylor-b:
I ended up with just going with "encode"/"decode", without a third "encode-decode" option. The trouble is, `RTCCodecStats` has `payloadType` and `implementation` fields. So we'd end up with the rule: "If the encode and decode directions use the same codec implementation, and the same payload type, you get one object, otherwise you get two." And this seemed more confusing to deal with, for both the implementers and `getStats` API users, than just "you always get two objects."

vr000m:
I am wondering if the standard case is that we use the same payload type and codec for encode and decode, then we will always get two objects even though one would suffice.

I am thinking inbound and oubtbound point to codecIds, if the same codec is used for encode and decode, maybe both pointing to the same RTCCodecStats would make sense. I am on the fence on this... because in our use-case I will definitely filter and combine the codec stats if they are exactly the same.

cc: @karthikbr82

henbos:
Actually in Chrome we typically use the same payload type and implementation string in both directions. Having to have more than one stats object would be a special case and I suspect(?) this would be the case for other UAs. An "encode-decode" type should cover the vast majority of cases and relieve us from otherwise having duplicate codec stats.

vr000m:
@taylor-b WDYT about the above comments?

taylor-b:
Even if "encode-decode" handles the majority of cases, it's the minority that cause complexity. What happens if you go from "decode" to "encode-decode" or vice versa? Is there a new codec object (new stats ID) or not? What happens if a payload type changes on one side for an "encode-decode" codec (I've seen this happen); does it split into two new objects (new IDs)?

Also, it just feels like a cleaner object model to imagine encoders and decoders as separate entities. The codec bi-directionality is just a consequence of SDP, and I like avoiding SDP-isms when possible; it makes `getStats` more compatible with ORTC.

However, if it's 2 votes against 1, I can change it, as long as we can come up with answers to the above questions.

henbos:
For simplicity I guess we should decide how to uniquely identify a codec for all cases instead of worrying about the number of stats objects.

Another complexity is what if the |implementation| is different. You could have one implementation for encode and another for decode, and a third that works for both (having one or two objects as necessary assuming payloadType is the same).

I want to avoid the "what if it changes, do we split?" complexity. So far all codecId references are from RTCRTPStreamStats, so the encode/decode is implied in the reference, all you would need to know to correctly identify the codec in this case is (payloadType, implementation). The codec.codecType could be kept up-to-date about if its currently only for encode, only decode or (perhaps) both, but when you're referencing a codec for encoding you don't care if it's encode-only or both and don't need that to be part of its id or care if that changes.

If an encode-decode splits it would split up into two different payloadTypes would it not, so that would require two codec objects anyway?

(payloadType, implementation) are necessary identifiers. Is codecType also necessary for identifying a codec? It sounds simpler, but I don't know if it matters in practice. I don't have a strong opinion on this. Am I missing anything?

taylor-b:
I think the question we need to answer is: "what object does an `RTCCodecStats` represent"? This is missing from the description of `RTCCodecStats` right now.

We could say it represents an SDP media format, or it represents an instance of an encoder/decoder. The fact that it has an `implementation` field made me think it was the latter.

Or maybe it *was* intended to be just an SDP media format, and `implementation` is the abnormality. But I can imagine adding things like "encode/decode time in milliseconds" or "encode/decode CPU utilization" (which are stats Chrome collects, but doesn't expose). In which case the "instance of encoder/decoder" interpretation would make more sense.

henbos:
Considering implementation, I like to think of it as an instance of an encoder/decoder, and it would make sense to put encoder/decoder stats on it. Encoder and decoder being separate objects, I vote to have the stats objects the same - "encode" and "decode", no "encode-decode". What's your take on this @vr000m ?

vr000m:
After re-reading the comments, I feel that separate CodecStats for object encode and decode would be appropriate if we foresee more encode/decode-dependent stats.

PR #196, opened 2017-04-11
RTCIceCandidatePairStats: Remove writable/readable, add consentExpiredTimestamp

### LINKED ISSUE #163: RTCIceCandidatePairStats.writable/readable: redundant?
`writable` is "Has gotten ACK to an ICE request."
Isn't this the same as `responsesReceived > 0`? `responsesReceived` is "Represents the total number of connectivity check responses received."

`readable` is "Has gotten a valid incoming ICE request."
Isn't this the same as `requestsReceived > 0`? `requestsReceived` is "Represents the total number of connectivity check requests received (including retransmissions)."

If so I propose we remove `writable` and `readable`.

henbos:
Fixes #163.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1771,7 +1771,6 @@ enum RTCStatsType {
              unsigned long long            priority;
              boolean                       nominated;
              boolean                       writable;
-             boolean                       readable;
              unsigned long long            bytesSent;
              unsigned long long            bytesReceived;
              DOMHighResTimeStamp           lastPacketSentTimestamp;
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1856,21 +1855,13 @@
...
-                <dfn><code>readable</code></dfn> of type <span class=
-                "idlMemberType"><a>boolean</a></span>
+                <dfn><code>writable</code></dfn> of type <span class="idlMemberType"><a>boolean</a>
+                </span>
               </dt>
               <dd>
                 <p>
-                  Has gotten a valid incoming ICE request.
+                  Has received a valid STUN binding response and the consent has not expired as
+                  defined in [[!RFC7675]] section 5.1.
```
vr000m: Probably state it is True when a valid ... is received and False when consent has expired. Probably need to know when the consent expired?

henbos: I replaced writable with consentExpiredTimestamp. That together with responsesReceived should be enough?

henbos:
PTAL @vr000m, I replaced writable with consentExpiredTimestamp.

webrtc-stats.html
```diff
@@ -2106,6 +2087,19 @@
...
               </dd>
+              <dt>
+                <dfn><code>consentExpiredTimestamp</code></dfn> of type <span class=
+                "idlMemberType"><a>DOMHighResTimeStamp</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the timestamp at which the latest valid STUN binding response expired,
+                  as defined in [[!RFC7675]] section 5.1. The time is relative to the UNIX epoch
+                  (Jan 1, 1970, UTC). If a valid STUN binding response has not been made
```
vr000m: maybe the reference to UNIX epoch can be made in the section where we define timestamps in the introduction

PR #197, opened 2017-04-11
RTCTransportStats: packetsSent/Received added.

### LINKED ISSUE #173: RTCTransportStats packet counters
RTCTransportStats have bytesSent and bytesReceived but not packetsSent and packetsReceived. Should we add those counters?

See also #172

henbos:
Fixes #173.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1535,7 +1535,9 @@ enum RTCStatsType {
         </p>
         <div>
           <pre class="idl">dictionary RTCTransportStats : RTCStats {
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1570,6 +1581,15 @@
                 </p>
               </dd>
               <dt>
+                <dfn><code>packetsReceived</code></dfn> of type <span class="idlMemberType"><a>
```
henbos: Will place this above bytesReceived before merging.

PR #206, opened 2017-05-10
packetsRetransmitted and bytesRetransmitted to RTCSentRTPStreamStats

### LINKED ISSUE #154: Stat for retransmitted bytes
The non-standardized Chromium getStats contains the following stat:
```
// The total retransmission bandwidth, in bits per second.
VideoBwe.googRetransmitBitrate
```
Is this useful? Should we add retransmitted bytes counters to the spec? We have bytesSent/bytesReceived on various stats dictionaries already: RTCTransportStats, RTCDataChannelStats, RTC[In/Out]boundRTPStreamStats and RTCIceCandidatePairStats. Does the existing bytesSent/bytesReceived include or exclude bytes that are retransmitted?

karthikbr82:
fixes #154

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1039,6 +1039,8 @@ enum RTCStatsType {
           <pre class="idl">dictionary RTCSentRTPStreamStats : RTCRTPStreamStats {
              unsigned long      packetsSent;
              unsigned long long bytesSent;
+             unsigned long      packetsRetransmitted;
+             unsigned long long bytesRetransmitted;
 };</pre>
           <section>
             <h2>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1066,6 +1068,26 @@
...
                 </p>
               </dd>
+              <dt>
+                <dfn><code>packetsRetransmitted</code></dfn> of type <span class="idlMemberType"><a>unsigned
+                long</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Total number of RTP packets retransmitted by the sender on receiving a retransmission request (e.g., NACK) from the receiver for this SSRC. The retransmitted packet may be sent as defined in
+                  [[!RFC4588]].
```
taylor-b: But RTX streams are their own stream with their own SSRC, so wouldn't they have a separate RTCSentRTPStreamStats?

vr000m: That is correct, I wonder what was the precipitated the original issue.

vr000m: @taylor-b any further comments? or should I just close the PR and the corresponding issue?

taylor-b: @vr000m I think you can just close them.

Issue #141, opened 2017-01-25
Unclear which framerate "framePerSecond" represents


taylor-b:
Current definition is: "It represents the nominal FPS value."

But we're unclear on which FPS value this is referring to. Is it the video source's, the MediaStreamTrack's, or the RtpSender's? Example:

1. Camera produces 60 frames per second.
2. MediaStreamTrack constraints bring it down to 30.
3. RtpSender parameters (degradation preferences or maxFramerate) bring it down further, to 15.

What should go in framesPerSecond? 60, 30 or 15? @vr000m, do you remember what the original intention was?

alvestrand:
Note - it is presently attached to the RTCRTPMediaStreamStats, which exists once per simulcast stream. I remember thinking that when the "main" stream is configured for 60 fps, and the "secondary" is configured for 15 fps (4x degradation), the values would be 60 and 15 respectively, even if the "main" stream is not sending this fast (for whatever reason it might find to slow down - perhaps there aren't that many incoming frames).

gunnarhm:
It can also be
4. Really achieved frame rate. 

And there should be two figures, one for sending and one for receiving. 

The name "Nominal ..." indicates however that it is not the really achieved framerate. 
I suggest that 4 values are provided with different names:
Nominal transmit frame rate (after all negotiations, that is figure 3 for transmission )
Nominal receive frame rate (after all negotiations, that is figure 3 for reception)
Really achieved mean transmitted frame rate.
Really achieved mean received frame rate.

alvestrand:
I must correct myself: FrameRate (nominal) attaches to  RTCMediaStreamTrackStats. There are no frame-related stats connected to RTPStreams.

MediaStreamTracks are unidirectional, so there's only one frame rate. Outgoing and incoming frame rates (if we have both) are on different tracks.
Stat values should normally not be averages - see the discussion at http://rawgit.com/w3c/webrtc-stats/master/webrtc-stats.html#guidelines-for-design-of-stats-objects (I don't think we've published a dated editor's draft since this was merged). We already have frame counts at multiple levels.

taylor-b:
> Nominal receive frame rate (after all negotiations, that is figure 3 for reception)

Since this isn't signaled explicitly in SDP, I think this would only make sense for codecs that put a framerate in the actual bitstream, such as H.264 (https://cardinalpeak.com/blog/the-h-264-sequence-parameter-set/)

> Really achieved mean transmitted/received frame rate.

[...]

PR #209, opened 2017-05-10
definition for 'framesPerSecond'

karthikbr82:
Fixes - #141

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1428,7 +1428,9 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Only valid for video. It represents the nominal FPS value.
+                  Only valid for video. It represents the nominal FPS value before the degradation
+                  preference is applied. For sending tracks it is the current captured FPS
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1428,7 +1428,9 @@
               </dt>
               <dd>
                 <p>
-                  Only valid for video. It represents the nominal FPS value.
+                  Only valid for video. It represents the nominal FPS value before the degradation
+                  preference is applied. For sending tracks it is the current captured FPS
+                  and for the receiving tracks it is the current decoding framerate.
```
henbos: How is the FPS calculated? Over the last second, based on the time between the last two frames, etc?

vr000m: Elsewhere on the internet (Apple, states: https://developer.apple.com/reference/avfoundation/avassettrack/1386182-nominalframerate?language=objc): 
> The nominal frame rate indicates the number of frames per second for tracks that carry a full frame per media sample

Issue #61, opened 2016-09-21
Implementing the Circuit Breaker based on StatsAPI


vr000m:
Information for calculating the Circuit breaker with the information in the Stats.

CB Trigger 1. latest timestamp of a packet for each SSRC. Covers both sent and received streams
CB Trigger 2. latest timestamp of the RTCP packet for each SSRC. I believe this is covered with the timestamp in the remote stats.
CB Trigger 3. TFRC can be computed with the information, although one piece of information that is missing is the average RTCP interval.

alvestrand:
We could also expose a count of circuit-breaker actions triggered (and leave it as "undefined" for browsers that don't implement circuit-breakers), but the information listed above is interesting anyway.

vr000m:
In addition:
+ target RTCP interval
+ average RTCP Interval

PR #211, opened 2017-05-10
timestamps needed for circuit breakers and a/v sync.

vr000m:
Fixes information needed for #61 and #158.


vr000m:
@taylor-b WDYT about reporting the average RTCP interval as part of the sent RTP stats.

webrtc-stats.html
```diff
@@ -768,6 +768,7 @@
              double             burstDiscardRate;
              double             gapLossRate;
              double             gapDiscardRate;
+             DOMHighResTimeStamp  lastPacketReceivedTimestamp;
```
jan-ivar: When will `lastPacketReceivedTimestamp` differ from `timestamp`?

vr000m: In the local reports, the timestamp indicates when the report was generated.
In the remote reports, the timestamp indicates when the report was received.

I always thought that the timestamp in the getStats() for local objects was timestamp when the snapshot of the getStats was created, which may have been when the getStats was called.

webrtc-stats.html
```diff
@@ -936,6 +937,15 @@
                   [[!RFC7004]], however, the actual value is reported without multiplying by 32768.
                 </p>
               </dd>
+              <dt>
+                <dfn><code>lastPacketReceivedTimestamp</code></dfn> of type <span class=
+                "idlMemberType"><a>DOMHighResTimeStamp</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the timestamp at which the last packet was received for this media stream track.
```
taylor-b: Do we say "media stream track" anywhere else?

webrtc-stats.html
```diff
@@ -1066,6 +1078,27 @@
...
+                <dfn><code>lastPacketSentTimestamp</code></dfn> of type <span class=
+                "idlMemberType"><a>DOMHighResTimeStamp</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the timestamp at which the last packet was sent for this media stream track.
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>averageRTCPInterval</code></dfn> of type <span class=
```
taylor-b: We generally don't have averages; could we have "sent compound RTCP packet count" and "first/last compound RTCP packet timestamp" stats, like the approach taken in #213?

webrtc-stats.html
vr000m: I believe "SSRC" is used more often, will change.

webrtc-stats.html
vr000m: We currently do not have timestamps nor counters on when the RTCP reports were sent or received. Although a cleaner approach, I am not convinced if those counters brings value.

IMO, the variation in the RTCP interval is not interesting, so the average as is works for me. Having it available as a single metric vs calculating by three metrics, I am on the fence on this, If there would be more use-cases for those additional metrics, I'd be more than happy to change my position on this.

webrtc-stats.html
jan-ivar: > I always thought that the timestamp in the getStats() for local objects was timestamp when the snapshot of the getStats was created, which may have been when the getStats was called.

Thanks, I got confused. That sounds right (ironically, [remoteTimestamp](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-remotetimestamp) contains a clearer definition of the local timestamp than timestamp does: *"This differs from [timestamp](https://w3c.github.io/webrtc-pc/#dom-rtcstats-timestamp), which represents the time at which the statistics were generated or received by the local endpoint."*)

So in remote reports, `lastPacketReceivedTimestamp` is in the clock of the remote?

webrtc-stats.html
jan-ivar: Isn't when the RTCP report were received, [timestamp](https://w3c.github.io/webrtc-pc/#dom-rtcstats-timestamp)? The [guideline](https://w3c.github.io/webrtc-stats/#guidelines-for-design-of-stats-objects) says "no averages".

taylor-b: Maybe the additional counters wouldn't provide much additional value, but it's just as easy to implement, so the only disadvantage is a larger API surface. I'd say it's worth doing just to be consistent.

webrtc-stats.html
vr000m: It is the local timestamp of receiving a packet from the remote participant/endpoint.

webrtc-stats.html
vr000m: average RTCP is calculated by the endpoint for scheduling RTCP report. So this is about exposing an existing value from the RTP stack. RFC3550/RFC4585 has:
+ target RTCP interval
+ average RTCP interval

webrtc-stats.html
jan-ivar: @vr000m Does this merit a disclaimer similar to the one in [remoteTimestamp](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-remotetimestamp) for the case where this represents remote time?

vr000m: @jan-ivar it is the local timestamp of the latest sent packet and latest received packet. not remote timestamps.

jan-ivar: You're changing `RTCReceivedRTPStreamStats` here, which is the base of both `RTCInboundRTPStreamStats` and `RTCRemoteInboundRTPStreamStats`. In the latter case, the timestamps will not be based on the local clock of the reader, which might warrant a warning, so people don't try to compare it to other timestamps in the report.

vr000m: I see, @jan-ivar I could move the timestamp to the appropriate inbbound/outbound local objects, that might seem cleaner. WDYT?

webrtc-stats.html
vr000m: @taylor-b: WDYT, expose an existing value or do you want to calculate the average for the RTCP interval from:
+ rtcpPacketsSent
+ totalRtcpInterval

henbos: @stefanholmer Do we have target and average RTCP interval in our stack already (RFC3550/RFC4585)?

PR #213, opened 2017-05-10
RTCIceCandidatePairStats.last[Request/Response]Timestamp added

henbos:
Fixes #201 and #63.

PR #214, opened 2017-05-10
adding total consent interval

### LINKED ISSUE #63: Missing consent interval in RTCIceCandidatePairStats

vr000m:
fixes #63.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2070,6 +2070,7 @@ enum RTCStatsType {
              unsigned long long            retransmissionsSent;
              unsigned long long            consentRequestsSent;
              DOMHighResTimeStamp           consentExpiredTimestamp;
+             unsigned long long            sumCheckInterval;
 };</pre>
           <section>
             <h2>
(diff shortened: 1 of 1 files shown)


henbos:
If lastResponseTimestamp of #213 is not enough of a heartbeat metric.

taylor-b:
If we're adding `lastRequestTimestamp`, it would be more consistent to add `firstRequestTimestamp` rather than `sumCheckInterval`. Because `lastRequestTimestamp - firstRequestTimestamp` would be equivalent to `sumCheckInterval`.

vr000m:
@taylor-b I am fine with your change, is the firstRequestTimestamp, the timestamp of the first connectivity check?

@henbos could you do taylor's change as part of your PR #213, so that I can just close this one?

henbos:
SGTM

Issue #151, opened 2017-02-01
Stat for audio playout delay


henbos:
The non-standardized Chromium getStats contains the following stats:
```
// The total audio playout delay, including algorithmic, jitter buffer,
// and soundcard delay, in milliseconds.
ssrc.googCurrentDelayMs
```
Is this useful? Should we standardize it or something similar? Double seconds is the preferred unit.
Should we add `double audioDelay` to RTCMediaStreamTrackStats?

alvestrand:
This is related to the need that was the basis for https://github.com/alvestrand/webrtc-framelog
The critical part of a definition here are the measurement points: At what point do we start the clock running, and at what point do we say "the delay is completed"?

Side point: If we can't get exact values (such as delays in the audio subsystem after the samples have been handed over), should we add an estimate into the delay, or use the last time at which we were able to measure time?

vr000m:
I agree with @alvestrand that the delay measurements need exact details on when the clock starts and when the clock stops. There are several components in this calculation: the more difficult one is soundcard delay, although the algorithmic and jitter buffer can be measured very accurately. 

I would start with something that defines the sum of the playout algorithm+jitter buffer. Both the current (which could be the last played sample or the delay if there were a current sample to play) and a long-term total (based on each frame).

henbos:
This may or may not be related to #158 which discusses timestamps for the playout of audio and video.

henbos:
+@hlundin

Is this right?
+ **Jitter buffer delay**: The currently expected delay due to jitter buffer, which is the difference between the timestamp of the latest audio sample coming out of the jitter buffer and the timestamp it had when it was inserted into the jitter buffer.
+ **Sound card delay**: This has to do with the size of the buffer we feed to the sound card, since the entire buffer has to be filled up before fed to the sound card. The delay is (_number of samples that fit into the buffer_ - 1) * (_the time it takes to play one sample_).
+ **Algorithmic delay**: _???_

If we get these right we can define `RTCMediaStreamTrackStats.audioDelay` as the sum of these delays.

henbos:
Do we want?
- Jitter buffer delay
- Decode buffer delay
- Sound card buffer delay

vr000m:
Jitter buffer delay: is the delay between the first packet belonging to an audio/video frame entering the jitter buffer and the complete frame exiting the jitter buffer. 

[...]

PR #217, opened 2017-05-10
jitter buffer delay added to media stream tracks

vr000m:
Fixes the last open issue proposed in #151.


henbos:
PTAL @hlundin

henbos:
Wait, is jitter buffer before or after decoding, or does it encompass jitter and decoding?

PR #221, opened 2017-05-11
Adding audio level stat that can be used to compute averages.

### LINKED ISSUE #220: Need some way of getting an averaged audio level stat
`RTCMediaStreamTrackStats.audioLevel` only represents the level of the last encoded/decoded chunk of audio, or so I assume (may need to be clarified, see https://github.com/w3c/webrtc-stats/issues/193). So it's not useful for all use cases. If `getStats` is called at the wrong time, it may see an `audioLevel` of 0, just because it happened to be called during a blip of silence. The application may really be more interested in an average level over a certain time period.

The guideline for "averaged" stats is to provide "sum" and "count" fields, such that the application can compute an average over any time interval by dividing "sum/count" for that interval. This is a bit more complicated for audioLevel, since it's based on the root mean square (RMS) of audio samples. So we would have to do something like provide a sum of "mean square X sample duration" (say, "totalAudioEnergy"), and a "totalDuration", and let the application compute the average audio level as "sqrt(totalAudioEnergy)/totalDuration".

taylor-b:
Fixes #220.

It's slightly different than most of the stats of this sort, since the
application needs to compute "sqrt(a/b)", not just "a/b". But the
concept is the same.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1560,6 +1560,7 @@ enum RTCStatsType {
              double              echoReturnLossEnhancement;
              unsigned long long  totalSamplesSent;
              unsigned long long  totalSamplesReceived;
+             double              totalSamplesDuration;
              unsigned long long  concealedSamples;
              unsigned long long  concealmentEvents;
              double              jitterBufferDelay;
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1749,6 +1750,32 @@
...
+                <dfn><code>totalAudioEnergy</code></dfn> of type <span class=
+                "idlMemberType"><a>double</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Only valid for audio. This value MUST be computed as follows:
+                  for each audio sample sent/received for this object (and
+                  counted by <code><a>totalSamplesSent</a></code> or
+                  <code><a>totalSamplesReceived</a></code>), add the square of
+                  the sample's value divided by the highest-intensity encodable
```
taylor-b: Is this too ambiguous? It's meant to mean `square(sample/max)`, not `square(sample)/max`.

taylor-b:
@vr000m, @henbos, can you take a look?

taylor-b:
@vr000m @henbos Ping.

webrtc-stats.html
vr000m: Is the expression below and the explanation in the text here in sync?

vr000m: My bad, I think I get what you mean.

The first paragraph is about how totalAudioEnergy is computed. And the second paragraph is how it could be used by an application.

vr000m: LGTM

webrtc-stats.html
```diff
@@ -1775,6 +1777,43 @@
...
+                  For example, if a 10ms packet of audio is received with an
+                  RMS of 0.5 (out of 1.0), this should add <code>0.5 * 0.5 *
+                  0.01 = 0.0025</code> to <code>totalAudioEnergy</code>. If
+                  another 10ms packet with an RMS of 0.1 is received, this should
+                  similarly add <code>0.0001</code> to
+                  <code>totalAudioEnergy</code>. Then,
+                  <code>Math.sqrt(totalAudioEnergy/totalSamplesDuration)</code>
+                  becomes <code>Math.sqrt(0.0026/0.02) = 0.36</code>, which is
+                  the same value that would be obtained by doing an RMS
+                  calculation over the contiguous 20ms segment of audio.
```
henbos: nit: Can you change this example to include two getStats calls so that people don't get confused about where to do the sqrt?

webrtc-stats.html
```diff
@@ -1775,6 +1777,43 @@
...
+                  counted by <code><a>totalSamplesSent</a></code> or
+                  <code><a>totalSamplesReceived</a></code>), add the square of
+                  the sample's value divided by the highest-intensity encodable
+                  value, multiplied by the duration of the sample in seconds.
+                </p>
+                <p>
+                  This can be used to obtain a root mean square (RMS) value
+                  that uses the same units as <code><a>audioLevel</a></code>,
+                  using the formula
+                  <code>Math.sqrt(totalAudioEnergy/totalSamplesDuration)</code>.
```
henbos: nit: Can you also have the RFC reference here? (Same as the audioLevel reference)

webrtc-stats.html
henbos: I think it's clear but you could clarify with a formula at the end of the sentence, (sampleEnergy^2/maxEnergy)*sampleDuration.

vr000m:
@taylor-b if you can make the changes in the upcoming days, after that I will make a new editor's release.

taylor-b:
Updated and fixed merge conflicts.

PR #226, opened 2017-06-15
Fix markup issues

vivienlacourba:
Fix recently introduced markup issues.


vr000m:
@vivienlacourba, how do we fix this in the released version. Also noticed that the URL for the previous editor draft is incorrect.

vivienlacourba:
@vr000m I guess we can apply the same modifications directly to `archives/20170614/webrtc-stats.html`

vr000m:
@vivienlacourba could you do that? meanwhile I will merge this one

PR #237, opened 2017-08-24
Typo fix datachannelid to datachannelId

soareschen:
Was it accidental or intentional to name the field `datachannelid` instead of `datachannelId`?

If it was not intentional, this PR rename  the field to follow the same camelCase convention as other Id fields.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -220,7 +220,7 @@
           (camelCase).
         </p>
         <p>
-          Names ending in "Id" (such as "datachannelId") are always a <a>stats object
+          Names ending in "Id" (such as "transportId") are always a <a>stats object
           reference</a>; names ending in "Ids" (such as "trackIds") are always of type
           sequence&lt;DOMString&gt;, where each DOMString is a <a>stats object reference</a>.
         </p>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1982,7 +1982,7 @@
           <pre class="idl">dictionary RTCDataChannelStats : RTCStats {
              DOMString           label;
              DOMString           protocol;
-             long                datachannelid;
+             long                datachannelId;
```
henbos: fooId should refer to a stats object, for example trackId or trackIds, but identifiers that are not for stats objects but say the underlying webrtc object should have fooIdentifier, for example the MediaStream.id is surfaces as RTCMediaStreamStats.streamIdentifier to not be confused with the id of the RTCMediaStreamStats dictionary.

This should probably be changed to datachannelIdentifier and we should update the example in Guidelines for designing stats objects which use datachannelid as an example. This section should talk about "Identifiers"

henbos: Or dataChannelIdentifier based on RTCDataChannel having capitalized C

vr000m: +1 dataChannelIdentifier

soareschen: Ah I see. I have changed it to `dataChannelIdentifier` then. I have also updated the "Names ending in "Id"" example to refer to `transportId` instead of `datachannelId` for consistency.

vr000m:
@jan-ivar any comments? just want to get more eyes on this bit, since it might break working implementations.

jan-ivar:
FWIW Firefox does not implement RTCDataChannelStats yet, so we're fine with this.

PR #243, opened 2017-09-12
Added Guidelines for getStats() results caching

### LINKED ISSUE #177: Discuss caching and consistency of getStats() return
If the implementation caches getStats() results, there may be inconsistencies - for instance, if the application calls addTrack(), it would expect to see the new track in the stats immediately.

This needs discussion in the implementation guidelines section.

henbos:
Fixes #177

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -278,11 +278,13 @@
           calls for performance benefits, however any implementation must adhere to the following:
         </p>
         <p>
-          When the state of the PeerConnection visible changes as a result of an API call, a
+          When the state of the PeerConnection visibly changes as a result of an API call, a
           promise resolving or an event firing, subsequent getStats() calls MUST return up-to-date
-          dictionaries for the affected objects. For example, if a track is added (call to
-          addTrack() or the setRemoteDescription() promise resolving or associated events firing)
-          this is reflected in subsequent getStats() calls.
+          dictionaries for the affected objects. For example, if a track is added with addTrack()
+          subsequent getStats() calls MUST resolve with a corresponding RTCMediaStreamTrackStats
+          object. If you call setRemoteDescription() removing a remote track, upon the promise
+          resolving or an associated event (stream's onremovetrack or track's onmute) firing,
+          calling getStats() MUST resolve with an up-to-date RTCMediaStreamTrackStats object.
         </p>
       </section>
     </section><!-- VS: getStats() remains in base spec! -->
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -268,6 +268,26 @@
           PeerConnection.
         </p>
       </section>
+      <section>
+        <h3>
+          Guidelines for getStats() results caching
+        </h3>
+        <p>
+          A call to getStats() touches many components of WebRTC and may take significant time to
+          execute. An implementation may utilize caching for performance benefits and return the
+          same result as a previous getStats() call.
+        </p>
+        <p>
+          The cache MUST NOT be valid for more than 1 second. The recommended time for cache
+          consistency is no greater than 200 ms.
+        </p>
+        <p>
+          API calls that modify the state of the PeerConnection or its components MUST invalidate
+          the cache. For example, calling getStats() immediately after addTrack() or
+          setRemoteDescription() has been resolved is guaranteed to reflect the changes that these
+          calls had on the PeerConnection.
+        </p>
+      </section>
     </section><!-- VS: getStats() remains in base spec! -->
     <section>
       <h2>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -279,12 +279,12 @@
         </p>
         <p>
           When the state of the PeerConnection visibly changes as a result of an API call, a
-          promise resolving or an event firing, subsequent new getStats() calls MUST return up-to-date
+          promise resolving or an event firing, subsequent new getStats() calls must return up-to-date
           dictionaries for the affected objects. For example, if a track is added with addTrack()
-          subsequent getStats() calls MUST resolve with a corresponding RTCMediaStreamTrackStats
+          subsequent getStats() calls must resolve with a corresponding RTCMediaStreamTrackStats
           object. If you call setRemoteDescription() removing a remote track, upon the promise
           resolving or an associated event (stream's onremovetrack or track's onmute) firing,
-          calling getStats() MUST resolve with an up-to-date RTCMediaStreamTrackStats object.
+          calling getStats() must resolve with an up-to-date RTCMediaStreamTrackStats object.
         </p>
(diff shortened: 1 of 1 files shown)


jan-ivar:
How does this address @alvestrand 's concerns in https://github.com/w3c/webrtc-stats/issues/177#issue-211319269 ?

@henbos Why not encourage throttling of the requests instead? They're asynchronous after all.

henbos:
What do you mean @jan-ivar ? After a PC state change (e.g. addTrack or SRD), instead of invalidating the cache you wait to do the stats collection until the cache has become invalidated on its own?

jan-ivar:
@henbos Sorry, I was reacting to this being the first mention of any caching at all. In particular the introduction of this (non-normative?) allowance: *"An implementation MAY utilize caching for performance benefits and return the same result as a previous getStats() call. "*

Is there normative support for this? This would seem to conflict with many MUST statements elsewhere in this spec and the WebRTC spec, and as such seems ambiguous, and best avoided.

To the extent we want to disallow caching, I think specs do this by specifying the invariant guarantees an API MUST offer. These become the guarantees that must not be broken by any implementation optimizations; they define the boundaries within which implementations are implicitly allowed to experiment with things like caching, without a spec endorsing caching, or ingraining it in the API.

If there are guarantees missing in this spec, I'm all for adding them. I was hoping they were there already, but if they're unclear, I'm all for making them more explicit.

jan-ivar:
From a practical view, I don't think we should optimize aggressive JavaScript, or spec how to do so.

E.g. if a page is aggressively calling `getStats` in a tight loop, the answer might not be to optimize those calls, but instead throttle those calls by delaying results. Any backlog of in-parallel async requests that are more frequent than we can handle, can be satisfied with the same results, without breaking any spec guarantees.

vr000m:
I like the idea of returning the old data if someone is calling the API too often. I believe the timestamp of the objects will remain the same, which makes it easier for the developer to compare with the previous set of stats, etc.

vr000m:
Specifically, do we need cache timing values, would it not make the client code a bit more complex?

jan-ivar:
@vr000m I can't find anything in the spec that prevents implementations from optimizing by returning the same data as a previous call, so aren't we fine there?

I agree the `timestamp` should not increment in those cases. We could add text to [timestamp](http://w3c.github.io/webrtc-pc/#dom-rtcstats-timestamp) to clarify this. That would make timestamps for local stats more consistent with timestamps for remote stats, where we already clarify. Maybe something like *"For local stats, the timestamp should reflect the most recentness of the data set from this object"* ?

vr000m:
most recentness, or the timestamp they were last updated by the underlying objects?

vr000m:
@jan-ivar could you create an issue and PR for the local timestamp? so that we can discuss this separately.

webrtc-stats.html
```diff
@@ -268,6 +268,26 @@
...
+          Guidelines for getStats() results caching
+        </h3>
+        <p>
+          A call to getStats() touches many components of WebRTC and may take significant time to
+          execute. An implementation may utilize caching for performance benefits and return the
+          same result as a previous getStats() call.
+        </p>
+        <p>
+          The cache MUST NOT be valid for more than 1 second. The recommended time for cache
+          consistency is no greater than 200 ms.
```
henbos: The problem we're trying to solve is when something is doing something like
addTrack()
getStats()
And expecting that the thing you changed has shown up.

We should rewrite the guidelines not to say anything about how often the stats update - that is up to the engine - but we should guarantee that objects added/removed are reflected in subsequent getStats calls.

henbos: For specific events that aren't the direct result of an API call, such as camera dying, then on an observable event like onmuted, subsequent getStats calls need to give updates values for that object.

henbos: We can add a comment like that for the RTCMediaStreamTrackStats dictionary, and for the guidelines section here say "unless otherwise specified..."

webrtc-stats.html
```diff
@@ -268,6 +268,23 @@
...
+        <h3>
+          Guidelines for getStats() results caching/throttling
+        </h3>
+        <p>
+          A call to getStats() touches many components of WebRTC and may take significant time to
+          execute. The implementation may or may not utilize caching or throttling of getStats()
+          calls for performance benefits, however any implementation must adhere to the following:
+        </p>
+        <p>
+          When the state of the PeerConnection visible changes as a result of an API call, a
```
vr000m: visibly

henbos:
PTAL @jan-ivar, @vr000m

alvestrand:
Looks good to me (better than no advice).

webrtc-stats.html
```diff
@@ -268,6 +268,25 @@
...
+          Guidelines for getStats() results caching/throttling
+        </h3>
+        <p>
+          A call to getStats() touches many components of WebRTC and may take significant time to
+          execute. The implementation may or may not utilize caching or throttling of getStats()
+          calls for performance benefits, however any implementation must adhere to the following:
+        </p>
+        <p>
+          When the state of the PeerConnection visibly changes as a result of an API call, a
+          promise resolving or an event firing, subsequent getStats() calls MUST return up-to-date
```
jan-ivar: I think this is fine (and non-normative). I'm assuming this means "subsequent *new* getStats() calls". For in-flight calls, I don't think we have or want any guarantees.

henbos: Indeed. Changed it to "subsequent new getStats() calls"

henbos:
Let's merge this when the checks pass.

alvestrand:
Rebase, and it should pass.

webrtc-stats.html
```diff
@@ -268,6 +268,25 @@
...
+          calls for performance benefits, however any implementation must adhere to the following:
+        </p>
+        <p>
+          When the state of the PeerConnection visibly changes as a result of an API call, a
+          promise resolving or an event firing, subsequent new getStats() calls MUST return up-to-date
+          dictionaries for the affected objects. For example, if a track is added with addTrack()
+          subsequent getStats() calls MUST resolve with a corresponding RTCMediaStreamTrackStats
+          object. If you call setRemoteDescription() removing a remote track, upon the promise
+          resolving or an associated event (stream's onremovetrack or track's onmute) firing,
+          calling getStats() MUST resolve with an up-to-date RTCMediaStreamTrackStats object.
```
jan-ivar: Should be lower-case MUST throughout since this section is non-normative.

vr000m:
TPAC17: @jan-ivar could you confirm that this is good to go.. The authors are ready to merge this.

alvestrand:
@henbos will update, rebase and merge.

webrtc-stats.html

Issue #99, opened 2016-12-12
Privacy & Security self review


dontcallmedom:
It would be good to review the document through the [privacy & security questionnaire](https://w3ctag.github.io/security-questionnaire/) - this would in particular help for the wide review of the document pre-CR.

hadleybeeman:
Hello! I'm sure this will come up in your self-review, but I'd be interested in hearing your thoughts on if/how you think this feature might expand the fingerprinting surface area for identifying a browser/user.

alvestrand:
There's a few words in https://w3c.github.io/webrtc-stats/#security-considerations (calling out IP addresses as a consideration). There are other attributes that will expand fingerprinting surface (such as codec implementation strings). I don't see how much mitigation we  can do here, either - these stats are all needed for operation of services.

hadleybeeman:
Thanks for the quick response, @alvestrand! We will look forward to seeing more detail in your response to the privacy and security questionnaire. We are interested in things like whether the latency exposed by `roundTripTime` in `RTCOutboundRTPStreamStats` or `RTCIceCandidatePairStats` might expose something about the user's distance from the server (and therefore, where in the world they are). 

We do understand that mitigations can be tricky. It might be helpful to have different levels of statistics based on the privacy invasiveness... for example, the media-layers stats might be different to the transport-layer stats? Where mitigation isn't possible, exposing all the unmitigated threats in the spec will help implementers make informed decisions about how to present them to users.

hadleybeeman:
Hi all — we are meeting again and checking on the progress of things. How is this going? Can we help at all?

hadleybeeman:
@dontcallmedom Hello! We're just looking at this in our TAG review. It seems that no one is assigned to it... should there be someone?  We're not sure who to work with. Thanks for any help. :)

dontcallmedom:
the chairs, editors and I plan on getting a first stab at the self review in the upcoming few weeks

hadleybeeman:
That's great, @dontcallmedom. Thanks for the quick reply! We'll check in with you after that then.

dontcallmedom:
Here is a first stab at reviewing the spec through the questionnaire for discussion tomorrow with @vr000m and @aboba.

I think the relevant [questions in the questionnaire](https://w3ctag.github.io/security-questionnaire/#questions) for this spec are:
* personally identifiable information
* persistent state
* cross-origin perstatent state
* access to new data
* new exposure on local device configuration?
[...]

PR #251, opened 2017-10-20
 Complete security and privacy considerations based on self-review

dontcallmedom:
based on analysis jointly conducted with @vr000m @alvestrand @aboba

### DIFF
--- a/security-questionnaire.md
+++ b/security-questionnaire.md
@@ -20,7 +20,7 @@ Not beyond what is already exposed by the [WebRTC 1.0] specification itself
 
 The properties exposed by `RTCReceivedRTPStreamStats`, `RTCRemoteInboundRTPStreamStats`, `RTCSentRTPStreamStats`, `RTCOutboundRTPStreamStats`, `RTCRemoteOutboundRTPStreamStats`, `RTCIceCandidatePairStats`, `RTCTransportStats` expose network-layer data not currently available to the JavaScript layer.
 
-In addition, the behavior of WebRTC Stats (in particular `RTCMediaStreamTrackStats`) on [isolated media streams] (which aren't otherwise available to the JavaScript layer) is currently ill-defined.
+In addition, we expect that the behavior of WebRTC Stats (in particular `RTCMediaStreamTrackStats`) on [isolated media streams] (which aren't otherwise available to the JavaScript layer) will need further work (see https://github.com/w3c/webrtc-pc/issues/1613).
 
 ## [3.6 Does this specification enable new script execution/loading mechanisms?]
 
--- a/security-questionnaire.md
+++ b/security-questionnaire.md
@@ -1,0 +1,99 @@
+# [Security and Privacy Self-Review Questionnaire] for [Identifiers for WebRTC's Statistics API]
+
+## [3.1 Does this specification deal with personally-identifiable information?]
+
+Not beyond what is already exposed in [WebRTC 1.0].
+
+## [3.2 Does this specification deal with high-value data?]
+
+No
+
+## [3.3 Does this specification introduce new state for an origin that persists across browsing sessions?]
+
+No.
+
+## [3.4 Does this specification expose persistent, cross-origin state to the web?]
+
+Not beyond what is already exposed by the [WebRTC 1.0] specification itself
+
+## [3.5 Does this specification expose any other data to an origin that it doesn’t currently have access to?]
+
+The properties exposed by `RTCReceivedRTPStreamStats`, `RTCRemoteInboundRTPStreamStats`, `RTCSentRTPStreamStats`, `RTCOutboundRTPStreamStats`, `RTCRemoteOutboundRTPStreamStats`, `RTCIceCandidatePairStats`, `RTCTransportStats` expose network-layer data not currently available to the JavaScript layer.
+
+In addition, the behavior of WebRTC Stats (in particular `RTCMediaStreamTrackStats`) on [isolated media streams] (which aren't otherwise available to the JavaScript layer) is currently ill-defined.
+
+## [3.6 Does this specification enable new script execution/loading mechanisms?]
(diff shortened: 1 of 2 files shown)


dontcallmedom:
see also #99

webrtc-stats.html
```diff
@@ -2936,13 +2936,13 @@
...
+      <p>In addition, the properties exposed by <a>RTCReceivedRTPStreamStats</a>, <a>RTCRemoteInboundRTPStreamStats</a>, <a>RTCSentRTPStreamStats</a>, <a>RTCOutboundRTPStreamStats</a>, <a>RTCRemoteOutboundRTPStreamStats</a>, <a>RTCIceCandidatePairStats</a>, <a>RTCTransportStats</a> expose network-layer data not currently available to the JavaScript layer.</p>
+      <p>Beyond the risks associated with revealing <a data-cite="WEBRTC#revealing-ip-addresses">IP addresses as discussed in the WebRTC 1.0 specification</a>, some combination of the network properties uniquely exposed by this specification can be correlated with  location.</p>
+      <p>For instance, the round-trip time exposed in <a>RTCRemoteInboundRTPStreamStats</a> can give some coarse indication on how far aparts the peers are located, and thus, if one of the peer's location is known, this may reveal information about the other peer.</p>
+      <p class="issue">TODO: Clarify impact of <a>RTCMediaStreamTrackStats</a> on <a data-cite="WEBRTC#isolated-media-streams">isolated media streams</a>.</p>
```
alvestrand: Suggested additional text:
When applied to isolated streams, media metrics may allow an application to infer some characteristics of the isolated stream, such as if anyone is speaking (by watching the "volume" statistic).
TODO: Consider whether certain stats need to be not reported on isolated streams.

dontcallmedom: do you mean in replacement of the existing TODO, or in addition to it? (replacing it with your text sounds good tom fwiw)

security-questionnaire.md
```diff
@@ -0,0 +1,99 @@
...
+
+## [3.4 Does this specification expose persistent, cross-origin state to the web?]
+
+Not beyond what is already exposed by the [WebRTC 1.0] specification itself
+
+## [3.5 Does this specification expose any other data to an origin that it doesn’t currently have access to?]
+
+The properties exposed by `RTCReceivedRTPStreamStats`, `RTCRemoteInboundRTPStreamStats`, `RTCSentRTPStreamStats`, `RTCOutboundRTPStreamStats`, `RTCRemoteOutboundRTPStreamStats`, `RTCIceCandidatePairStats`, `RTCTransportStats` expose network-layer data not currently available to the JavaScript layer.
+
+In addition, the behavior of WebRTC Stats (in particular `RTCMediaStreamTrackStats`) on [isolated media streams] (which aren't otherwise available to the JavaScript layer) is currently ill-defined.
```
henbos: Can you clarify what you mean by "the behavior of WebRTC Stats (...) is currently ill-defined"?

dontcallmedom: it's not clear how much statistics should be reported on isolated media streams given that these statistics leak information about their content - see https://github.com/w3c/webrtc-pc/issues/1613

security-questionnaire.md
```diff
@@ -0,0 +1,99 @@
...
+
+## [3.6 Does this specification enable new script execution/loading mechanisms?]
+
+No.
+
+## [3.7 Does this specification allow an origin access to a user’s location?]
+
+Beyond the risks associated with IP addresses as discussed in the WebRTC 1.0 specification, some combination of some of the network properties uniquely exposed by this specification can be correlated with  location.
+
+For instance, the [round-trip time] exposed in `RTCRemoteInboundRTPStreamStats` can give some coarse indication on how far aparts the peers are located, and thus, if one of the peer's location is known, this may reveal information about the other peer.
```
vr000m: There are many ways the latency can be calculated, so I do not know specifically calling out RTT makes sense here.
see https://w3c.github.io/resource-timing/#introduction, shows how you can use resource timing API to calculate latency using HTTP.

webrtc-stats.html
alvestrand: Replacing. I think impact is clear enough with the suggested text that we can drop the TODO for more clarification.

security-questionnaire.md
alvestrand: The RTT gives you a metric for client-to-client distance, while HTTP just gives you client-to-server distance. That might matter a little.
You can always measure RTT using datachannel (just send a message and have it returned), so hiding RTT doesn't mean RTT can't be observed (or at least given an upper bound).

security-questionnaire.md
alvestrand: I think the current behavior is perfectly defined (the same stats are available as for other tracks), but there's a good chance that this is not what we want.
My suggestion: s/is currently ill-defined/is a topic that needs further discussion, since the current lack of special treatment for these streams may raise issues/.

alvestrand:
Waiting for Dom's response to comments.

security-questionnaire.md
dontcallmedom: I corrected this into "we expect that the behavior of WebRTC Stats (in particular `RTCMediaStreamTrackStats`) on [isolated media streams] (which aren't otherwise available to the JavaScript layer) will need further work (see https://github.com/w3c/webrtc-pc/issues/1613)"

webrtc-stats.html
dontcallmedom: I've adopted this text in my latest revision

security-questionnaire.md
vr000m: @alvestrand that is correct, RTT can be observed by implementing a ping mechanism within the datachannel.

alvestrand:
No specific changes suggested from TPAC. Seems ready to merge.

PR #259, opened 2017-10-23
Adding "networkType" field to RTCIceCandidateStats.

### LINKED ISSUE #257: Add stat for network type of ICE candidate
It would be pretty useful to know what type of network a local ICE candidate is using (for example, Wi-Fi, Ethernet, cellular, etc.).

Here's an enum we may be able to use: http://wicg.github.io/netinfo/#dom-connectiontype

Or at least copy, assuming referencing a WICG standard isn't allowed.

taylor-b:
Fixes #257.

Exposes the type of network interface a local ICE candidate is using.
Defined such that it will be present for srflx/relay candidates as well,
exposing the type of network interface used to communicate with the
STUN/TURN server.

RTCNetworkType is based on this enum:
http://wicg.github.io/netinfo/#dom-connectiontype

With the only differences being that "none" isn't applicable, since every
ICE candidate is using *some* network interface, and "mixed" isn't
applicable since every ICE candidate is tied to only one interface.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2377,10 +2377,10 @@ enum RTCNetworkType {
     "bluetooth",
     "cellular",
     "ethernet",
-    "other",
-    "unknown",
     "wifi",
-    "wimax"
+    "wimax",
+    "other",
+    "unknown"
 };</pre>
             <table data-link-for="RTCNetworkType" data-dfn-for=
             "RTCNetworkType" class="simple">
@@ -2422,6 +2422,26 @@ enum RTCNetworkType {
                 </tr>
                 <tr>
                   <td>
+                    <dfn><code>wifi</code></dfn>
+                  </td>
+                  <td>
+                    <p>
+                      A Wi-Fi connection.
+                    </p>
+                  </td>
+                </tr>
+                <tr>
+                  <td>
+                    <dfn><code>wimax</code></dfn>
+                  </td>
+                  <td>
+                    <p>
+                      A WiMAX connection.
+                    </p>
+                  </td>
+                </tr>
+                <tr>
+                  <td>
                     <dfn><code>other</code></dfn>
                   </td>
                   <td>
@@ -2439,26 +2459,6 @@ enum RTCNetworkType {
                     <p>
                       The user agent is unable (or unwilling) to determine the
                       underlying connection technology.
-                    </p>
-                  </td>
-                </tr>
-                <tr>
-                  <td>
-                    <dfn><code>wifi</code></dfn>
-                  </td>
-                  <td>
-                    <p>
-                      A Wi-Fi connection.
-                    </p>
-                  </td>
-                </tr>
-                <tr>
-                  <td>
-                    <dfn><code>wimax</code></dfn>
-                  </td>
-                  <td>
-                    <p>
-                      A WiMAX connection.
                     </p>
                   </td>
                 </tr>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2222,6 +2222,7 @@ enum RTCStatsType {
           <pre class="idl">dictionary RTCIceCandidateStats : RTCStats {
              DOMString                transportId;
              boolean                  isRemote;
+             RTCNetworkType           networkType;
              DOMString                ip;
              long                     port;
              DOMString                protocol;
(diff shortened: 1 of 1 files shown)


alvestrand:
Questions:
- Should there be a "won't say" network type?
- Are VPN connections grouped with the native API they run over? (Does this leak any sensitive information? My instinct says "no" - in most cases, it would be obvious anyway).
As a matter of good enum design, I'd place "unknown" and "other" first in the enum.
Do enums have extensibility this week? I know it's been proposed, but don't know the status of it. If possible, we should make it clear that this enum is extensible.

taylor-b:
> Should there be a "won't say" network type?

The idea is that "unknown" would be used for that.

> Are VPN connections grouped with the native API they run over?

This would make sense, if the browser has enough information to know.

> As a matter of good enum design, I'd place "unknown" and "other" first in the enum.

Hmm. Would last be better? It makes sense to see "other" after all the available options.

> Do enums have extensibility this week?

I figured it would be easy to extend it in this specification itself when we need to. Or is that what you're talking about? I thought the only issue was extensibility across different specifications.

taylor-b:
Actually, about VPNs: Would it make sense to treat a VPN as a different network type? Knowing that traffic is going through a VPN may be more valuable from a statistics point of view than knowing what type of local network interface is used. And I doubt it's even possible to detect what type of network interface a VPN is using, on most OSes. The "goog" stat we have right now does treat "vpn" as a separate type.

vr000m:
I would recommend a different metric for VPN, in case an endpoint is able to report that the VPN is over eth or wifi.

alvestrand:
My thinking about the beginning vs end was with extensibility, but when I think about it again, I realize that I was thinking C++ enums that map to numbers; WebIDL enums map to strings, so order really doesn't matter.

wrt partial enums, they don't seem to have made it into WebIDL so far.

https://www.w3.org/Bugs/Public/show_bug.cgi?id=27048 is a bug filed in 2014 about it.

wrt vpn: my thinking is that if we know something's going out over Ethernet, we should say so, even if it's vpn-encapsulated. But it makes sense to me to differentiate between "can't tell" and "can't tell because it's a VPN", so having a vpn member for the case where we can't tell what's the real interface makes sense.

(My thinking of usage of this enum is that it will be used for collecting stats on "what's the typical distribution of bandwidths when people have an Ethernet interface vs a WiFi interface" - VPN is more useful than "unknown", but less useful than the physical-interface distinction.)

taylor-b:
> wrt vpn: my thinking is that if we know something's going out over Ethernet, we should say so, even if it's vpn-encapsulated.

Is there any situation where we *would* know, though? Or is this just hypothetical?

webrtc-stats.html
```diff
@@ -2345,6 +2367,105 @@
...
+                  </td>
+                </tr>
+                <tr>
+                  <td>
+                    <dfn><code>other</code></dfn>
+                  </td>
+                  <td>
+                    <p>
+                      The connection type is known, but isn't one of the
+                      enumerated types.
```
henbos: What might fall into this category? If something is "other", would it make sense to add it to the spec or is "other" something we want to use?

taylor-b: Hopefully nothing with statistical significance. But there certainly are more physical-layer protocols than are listed here.

I'd be up for just merging "other" and "unknown", personally. Different browser versions may have different breakdowns of "other" vs "unknown" which doesn't make it very useful as a standard stat.

henbos: Merging SGTM

alvestrand: Let's merge them - "unknown to stats writer". Might want to add a note that this list is expected to be updated when new connectivity types are defined, and hope that coders who would like to use "other" instead send a PR in our direction.

alvestrand:
@taylor-b You know the APIs available on the OSes we deploy on better than I do - if you say that we can tell if an interface is a VPN, but not whether the packets for that VPN will be routed over the Ethernet or the 3G connection (if both are up), I am happy with saying "VPN is a separate interface type".

alvestrand:
if you rebase this, that should pull in the updated travis.yml files, which makes the checking use npm 8, which hopefully fixes the travis error.

webrtc-stats.html
vr000m: +1 to merge the  unknown/other.

garyliu33:
When can we get it merged?

alvestrand:
I've rebased this as #274 (no changes). Will integrate comments from this thread on that PR.

alvestrand:
Replaced by #274

Issue #235, opened 2017-08-16
Is keeping stats around a memory problem?


stefhak:
This surfaced in a discussion in a webrtc-pc [Issue](https://github.com/w3c/webrtc-pc/issues/1507). Basically the question is whether keeping stats for objects that are no longer in use will cause problems from a memory perspective. Please refer to https://github.com/w3c/webrtc-pc/issues/1507 for more input.

jesup:
> But Dan's point about stats is that we already have another problem about memory use increasing indefinitely. Stats for deleted objects are kept around forever:

>>    The general rule is that stats objects, once created, exist for the lifetime of the PeerConnection that contains them, even when the underlying object they report on is stopped or deleted. This is important in order to report consistently on short-lived objects and to be able to report totals over the lifetime of a PeerConnection.

Unless I'm missing something (possible), the memory usage should be proportional to the objects in a peerconnection.  unless the number of objects grows without bound, we won't have the stats growing without bound.  If you add a new track, there may/will be a new trackstats object as well - but not an infinite number.  Perhaps some of the candidate/transport objects might be proportional to the number of ICE restarts/renegotiations. 

The memory may be in use until the peerconnection itself has gone away. (Note: gone away doesn't mean closed.)

taylor-b:
The number of objects *can* theoretically grow without bound. Suppose I join a persistent group call where participants are regularly leaving and joining, each time using a new SSRC, resulting in a new stream stats object. The list of historical `RTCInboundRtpStreamStats` objects kept around would continue to grow.

stefhak:
@taylor-b I guess the provider of the service would be aware, and could work around (e.g. by having the client close and de-reference the PeerConnection and creating a new one when the number of unused objects grow over a certain threshold)?

taylor-b:
@stefhak I'm sure that's possible; people have gone to such lengths just to work around bugs. I'd also assume that the amount of applications that will run into this problem will be relatively low.

But I wonder if anyone would be opposed to adding a `clearOldStats` method to `RTCPeerConnection`. I can't imagine it being difficult to implement.

vr000m:
My challenge with `clearOldStats` is, there are several third-party libraries polling getStats, any one (app or third-party) calling clearOldStats would mean loss of information for any of those libraries.

taylor-b:
Good point. I guess another approach would be to do something like this:

[...]

PR #262, opened 2017-10-24
Added 'objectDeleted' attribute

### LINKED ISSUE #131: Consistent marker for "non-active" object?
We have a number of objects where stats go into a "frozen" mode once it's detached, deleted, no longer used or somethign like that.

Should there be a toplevel marker ("active default true" or something like that) so that we can filter all "inactive" stats consistently?

alvestrand:
Fixes #131 
Note: Attribute could have been named "deleted", but that name is already in use in the "candidate" sub-directory with a different meaning.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -193,7 +193,6 @@
       </p>
       <section>
       <h3>Additional RTCStats attributes that are defined in this specification</h3>
-      </p>
       <pre class="idl">
         partial dictionary RTCStats {
         boolean objectDeleted = false;
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -192,6 +192,37 @@
         a local clock.
       </p>
       <section>
+      <h3>Additional RTCStats attributes that are defined in this specification</h3>
+      </p>
+      <pre class="idl">
+        partial dictionary RTCStats {
+        boolean objectDeleted = false;
+        };
+      </pre>
+      <p>The following housekeeping objects are added to <code>RTCStats</code>:
+      </p>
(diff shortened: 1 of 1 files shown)


vr000m:
TPAC17: This is blocked on the status of https://github.com/w3c/webrtc-stats/issues/235, which was reopened based on the feedback from Mozilla.

vr000m:
@jan-ivar can you comment on this?

alvestrand:
Approved for merge (after rebase). If changes are done because of #235, we'll revisit.

jan-ivar:
Sorry for the late reply, but I don't approve.

Putting `objectDeleted` in the base class may be an implementation convenience, but it alters the contract of every stats object, and implies that implementations suddenly may return a whole second set of objects with this boolean set to `true`, which would break existing code (which assumes only non-deleted obects). That's a big API change to foist on users IMHO. Users now essentially need to test this member for every stats object before reading or trusting its data to be current, whereas before they didn't need to do this.

Do we really need this for all stats objects? What sub-set of objects can be deleted?

From an API point alone, overloading an API with a boolean is generally considered bad design. Would it be cleaner, clearer and more performant to introduce a `getDeletedStats()` function?

We broke `getSources()`  into `getContributingSources()` & `getSynchronizationSources()` over less.

alvestrand:
Jan-ivar, implementations don't "suddenly return a whole second set of objects" (as in a runtime change). They either have preserved objects after object deletion (which the spec *always* required, even though implementations have varied), or they don't.
*All* objects apart from the PeerConnection can be deleted.
Let's continue the discussion of the API we want under #235.

PR #267, opened 2017-10-26
jitterBufferDelay in seconds

### LINKED ISSUE #244: RTCMediaStreamTrackStats.jitterBufferDelay: Specify unit in seconds
jitterBufferDelay should be the delay in seconds, it just says the "total time" now.

henbos:
Fixes #244

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1970,8 +1970,8 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  It is the total time in seconds each audio sample or video frame takes from the
-                  time it is received to the time it is rendered. The delay is measured from the
+                  It is the sum of the time, in seconds, each audio sample or video frame takes from
+                  the time it is received to the time it is rendered. The delay is measured from the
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1970,12 +1970,12 @@
               </dt>
               <dd>
                 <p>
-                  It is the total time each audio sample or video frame takes from the time it is
-                  received to the time it is rendered. The delay is measured from the time the
-                  first packet belonging to an audio/video frame enters the jitter buffer to the
-                  time the complete frame is sent for rendering after decoding. The average jitter
-                  buffer delay can be calculated by dividing the <code>jitterBufferDelay</code>
-                  with the <code>framesDecoded</code> (for video) or
+                  It is the total time in seconds each audio sample or video frame takes from the
```
alvestrand: While we're here, can we say "It is the sum of the time, in seconds, that each audio sample or video frame spends from the time it is received to the time it is rendered, across all frames/samples."?
The fact that it's a sum isn't immediately obvious from the text.

henbos: Done.

PR #268, opened 2017-10-26
jitterBufferEmittedCount added (jitterBufferOutput)

### LINKED ISSUE #246: jitterBufferDelay and concealed samples, DTX/CNG samples
jitterBufferDelay is total time each audio or video samples has spent in the jitter buffer before being decoded and sent out for rendering.

For video, average time spent in jitter buffer is jitterBufferDelay/framesDecoded. This makes sense.

For audio, it is said to be jitterBufferDelay/totalSamplesReceived. Possible problems:
- totalSamplesReceived include concealedSamples, which have spent no time in the jitter buffer. Should we instead divide by (totalSamplesReceived - concealedSamples)?
- We may have received samples that have not been decoded and ready to be played out, and these samples have not contributed to jitterBufferDelay yet. There is no "samplesDecoded", is totalSamplesReceived >= the value we should divide by?
- How are DTX/CNG samples handled? Do these contribute to jitterBufferDelay? Do these contribute to totalSamplesReceived?

henbos:
Fixes #246

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1972,10 +1972,10 @@ enum RTCStatsType {
               <dd>
                 <p>
                   It is the total time each audio sample or video frame takes from the time it is
-                  received to the time it is rendered. The delay is measured from the time the
-                  first packet belonging to an audio/video frame is received to the time the
-                  complete frame is sent for rendering after decoding. The average jitter buffer
-                  delay can be calculated by dividing the <code>jitterBufferDelay</code> with the
+                  received and enters the jitter buffer to the time it exits the jitter buffer. This
+                  increases upon samples/frames exiting, having completed their time in the buffer
+                  (and increasing <code>jitterBufferOutput</code>). The average jitter buffer delay
+                  can be calculated by dividing the <code>jitterBufferDelay</code> with the
                   <code>jitterBufferOutput</code>.
                 </p>
               </dd>
@@ -1985,9 +1985,8 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  The total number of audio samples or video frames that have exited the jitter
-                  buffer (and increased the <code>jitterBufferDelay</code> by the time spent in the
-                  jitter buffer).
+                  The total number of audio samples or video frames that have come out of the jitter
+                  buffer (increasing the <code>jitterBufferDelay</code>).
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1974,7 +1974,7 @@ enum RTCStatsType {
                   It is the total time each audio sample or video frame takes from the time it is
                   received and enters the jitter buffer to the time it exits the jitter buffer. This
                   increases upon samples/frames exiting, having completed their time in the buffer
-                  (and increasing <code>jitterBufferOutput</code>). The average jitter buffer delay
+                  (incrementing <code>jitterBufferOutput</code>). The average jitter buffer delay
                   can be calculated by dividing the <code>jitterBufferDelay</code> with the
                   <code>jitterBufferOutput</code>.
                 </p>
@@ -1986,7 +1986,7 @@ enum RTCStatsType {
               <dd>
                 <p>
                   The total number of audio samples or video frames that have come out of the jitter
-                  buffer (increasing the <code>jitterBufferDelay</code>).
+                  buffer (increasing <code>jitterBufferDelay</code>).
                 </p>
               </dd>
               <dt>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1971,12 +1972,21 @@
               <dd>
                 <p>
                   It is the total time each audio sample or video frame takes from the time it is
-                  received to the time it is rendered. The delay is measured from the time the
-                  first packet belonging to an audio/video frame enters the jitter buffer to the
-                  time the complete frame is sent for rendering after decoding. The average jitter
-                  buffer delay can be calculated by dividing the <code>jitterBufferDelay</code>
-                  with the <code>framesDecoded</code> (for video) or
-                  <code>totalSamplesReceived</code> (for audio).
+                  received and enters the jitter buffer to the time it exits the jitter buffer. This
```
alvestrand: Just delete "and enters the jitter buffer".

webrtc-stats.html
```diff
@@ -1971,12 +1972,21 @@
...
+                  increases upon samples/frames exiting, having completed their time in the buffer
+                  (incrementing <code>jitterBufferOutput</code>). The average jitter buffer delay
+                  can be calculated by dividing the <code>jitterBufferDelay</code> with the
+                  <code>jitterBufferOutput</code>.
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>jitterBufferOutput</code></dfn> of type <span class=
```
alvestrand: The name doesn't sound right to me, but I can't think of a better name, because we want it to cover both samples and frames, and have all the jitterbuffer items start with "jitterBuffer".
jitterBufferEmitted? jitterBufferEmittedCount? jitterBufferEmittedSamplesCount? (too long!)

vr000m: Maybe we should have separate counters for frames (video) and samples (audio).

vr000m:
Is this replacing jitterBufferDelay with jitterBufferOutput.

henbos:
This is not replacing jitterBufferDelay, this gives us frames/samples counters that can be used to divide jitterBufferDelay with to get average delay over a period of time. The separate counters are because all the other frames/samples counters were not specific to "coming out of the jitter buffer" when the jitterBufferDelay goes up, and proposed counters had problems or relied on assumptions about events correlating.

vr000m:
What does one interpret with difference between the samples input to the jitter buffer and the output?

henbos:
Yes and even if you could discard concealed samples from framesDecoded (ugh, sometimes we talk about samples and sometimes frames, but whatever) what if there are decoding errors or what if frames are discarded for other reasons. Or what if there are frames/samples still being decoded, does the decoded counter increase before or after they are processed? Frames received could be dropped without ever entering the jitter buffer. Things like this could cause nonzero "3-2" or "3-1".

alvestrand:
@henbos, are you rebasing and addressing the comments?

webrtc-stats.html

webrtc-stats.html
henbos: I went with jitterBufferEmittedCount.

henbos:
PTAL

vr000m:
In the case of audio, does the application keep track of the sampleDuration? or just divide delay/samples?

henbos:
Is the audio sample duration (or video frame duration for that matter) relevant here? We aggregate delay for each sample going through the buffer.

vr000m:
> Yes and even if you could discard concealed samples from framesDecoded (ugh, sometimes we talk about samples and sometimes frames, but whatever) what if there are decoding errors or what if frames are discarded for other reasons. 

if there is a decoding error, the framesDecoded wont go up. Perhaps framesCorrupted would go up... 
Media-source API says: a corrupted frame may or may not be dropped, depending of the corruption and of the browser algorithm. so...?

> Or what if there are frames/samples still being decoded, does the decoded counter increase before or after they are processed? 

decoded counter goes up after decoding is complete. while it is being decoded, the jitter buffer emitted counter would go up... 

> Frames received could be dropped without ever entering the jitter buffer. Things like this could cause nonzero "3-2" or "3-1".

The receiver puts packets into the Jitter buffer not frames... at least most implementations I have worked with in the past... Can someone confirm? Typically, the marker bit in the RTP header indicates that it is last packet of the frame, if there are packets out-of-order, the jitter buffer is used to re-order and combine the packets in to frame. 

Some one needs to confirm, otherwise this opens up the question to when the following are the following counters updated:
+ framesReceived;
+ keyFramesReceived;
+ partialFramesLost;
+ fullFramesLost;
+ framesDropped;

In my mind, most of the counters are updated when the frame is emitted from the jitter buffer.

And the following are updated at the time of decoding:
+ framesCorrupted;
+ framesDecoded;

PR #270, opened 2017-10-27
RTCQualityLimitationReason and friends

### LINKED ISSUE #256: Stats for adaptation reason, for realsies
Merging all of these issues into one: #144, #160.

Alternatives:
- sequence (enum) - "CPU", "Bandwidth", "other".
- Pick the one that's obvious - "None", "CPU", "Bandwidth", "other"

googStats are a bunch of booleans that can be turned into a bitmask, and googAdaptationChanges counters.

henbos:
Fixes #256

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -666,7 +666,7 @@ enum RTCStatsType {
                 <p>
                   The definition of QP value depends on the <a>codec</a>; for VP8, the QP value is
                   the value carried in the frame header as the syntax element "y_ac_qi", and
-                  defined in [[RFC6386]] section 19.2. Its range is 0..127. 
+                  defined in [[RFC6386]] section 19.2. Its range is 0..127.
                   <!-- Need appropriate references for VP9 and H.264 -->
                 </p>
                 <p>
@@ -1216,6 +1216,10 @@ enum RTCStatsType {
              unsigned long        framesEncoded;
              double               totalEncodeTime;
              double               averageRTCPInterval;
+             RTCQualityLimitationReason  qualityLimitationReason;
+             double                      qualityCpuLimitationDuration;
+             double                      qualityBandwidthLimitationDuration;
+             double                      qualityTotalLimitationDuration;
 };</pre>
           <section>
             <h2>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1216,6 +1216,10 @@
              unsigned long        framesEncoded;
              double               totalEncodeTime;
              double               averageRTCPInterval;
+             RTCQualityLimitationReason  qualityLimitationReason;
+             double                      qualityCpuLimitationDuration;
+             double                      qualityBandwidthLimitationDuration;
+             double                      qualityTotalLimitationDuration;
```
henbos: ...or is record<DOMString, double> qualityLimitationDurations preferred?

henbos: And should we have enum qualityLimitationMethod = {none, framerate, resolution} (excluding quality due to https://github.com/w3c/webrtc-stats/issues/256#issuecomment-339838513) as per-Harald's suggestion? It sounds useful.

alvestrand: I'd like a record<reason, duration> rather than separate counters, because I suspect there will be more types added in the future (GPU restriction?) - a record is more future-proof in this regard than separate counters. I like breaking out the total limitation duration because I suspect that it will be the stat people look at first (during the call: "is it going up or not?")

alvestrand: record<> is used in the Fetch IDL in Chrome, so it's implemented and likely to stay around.

henbos: Done. A side-effect of this is that having record<reason, duration> implicitly means also having the duration for "none" unless we clarify what key values are expected to be implemented for the record. I'm not sure if I should say 'the key is any value of RTCQualityLimitationReason excluding the value "none"', say that "none" is optional to map, or not say anything.

A stats member is optional to implement, but if implemented it should be spec-compliant. What about record<> types? I expect you either implement none of it or all of it, so we should be clear about whether "none" needs to be implemented I suppose? PTAL and comment about this.

henbos:
Note, I wrote this (in case it got hidden by the fact that it was a comment on not the latest commit):
"Done. A side-effect of this is that having record<reason, duration> implicitly means also having the duration for "none" unless we clarify what key values are expected to be implemented for the record. I'm not sure if I should say 'the key is any value of RTCQualityLimitationReason excluding the value "none"', say that "none" is optional to map, or not say anything.

A stats member is optional to implement, but if implemented it should be spec-compliant. What about record<> types? I expect you either implement none of it or all of it, so we should be clear about whether "none" needs to be implemented I suppose? PTAL and comment about this."

webrtc-stats.html
```diff
@@ -1294,10 +1297,106 @@
...
+                </td>
+              </tr>
+              <tr>
+                <td>
+                  <dfn><code>bandwidth</code></dfn>
+                </td>
+                <td>
+                  <p>
+                    The resolution and/or framerate is primarily limited due to by bandwidth
+                    or packet loss.
```
vr000m: The resolution and/or framerate is primarily limited due to congestion cues during bandwidth estimation.

vr000m:
@lennart-csio PTAL

webrtc-stats.html
```diff
@@ -1216,6 +1216,9 @@ <h3>
              unsigned long        framesEncoded;
              double               totalEncodeTime;
              double               averageRTCPInterval;
+             RTCQualityLimitationReason                 qualityLimitationReason;
+             record&lt;RTCQualityLimitationReason, double&gt; qualityLimitationDurations;
+             double                                     qualityTotalLimitationDuration;
```
icydragons: For the counter, I would prefer it to be a counter when the resolution or fps changes due to an adaptation reason. Ie. if we went from 15fps to 30fps or if we went from VGA to QVGA. This will be the best for detecting issues that would be visible to a user, but not visible in the stats if the quality limitations is "flapping" between two resolutions. It matters not what caused the change, cpu, bw, other for that purpose.

webrtc-stats.html
```diff
@@ -1294,10 +1297,106 @@ <h2>
...
+                  Enumeration description
+                </th>
+              </tr>
+              <tr>
+                <td>
+                  <dfn><code>none</code></dfn>
+                </td>
+                <td>
+                  <p>
+                    The resolution and/or framerate is not limited.
```
icydragons: I think it's clearer if we make it such that none must be filled in if implemented or have a statement here that none shouldn't be filled in and the base for the time to use for this is the track time (or what the appropriate time to use as the difference).

I agree that having nearly any sane standard is better then having it up to each implementation.

webrtc-stats.html
icydragons: I like the above description. 

What would you think about expanding the dictionary and including bandwidth-interarrival, bandwidth-rtt,bandwith-loss, bandwidth-other in order to allow better classification of what piece of the network congestion control was enabled (and implementations could always just use just "bandwidth" if desired). For our customers, the thing they are most able to fix is network issues and having the type of network trouble clearer makes it much simpler to correct the right thing.

webrtc-stats.html
alvestrand: So you'd prefer "unsigned long qualityLimitationChangeCount" and increment it whenever qualityLimitationReason has a new value?
Works for me - the duration can be calculated by summing across the record, if needed, while the number of transitions can't e calculated in that way.

webrtc-stats.html
```diff
@@ -1216,6 +1216,9 @@ <h3>
              unsigned long        framesEncoded;
              double               totalEncodeTime;
              double               averageRTCPInterval;
+             RTCQualityLimitationReason                 qualityLimitationReason;
```
alvestrand: Should this have " = none" appended to make it clear that it's "none" by default?
Or should we make it explicitly "none" when none, so that we can tell that it's not supported when the value is missing?
Description should probably make this clear.

webrtc-stats.html
henbos: I don't see a way to get around the problem of "flapping" between different categories and numeric values (fps, resolution, bitrate) being continuous. What if framerate changes continuously akin to bitrate? You could try to make the categories more precise, such as "highly limited" or "slightly limited", but I don't think we can have counters for "number of times" something happens because the meaning of a high or low value of that would be implementation-specific. E.g. it might not go between 15 or 30 fps on an event, it might continuously set to every value in-between. In any case it might flap between different "reasons", or flap between limited and not limited.

Counters are problematic. Total duration works, and splitting up that duration into durations for different reasons gives you some analysis, but it does not reflect to what degree you are limited, only that you are. It explains _why_ other metrics are lower than expected, but it does not provide you the preciseness of those metrics.

If you want to know to what degree you are limited in something, or how the limitation manifests itself, like fps or resolution limitations, I think it makes more sense to poll getStats() for those metrics yourself. WDYT?

webrtc-stats.html
alvestrand: Deciding why bandwidth estimation took the decision it did seems to me to be a chancy process. If we want to do that, I think we should instrument the bandwidth estimator (likely on the ICETransport object), not the video track.

webrtc-stats.html
henbos: For "RTCQualityLimitationReason qualityLimitationReason" I explicitly state "or "none" if not limited". Only if not implemented or applicable would it be undefined (as all stats in this document can be in that case).

webrtc-stats.html
henbos: It's in the description, "or "none" if not limited".

webrtc-stats.html
henbos: I used vr00ms description. If needed, I'd prefer more details to be surfaced through other stats.

webrtc-stats.html
henbos: Removing qualityTotalLimitationDuration because it is calculable from `sum of qualityLimitationDurations - qualityLimitationDuration["none"]`.

henbos:
Note: See "outdated sections" to see all my responses, some of my comments are not visible by default.

vr000m:
Remind me, will this report only the current RTCQualityLimitationReason and duration for the current qualityLimitationDurations or current limitation and duration for all reasons?

For example, if I poll at the end of the call, I would see the last reason and duration for the last reason. although it could report the last reason and durations for all reasons observed during the session?

henbos:
All of them, I added "The record includes a mapping for all RTCQualityLimitationReason types." It's a total.

henbos:
Updated to:
```
<p>
  Only valid for video. A record of the total time, in seconds, that this stream
  has spent in each quality limitation state. The record includes a mapping for
  all <a>RTCQualityLimitationReason</a> types, including "none".
</p>
<p>
  The sum of all entries minus <code>qualityLimidationDurations["none"]</code>
  gives the total time that the stream has been limited.
</p>
```

vr000m:
@henbos should rebase, otherwise it looks good.

alvestrand:
henbos will rebase and merge

PR #272, opened 2017-11-06
Split RTCMediaStreamTrackStats into four dictionaries.

jan-ivar:
Fix for https://github.com/w3c/webrtc-stats/issues/230.

Issue #231, opened 2017-07-11
We need "sender" and "receiver" stats, not "track" stats


jan-ivar:
[RTCMediaStreamTrackStats](https://w3c.github.io/webrtc-stats/#dom-rtcmediastreamtrackstats) sets up a nice symmetry between local fields:
```js
    unsigned long       framesCaptured;
    unsigned long       framesSent;
```
and remote fields:
```js
    unsigned long       framesReceived;
    unsigned long       framesDecoded;
    unsigned long       framesDropped;
    unsigned long       framesCorrupted;
    unsigned long       partialFramesLost;
    unsigned long       fullFramesLost;
```
Would you expect to be able to compare `framesSent` and `framesReceived`?

Unfortunately, this was designed before transceivers, senders and replaceTrack, so there's an impedance mismatch: In the new model, remote tracks are tied to *senders*, not local tracks.

E.g. take the ["Hold" example](http://w3c.github.io/webrtc-pc/#hold-functionality) in the spec:
 1. I put you on hold using `audio.sender.replaceTrack(null)`,
 2. and then I bring you back from hold with `audio.sender.replaceTrack(micTrack)`

Would you still expect to be able to compare `framesSent` and `framesReceived`?

We can't because the *local* stats have now been reset from the time of the second `replaceTrack`, whereas the remote stats are *never reset*. The counters are out of sync.

There's also stats bloat (on the local side only), in that you'll see one additional `detached` RTCMediaStreamTrackStats record for each time `replaceTrack(micTrack)` was called. In theory, the sum of all the `framesSent` with the same track `id` should be comparable to `framesReceived`, but that's quite a bit of work, asymmetrical and unintuitive.

Instead, we think people care more about transmission data in relation to the *sender*, not the track.

### Proposal

After splitting up the dictionary in https://github.com/w3c/webrtc-stats/issues/230, replace the `"track"` [RTCStatsType](https://w3c.github.io/webrtc-stats/#dom-rtcstatstype) with `"sender"` and `"receiver"` and have those refer to the local and remote half of this dictionary respectively. e.g.

s/LocalVideoStreamTrackStats/VideoSenderStats/
s/RemoteVideoStreamTrackStats/VideoReceiverStats/
s/LocalAudioStreamTrackStats/AudioSenderStats/
s/RemoteAudioStreamTrackStats/AudioReceiverStats/

vr000m:
[...]

PR #273, opened 2017-11-06
Add "sender" and "receiver" stats.

jan-ivar:
Fix for https://github.com/w3c/webrtc-stats/issues/231 (includes/on-top-of patch from https://github.com/w3c/webrtc-stats/pull/272).

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -526,7 +526,11 @@ enum RTCStatsType {
           <dd>
             <p>
               Contains statistics related to a specific MediaStreamTrack and the corresponding
-              media-level metrics. It is accessed by the
+              media-level metrics. It is accessed by one of
+              <code><a>RTCLocalVideoStreamTrackStats</a></code>,
+              <code><a>RTCLocalAudioStreamTrackStats</a></code>,
+              <code><a>RTCRemoteVideoStreamTrackStats</a></code>, or
+              <code><a>RTCRemoteAudioStreamTrackStats</a></code>, all inherited from
               <code><a>RTCMediaStreamTrackStats</a></code>.
             </p>
           </dd>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1941,11 +1941,10 @@ enum RTCStatsType {
         </p>
         <p>
           If the track is detached from the PeerConnection (via removeTrack or via replaceTrack),
-          it continues to appear, but with the "detached" member set to <code>true</code>.
+          it continues to appear, but with the "objectDeleted" member set to <code>true</code>.
         </p>
         <div>
           <pre class="idl">dictionary RTCLocalVideoStreamTrackStats : RTCVideoSenderStats {
-             boolean             detached;
 };</pre>
           <section>
             <h2>
@@ -1953,16 +1952,6 @@ enum RTCStatsType {
             </h2>
             <dl data-link-for="RTCLocalVideoStreamTrackStats" data-dfn-for="RTCLocalVideoStreamTrackStats"
             class="dictionary-members">
-              <dt>
-                <dfn><code>detached</code></dfn> of type <span class=
-                "idlMemberType"><a>boolean</a></span>
-              </dt>
-              <dd>
-                <p>
-                  True if the track has been detached from the PeerConnection object. If true, all
-                  stats reflect their values at the time when the track was detached.
-                </p>
-              </dd>
             </dl>
           </section>
         </div>
@@ -2303,11 +2292,10 @@ enum RTCStatsType {
         </p>
         <p>
           If the track is detached from the PeerConnection (via removeTrack or via replaceTrack),
-          it continues to appear, but with the "detached" member set to <code>true</code>.
+          it continues to appear, but with the "objectDeleted" member set to <code>true</code>.
         </p>
         <div>
           <pre class="idl">dictionary RTCLocalAudioStreamTrackStats : RTCVideoSenderStats {
-             boolean             detached;
 };</pre>
           <section>
             <h2>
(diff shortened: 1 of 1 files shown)


vr000m:
As per the discussion at TPAC17, I am opposed to removing the track stats. The decision was to keep the both the track and the sender and receiver stats for the forseeable future. If implementation experience suggests that the tracks can be removed, we can revisit the discussion at the time.

jan-ivar:
@vr000m I'll update this PR to keep the track stats around. Just haven't had the time.

alvestrand:
@jan-ivar ping?
This is probably the PR that is most likely to invalidate other PRs, so it would be nice to land it.

webrtc-stats.html
```diff
@@ -472,15 +473,27 @@ <h3>
             </p>
           </dd>
           <dt>
-            <dfn><code>track</code></dfn>
+            <dfn><code>sender</code></dfn>
           </dt>
           <dd>
             <p>
-              Contains statistics related to a specific MediaStreamTrack and the corresponding
+              Contains statistics related to a specific sender and the corresponding
```
henbos: Here and elsewhere: prefer RTCRtpSender/RTCRtpReceiver. I realize this is a mix of moved text and new text though so we can do it separately if need be.

webrtc-stats.html
```diff
@@ -472,15 +473,27 @@ <h3>
             </p>
           </dd>
           <dt>
-            <dfn><code>track</code></dfn>
+            <dfn><code>sender</code></dfn>
```
henbos: Should we have sender, receiver AND track? The sender/receiver referencing trackId.

In the sender case, you might expect counters to go up for the sender and for the current track, but as soon as you replaceTrack a different counter is used.

How useful are the track stats after we have sender/receiver stats? Which should be kept? @vr000m

vr000m:
This should be to keep track, sender, and receiver. @jan-ivar could you update this PR?

jan-ivar:
Working on it.

alvestrand:
@jan-ivar happy new year! Still working on it?

jan-ivar:
Incorporated feedback from @henbos in https://github.com/w3c/webrtc-stats/pull/272#discussion_r156664045.

Note: It's sufficient to merge this PR only rather than https://github.com/w3c/webrtc-stats/pull/272 (if https://github.com/w3c/webrtc-stats/pull/272 is merged I'll need to rebase again).

alvestrand:
I noticed that since we've now added "objectDeleted", the only reason for having a class for track stats has disappeared - "detached" is now served by "objectDeleted" (at least until we decide to drop them both).

webrtc-stats.html
```diff
@@ -1859,23 +1841,56 @@ <h2>
...
                   is the current decoding framerate.
                 </p>
               </dd>
+            </dl>
+          </section>
+        </div>
+      </section>
+      <section id="vsstats-dict*">
+        <h3>
+          <dfn>RTCVideoSenderStats</dfn> dictionary
```
henbos: We have these hierarchies (and similar for audio and receive):

`RTCMediaStreamTrackStats <- RTCVideoStreamTrackStats <- RTCVideoSenderStats <- RTCLocalVideoStreamTrackStats`

In other words

`Track <- [Audio/Video]Track <- [Audio/Video][Sender/Receiver] <- Local[Audio/Video]Track`

It gets a bit confusing about going back and forth between track language and sender/receiver language.

The way I see it, it's sender/receiver all the way down. The `RTCLocal[Audio/Video]StreamTrackStats` are the only special case where now we're not talking about a _sender_, now it changes to an _attachment_ of a track (it's not even a track-stat, since attaching the same track multiple times yields different attachment stats). Which really is a subview of the sender stats for the duration a track is attached.

Lacking a better word that can be used to refer to both a sender and a receiver than "SenderReceiver" (and avoiding confusion with `RTCRtpTransceiver`), I would name them:
```
RTCSenderReceiverStats (RTCMediaStreamTrackStats)
RTC[Audio/Video]SenderReceiverStats (RTC[Audio/Video]StreamTrackStats)
RTC[Audio/Video][Sender/Receiver]Stats (same name as before)
RTCSender[Audio/Video]TrackAttachmentStats (RTCLocal[Audio/Video]StreamTrackStats)
```

In other words

`SenderReceiver <- [Audio/Video]SenderReceiver <- [Audio/Video][Sender/Receiver] <- [Audio/Video]TrackAttachment`

But renames can be done in separate PRs.

alvestrand:
I think we should merge this one, and iterate over a) naming and b) whether we need the track class at all.

jan-ivar:
I've just removed the `detached` member for now. We can refactor away the dictionary entirely later perhaps.

webrtc-stats.html
```diff
@@ -1859,23 +1841,56 @@ <h2>
...
               </dt>
               <dd>
                 <p>
-                  Only valid for local video. It represents the total number of frames captured for
-                  this MediaStreamTrack, before encoding. For example, if this track represents a
-                  camera this is the number of frames produced by the camera for this track, whose
-                  framerate could vary due to hardware limitations or environmental factors such as
-                  lighting conditions.
+                  Represents the total number of frames captured, before encoding,
+                  for this RTCRtpSender (or for this MediaStreamTrack, if
```
henbos: nit: Here and elsewhere: "...for this MediaStreamTrack _attachment_..."

henbos:
Thanks so much for doing this @jan-ivar, this was a large one!!

webrtc-stats.html
```diff
@@ -520,16 +522,40 @@ <h3>
...
-          <dt>
+           <dt>
             <dfn><code>track</code></dfn>
+           </dt>
+           <dd>
+             <p>
+              Contains statistics related to a specific MediaStreamTrack and the corresponding
+              media-level metrics. It is accessed by either
+              <code><a>RTCLocalVideoStreamTrackStats</a></code> or
+              <code><a>RTCLocalAudioStreamTrackStats</a></code>, both inherited from
```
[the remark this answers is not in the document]
jan-ivar: This doesn't remove "track" stats, so no breaking change (yet). The receiving end would presumably see both "track" and "receiver" with redundant info.

webrtc-stats.html
jan-ivar: Right, though that was probably an existing issue? This PR mostly moves words around.

alvestrand:
Editors meeting: Let's merge and fix naming etc. afterwards.

PR #276, opened 2017-12-05
Adding dtlsCipher and srtpCipher

### LINKED ISSUE #248: Add stats for the negotiated DTLS-SRTP and DTLS cipher suites.
Presumably this would go on `RTCTransportStats`. Chrome's non-standard stats implementation included these, with the names "srtpCipher" and "dtlsCipher". They would be `DOMString`s that use cipher suite names defined in RFCs, such as "TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA".

Any reason not to do this? It seems like it would actually help in implementing one of the requirements of rtcweb-security-arch:

>    UI Requirements:   A user-oriented client MUST provide an "inspector"
>       interface which allows the user to determine the security
>       characteristics of the media.
>    ...
>       *  The "security characteristics" MUST indicate the cryptographic
>          algorithms in use (For example: "AES-CBC" or "Null Cipher".)
>          However, if Null ciphers are used, that MUST be presented to
>          the user at the top-level UI.

alvestrand:
Fixes #248

Note: https://github.com/tobie/specref/pull/428 adds
[[IANA-DTLS-SRTP]] and [[IANA-TLS-CIPHERS]] to specref; PR should be
updated with these if it lands before this one is ready.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2141,6 +2141,8 @@ enum RTCStatsType {
              DOMString             selectedCandidatePairId;
              DOMString             localCertificateId;
              DOMString             remoteCertificateId;
+             DOMString             dtlsCipher;
+             DOMString             srtpCipher;
 };</pre>
           <section>
             <h2>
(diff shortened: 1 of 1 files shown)


alvestrand:
Tagging @ekr - would feel comfortable if he says it looks correct.

alvestrand:
Thanks @ekr!
Does the group name come from https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8 ? I suppose that's only applicable for EC cipersuites?
What's the correct reference for the TLS version field? RFC 5246 seems to define only the { 3, 3 } aka TLS 1.2 value - ideally, I'd like to point to somewhere that gives "TLS 1.2" as a string, but as a fallback, I could give a string representation of the version flag from the clientHello.
I want to make sure we don't get "tls-1.2" in one implementation and "TLS 1.2" in another.

alvestrand:
Harald will create an issue for the missing stuff.

vr000m:
Make an issue for @ekr's comments... and I will merge this one.

PR #277, opened 2017-12-05
Add packetsDuplicated

### LINKED ISSUE #254: add packetsDuplicated as a result of #253
to make sure that the negative losses are due to duplicate packets. 

Should duplicate packet counter increase when:

1. FEC is applied?
2. Retransmission?
3. it is an actual duplicate packet. because a router transmitted a packet twice.

alvestrand:
Fixes #254

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -937,8 +937,12 @@ enum RTCStatsType {
                 The cumulative number of packets discarded because
                 they are duplicated. Duplicate packets are not counted
                 in <code><a>packetsDiscarded</a></code>.<br/>
+                Duplicated packets have the same RTP sequence number and
+                content as a previously received packet. If multiple duplicates
+                of a packet are received, all of them are counted.<br/>
                 An improved estimate of lost packets can be calculated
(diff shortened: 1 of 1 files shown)


vr000m:
RFC3550 determines that there might be duplicates because the loss may be negative 
and RFC3611 do not really define a duplicate packet as receiving the same sequence number more than once in a reporting period.

May be should add the following to clarify what we mean by duplicates, i.e., exact duplicates because sequence numbers will wrap around for some sessions: 
> The counter increases each time a duplicate packet with the same sequence number and payload is received (even if the same packet is received multiple times).

alvestrand:
Clarify that multiple copies of the same packet make the counter go up multiple times.
Clarify that duplicates have same sequence number & content

vr000m:
I believe the current proposal clarifies that the packetsDiscarded do not count for duplicates.

alvestrand:
(if the payload has changed, SRTP will object, and we'll get packetFailedDecryption instead)

vr000m:
Then the clarification just needs to be for, counter increases even if multiple copies of the same packet is received.

PR #280, opened 2017-12-08
Use underlines instead of record<a,b> for qualityLimitationDurations

henbos:
#270 broke travis build. Despite using `&lt;` and `&gt;` it got confused. Attempted temporary fix to replace them with parenthesis but that also failed the travis build. Using underscores.

Broken build: https://travis-ci.org/w3c/webrtc-stats/builds/313432008

Temporary fix:
record_RTCQualityLimitationReason__double_
Should be:
record<RTCQualityLimitationReason, double>

I'll file a bug. Maybe ReSpec needs to be fixed.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1237,7 +1237,7 @@ enum RTCStatsType {
              double               totalEncodeTime;
              double               averageRTCPInterval;
              RTCQualityLimitationReason                 qualityLimitationReason;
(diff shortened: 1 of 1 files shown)


henbos:
Oh no, according to WebIDL record's key must be a string and we use an enum:
https://heycam.github.io/webidl/#idl-record

Issue #241, opened 2017-09-12
Is bytesReceived really available for RTCRemoteInboundRTPStreamStats?


na-g:
I am not sure how bytesReceived can be calculated on the sender side from RTCP receiver reports.  Am I missing something or should this stat be moved from RTCReceivedRTPStreamStats to RTCInboundRTPStreamStats?

jan-ivar:
"calculated as defined in" https://tools.ietf.org/html/rfc3550#section-6.4.1 which is Sender Report.

jan-ivar:
From Receiver Report https://tools.ietf.org/html/rfc3550#section-6.4.2 *"the five words of sender information are omitted (these are the NTP and RTP timestamps and sender's packet and octet counts)"*

vr000m:
make a PR with 6.4.2 to point to the correct section.

alvestrand:
Assigning to @vr000m since he promised to check it.

alvestrand:
RFC 3550 section 6.4.2 seems to say that the sender octets are omitted from the RR.

PR #284, opened 2017-12-13
moved bytes and packets received counters

vr000m:
from RTCReceivedRTPStreamStats to RTCInboundRTPStreamStats as per #241


webrtc-stats.html
```diff
@@ -1136,6 +1116,26 @@ <h2>
                   statistics were generated by the local endpoint.
                 </p>
               </dd>
+              <dt>
+                <dfn><code>packetsReceived</code></dfn> of type <span class=
```
alvestrand: We can calculate this from the RR using the calculation in 3550 appendix A (reverse the Appendix A calculation). Should we?

vr000m: In A.3, there is a point where the code does: `s->received++`, so all packets that are received irrespective of duplicates or lost are counted up.

alvestrand: A.3 says 


      expected_interval = expected - s->expected_prior;
      s->expected_prior = expected;
      received_interval = s->received - s->received_prior;
      s->received_prior = s->received;
      lost_interval = expected_interval - received_interval;

then sum(lost_interval) is transmitted in the RR as "cumulative number of packets lost".

The LSR difference betwen the start of session and current report gives "cumulative expected".
Expected - Lost = received.
Am I wrong?

vr000m: at the sender, there is only highest sequence number and loss reported.  I do not see an expected packets in the RTCP RR.

For example,
Sender sends: 1, 2, 3, 4, 5.
Receiver receives: 1, 2, 2, 5. Hence, 4 packets received.
Receiver reports: 5 as HSN and 2 as lost.
the sender could calculate HSN-lost = 5 - 2, which would be 3. 

Did I miscalculate?

alvestrand: Receiver will report loss = 1 based on 3550 calculation where duplicates are conted double.
So sender could calculate 4.

vr000m: correct, 5-4 = 1 lost.

vr000m:
Split this in to two PRs

webrtc-stats.html
icydragons: Drive by, sorry it's a bit off topic of this thread. Reporting loss rate of 1 in that case seems to be really not valuable. It prevents loss rate from being used as a proxy for the packets received or network quality (since 1, 2, 2, 2, 2, 5 has a loss rate of 0). It is also misleading since no one when seeing "loss rate" would expect it to count the total packets received as opposed to the packets that were lost. Is there a way to calculate the number desired, which would be a loss rate of 2?

vr000m: @icydragons that is more an RTP question, is the stack able to mark packets as lost vs re-ordered? how long does it wait before it marks them lost...

alvestrand:
Varun to fix the merge conflicts and merge.

vr000m:
Foxed #241

Issue #271, opened 2017-11-02
Add stat for inputAudioLevel, before the audio filter


huibk:
It could be useful to have a stat the represents the audio level before noise filtering or gain control. Use cases are:
- distinguish silence from missing microphone input from silence due to noise suppression
- determine the level of processing of an audio signal

alvestrand:
Doodling: This would logically be a stat on the source for a track, not a stat on the track itself, no?

if you have:

getUserMedia(audio, id=foo, volume=1.0) => track1
getUserMedia(audio, id=foo, volume=0.5) => track2

and get an input signal at 0.5 (-6dBov)

then track2 would get 0.25 (-12dBov) and track1 would get 0.5 as "level"; an input stat would get 0.5 for both.

One way of getting "unprocessed volume" would be

getUserMedia(audio, id=foo) => track1
getUserMedia(audio, id=foo, processing=none) => track2

track2 should then get the number you want.
Don't know if that works now.

huibk:
Good point, getting a second track without processing may achieve the same. Presumably at higher cost though.

vr000m:
from #288: 
My expectation was that input audio level and output level would match. And if someone did not hear anything then the volume stat being 0 would identify the problem.

I am trying to see now if we compare the input audio level with the audio output level and the audio output takes the volume into account, how do I diagnose if the issue is with the post decoding filter...?

alvestrand:
An implementation experiment showed that on current Chrome, you can't turn on echo cancellation on one track and turn it off on another track from the same source - which limits the usefulness of my other idea. https://crbug.com/802198
Firefox at least reports that echo has been turned off, but no discernible volume difference in simple tests. https://github.com/webrtc/samples/pull/993 is the test page.

na-g:
Would RTCRtp{Contributing,Synchronization}Source (from webrtc-pc) be an appropriate place for this? It seems like this is data that one might want to poll frequently.

na-g:
It took me too long to realize this was about local media. Still, there may be a better API possible (one that is synchronous and fast) than getStats for audio data.

alvestrand:
Thought: One possibility is to define a "source" stats object, which would have a reference from the "track" stats object. This is a place to hang both input audio level and input frame width and height (which has been requested in other contexts - see googFrameWidthInput non-standard stats).
@henbos @burnburn what do you think?

alvestrand:
[...]

PR #288, opened 2017-12-20
Additional description of audioLevel

### LINKED ISSUE #239: Do the "audio level" stats include MediaStreamTrack volume settings?
For example: If an application sets [volume](https://www.w3.org/TR/mediacapture-streams/#def-constraint-volume) to 0.5, should the `audioLevel` stat be multiplied by 0.5 as well? Or does it represent the level from the raw stream of RTP packets? I assumed it was the latter, but it would be useful to clarify.

alvestrand:
Fixes #239 (by answering "yes").

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1996,6 +1996,23 @@ enum RTCStatsType {
                   level = -round(log10(audioLevel) * 20), with audioLevel 0.0 and values above 127
                   mapped to 127.
                 </p>
+                <p>
+                  The <a>audioLevel</a> represents the output audio
+                  level of the track; thus, if the track is sourced
+                  from an <a>RTCReceiver</a>, does no audio
+                  processing, has a constant level, and has
+                  a <a>volume</a> setting of 1.0, the audio level is
+                  expected to be the same as the audio level of the
+                  source SSRC, while if the <a>volume</a> setting is
+                  0.5, the audioLevel is expected to be half that
+                  value.
+                </p>
+                <p>
+                  The audioLevel is averaged over some small
+                  interval, using the algortihm described
+                  under <a>totalAudioEnergy</a>. The interval used is
+                  implementation dependent.
+                </p>
               </dd>
               <dt>
                 <dfn><code>totalAudioEnergy</code></dfn> of type <span class=
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2008,6 +2008,10 @@ enum RTCStatsType {
                   value.
                 </p>
                 <p>
+                  For outgoing audio tracks, the audioLevel is the
+                  level of the audio being sent.
+                </p>
+                <p>
                   The audioLevel is averaged over some small
                   interval, using the algortihm described
                   under <a>totalAudioEnergy</a>. The interval used is


webrtc-stats.html
```diff
@@ -1996,6 +1996,23 @@ <h2>
...
                 </p>
+                <p>
+                  The <a>audioLevel</a> represents the output audio
+                  level of the track; thus, if the track is sourced
+                  from an <a>RTCReceiver</a>, does no audio
+                  processing, has a constant level, and has
+                  a <a>volume</a> setting of 1.0, the audio level is
+                  expected to be the same as the audio level of the
+                  source SSRC, while if the <a>volume</a> setting is
+                  0.5, the audioLevel is expected to be half that
```
taylor-b: Doesn't this go against @vr000m's recommendation?

webrtc-stats.html
```diff
@@ -1996,6 +1996,23 @@ <h2>
...
+                  value.
+                </p>
+                <p>
+                  The audioLevel is averaged over some small
+                  interval, using the algortihm described
```
taylor-b: This "small interval" part is new; what was the assumption before? That it was just the last decoded/sent packet?

alvestrand: The previous assumption was that it wasn't mentioned; I think the code averaged over a packet. This leaves it specified as "implementation defined", but the averaging has to use the sum-of-squares method rather than just averaging (averaging gives a completely bogus number).

webrtc-stats.html
alvestrand: Not sure which recommendation you're referring to.

webrtc-stats.html
```diff
@@ -1996,6 +1996,23 @@ <h2>
                   level = -round(log10(audioLevel) * 20), with audioLevel 0.0 and values above 127
                   mapped to 127.
                 </p>
+                <p>
+                  The <a>audioLevel</a> represents the output audio
+                  level of the track; thus, if the track is sourced
+                  from an <a>RTCReceiver</a>, does no audio
```
henbos: nit: RTCRtpReceiver

vr000m: did this metric not report on the input and an output audio level?

vr000m: Or at least there should be a section on input audio level as well.

webrtc-stats.html
vr000m: My expectation was that input audio level and output level would match. And if someone did not hear anything then the volume stat being 0 would identify the problem.

I am trying to see now if we compare the input audio level with the audio output level and the audio output takes the volume into account, how do I diagnose if the issue is with the post decoding filter...?

alvestrand:
Need to add text saying that for outgoing tracks, it's the audio level transmitted into the PeerConnection.

webrtc-stats.html
alvestrand: You ask for a solution to #271

webrtc-stats.html
alvestrand: #271

alvestrand:
That volume is a range is an accident of the constraint syntax. It should mean, formally, that the UA is going to pick a number within that range; it would be rather silly of the UA to not pick the middle.
There's no allowance in the mediacapture-main spec to say that volume only applies to certain types of track; I don't see any powerful reason to treat remote audio tracks differently.

PR #297, opened 2018-01-17
Remove reference to RTCRtpSynchronizationSource.audioLevel.

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1734 (along with https://github.com/w3c/webrtc-pc/pull/1736).


alvestrand:
Since this is just a pointer removal, it seems that we can merge this without waiting for webrtc-pc.

Issue #278, opened 2017-12-07
What happens when a partial keyFrames is received? 


vr000m:
does the keyFramesReceived counter go up?
partialFramesLost counter will definitely go up...

alvestrand:
We have to decide whether partial lost frames are counted as received.
My vote would be for "no" - it seems silly to count fully lost frames as "received", and counting a multipacket frame as "received" when we have received one useless packet seems wrong too.

So I'd think that we don't increment framesReceived, and don't increment keyFramesReceived.

(Q: When a frame is first partially received and then repaired using FEC or retransmission - does the partialFramesReceived counter go *down* when repair is finished?)

vr000m:
In https://github.com/w3c/webrtc-stats/pull/268, I made a comment on when these counters go up.

framesReceived are definitely full frames, so they should be counted after repair and reorder has been taken in to account.

so if a keyFrame is partially received, the partialFramesLost goes up.

PR #307, opened 2018-01-24
Remove references to MSE. Fixes #161

### LINKED ISSUE #161: Definitions from MSE need re-targeting
The following stats:

- framesDecoded
- framesDropped
- framesCorrupted

are defined with reference to section 5 of [MEDIA-SOURCE], which is said to be https://www.w3.org/TR/media-source/

However, section 5 of that document doesn't have any such counters.

There was an earlier set of metrics in the 20160705 version, as part fo the VideoPlaybackQuality object, but this has disappeared in later versions:

https://www.w3.org/TR/2016/CR-media-source-20160705/#videoplaybackquality

These definitions need retargeting.

vr000m:
Fixes #161, #278, #308

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -799,7 +799,7 @@ enum RTCStatsType {
                 <p>
                   The definition of QP value depends on the <a>codec</a>; for VP8, the QP value is
                   the value carried in the frame header as the syntax element "y_ac_qi", and
-                  defined in [[RFC6386]] section 19.2. Its range is 0..127. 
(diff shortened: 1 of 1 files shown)


vr000m:
In the process of re-targeting I noticed that there is little distinction between:
partialFramesLost and framesCorrupted...

alvestrand:
Suggestion: Delete framesCorrupted - we don't have a meaning for it.

vr000m:
@alvestrand PTAL

Issue #305, opened 2018-01-23
averageRTCPInterval should be averageRtcpInterval


alvestrand:
Follow general guidelines for field naming; we've chosen to use CamelCase for everything except RTC.

PR #311, opened 2018-01-25
fix camel case averageRtcpInterval and also put in RTCInboundRTPStreamStats

vr000m:
Fixes issue #305. However #272 introduced a bug that the rtcp interval is only available at the sending side and not the receiver side. this fixes that issue.


webrtc-stats.html
```diff
@@ -1381,7 +1394,7 @@ <h3>
              double               targetBitrate;
              unsigned long        framesEncoded;
              double               totalEncodeTime;
-             double               averageRTCPInterval;
+             double               averageRtcpInterval;
```
alvestrand: Please also add the old spelling to the "obsolete stats" section. It's been burned.

vr000m: was it implemented as yet by anyone? @jan-ivar @aboba and will mention that it was originally  in RTCRTPStreamStats

vr000m: fixed, PTAL @alvestrand

vr000m:
Need to fix travis comments.

webrtc-stats.html
jan-ivar: We haven't implemented it.

PR #314, opened 2018-01-30
fix for timestamp for remote and local objects

### LINKED ISSUE #313: When are "fractionLost", "packetsLost," "jitter", and other RFC3550-based stats updated?
There are a number of stats that are described as "calculated as defined in section 6.4.1. of [RFC3550]", which is the "SR: Sender Report RTCP Packet" section. So my question is: should these stats represent the values from the last generated SR? Or should `getStats` return "live" values?

I'm bringing this up because I found that [Chrome only returns values from the last SR](https://bugs.chromium.org/p/webrtc/issues/detail?id=8804), and it wasn't 100% clear if this is the correct behavior or not; "calculated as defined in section 6.4.1" leaves it a bit ambiguous.

My opinion: for "fractionLost", this makes sense, since it's defined as "the fraction of RTP data packets from source SSRC_n lost since the previous SR or RR packet was sent". It's a somewhat redundant stat anyway, since it can be computed with `packetsReceived` and `packetsLost`. But for the rest of the stats, I don't see any reason why we shouldn't return live values.

vr000m:
An attempt fix the issue with timestamp live (local) and reported (remote) stats. Fixes #313

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -799,7 +799,7 @@ enum RTCStatsType {
                 <p>
                   The definition of QP value depends on the <a>codec</a>; for VP8, the QP value is
                   the value carried in the frame header as the syntax element "y_ac_qi", and
-                  defined in [[RFC6386]] section 19.2. Its range is 0..127. 
+                  defined in [[RFC6386]] section 19.2. Its range is 0..127.
                   <!-- Need appropriate references for VP9 and H.264 -->
                 </p>
                 <p>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1251,7 +1252,9 @@ <h3>
         </h3>
         <p>
           The <dfn>RTCRemoteInboundRTPStreamStats</dfn> dictionary represents the remote endpoint's
-          measurement metrics for its incoming RTP stream (our outgoing RTP stream).
+          measurement metrics for a particular incoming RTP stream (corresponding to an outgoing
+          RTP stream at the sending endpoint). The timestamp reported in the statistics object is
+          the time at which the corresponding RTCP RR was received.
```
taylor-b: Should this say "RTCP RR or SR"? I'm not an expert on RTP, but from what I've gathered, both SR and RRs contain "reception report blocks".

vr000m: The information in this particular remote object is extracted from the RTCP RR. Although it could be a compound RTCP containing both an SR and an RR.

PR #315, opened 2018-01-31
Define object lifetimes and mention end-of-life emission

### LINKED ISSUE #235: Is keeping stats around a memory problem?

### LINKED ISSUE #300: Missing definition of "deleted". When exactly do stats get deleted?
One problem we ran into looking at `getCompletedStats` - which [using events](https://github.com/w3c/webrtc-stats/issues/235#issuecomment-358316276) might skirt, but nonetheless seems problematic - is the lack of timing guarantees or a definition of completed/deleted.

For instance - even using @alvestrand's [shim](https://github.com/w3c/webrtc-stats/issues/235#issuecomment-358316276) - would the following be guaranteed to work?
```js
let id = sender.track.id;
await sender.replaceTrack(null);
console.log([...pc.getCompletedStats().values()].find(s => s.trackIdentifier == id); // ?
```
Right now there's no guarantee around *when* stats objects end/complete/delete/finalize. Defining it might be hard as well. For instance, @vr000m has mentioned the importance of including every last byte sent wrt a removed track.

Luckily `replaceTrack` is already spec'ed to not resolve until the media switch has been made, but it doesn't talk about that, in order for the stats to be synchronously available in `replaceTrack`'s success callback, we would actually need to gather those stats as part of the replaceTrack algorithm.

This stuff is sort of only implied in the spec so far.

Alternatively, we could offer no such guarantees, and simply say these things happen eventually.

alvestrand:
Fixes #235 together with https://github.com/w3c/webrtc-pc/pull/1753
Fixes #300

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -113,15 +113,16 @@
         The term <dfn>RTP stream</dfn> is defined in [[RFC7656]] section 2.1.10.
       </p>
       <p>
-        The terms <code><dfn data-cite="!WEBRTC#dom-rtcstats">RTCStats</dfn></code>, The terms
+        The terms <code><dfn data-cite="!WEBRTC#dom-rtcstats">RTCStats</dfn></code>,
         <code><dfn data-cite="!WEBRTC#dom-rtcstats-timestamp">RTCStats.timestamp</dfn></code>,
         <code><dfn data-cite="!WEBRTC#dom-rtcstats-type">RTCStats.type</dfn></code>,
         <code><dfn data-cite="!WEBRTC#dom-rtcstats-id">RTCStats.id</dfn></code>,
-        <code><dfn data-cite="!WEBRTC#dom-rtccertificate">RTCCertificate</dfn></code>, are defined
-        <code><dfn data-cite="!WEBRTC#dom-rtccertificate">RTCCertificate</dfn></code>,
-        <code><dfn data-cite="!WEBRTC#event-statslifetimeended">statslifetimeended</dfn></code> are
+        <code><dfn data-cite="!WEBRTC#dom-rtccertificate">RTCCertificate</dfn></code>, and
+        <code><dfn data-cite="!WEBRTC#event-statsended">statslifetimeended</dfn></code> are
         defined in [[!WEBRTC]].
-        <code><dfn data-cite="!HIGHRES-TIME#dom-performance-timeOrigin">performance.timeOrigin</dfn></code>
+      </p>
+      <p>
+        The terms <code><dfn data-cite="!HIGHRES-TIME#dom-performance-timeOrigin">performance.timeOrigin</dfn></code>
         and <code><dfn data-cite="!HIGHRES-TIME#dom-performance-now">performance.now()</dfn></code>
         are defined in [[!HIGHRES-TIME]].
       </p>
@@ -252,6 +253,7 @@
           For stats that are measured against time, such as byte counts, no separate counter is
           needed; one can instead divide by the difference in the timestamps.
         </p>
+        
       </section>
       <section>
         <h3>
@@ -278,22 +280,42 @@
           Lifetime considerations for monitored objects
         </h3>
         <p>
-          The object descriptions will say what the lifetime of an object is. If no lifetime is
-          noted, the object lasts until the PeerConnection object is destroyed.
+          The object descriptions will say what the lifetime of a <a>monitored object</a> from the
+          perspective of stats is. When a monitored object is "deleted", it no longer appears in stats;
+          until this happens, it will appear. This may or may not correspond to the actual lifetime
+          of an object in an implementation; what matters for this specification is what appears in stats.
         </p>
         <p>
-          When a monitored object is destroyed, a final stats object is produced, carrying the
-          values current at the time of destruction. This object will be made available, with a
-          timestamp reflecting the time of destruction, using the
+          If a monitored object can only exist
+          in a few instances over the lifetime of a PeerConnection, it may be simplest to consider it
+          "eternal" and never delete it from the set of objects reported on in stats.
+          This type of object will remain visible until the PeerConnection
+          is no longer available; it is also visible in getStats() after pc.close(). This is the default
+          when no lifetime is mentioned in its specification.
+        </p>
+        <p>
+          Objects that might exist in many instances over time SHOULD have a defined time at which
+          they are deleted, at which time a "statsended" event is fired on their behalf. Each
+          event that causes deletions to happen SHOULD fire only one "statsended" event; for instance,
+          an ICE restart SHOULD fire only one event containing the stats for all the discarded candidates
+          and pairs, but may also cause a later event to be fired when the currently-in-use candidate pair
+          and its candidates are discarded.
+        </p>
+        <p>
+          When a <a>monitored object</a> is deleted, a final stats object is produced, carrying the
+          values current at the time of deletion. This object will be made available, with a
+          timestamp reflecting the time of deletion, using the
           <code><a>statslifetimeended</a></code> event on the associated PeerConnection. This is
           important in order to report consistently on short-lived objects and to be able to
           consistently report totals over the lifetime of a PeerConnection.
         </p>
         <p>
-          When an object is destroyed, we can guarantee that no subsequent getStats() call will
-          reference the destroyed object. We also guarantee that the stats id of the destroyed
+          When an object is deleted, we can guarantee that no subsequent getStats() call will
+          contain a <a>stats object reference</a> that references the deleted object
+          We also guarantee that the stats id of the deleted
           object will never be reused for another object. This ensures that an application that
-          collects deleted objects will always be able to uniquely identify the object pointed to
+          collects <a>stats object</a>s for deleted <a>monitored object</a>s will always be able to
+          uniquely identify the object pointed to
           in the result of any getStats() call.
         </p>
       </section>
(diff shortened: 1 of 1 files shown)


vr000m:
@lennart-csio @karthikbr82 PTAL.

webrtc-stats.html
```diff
@@ -596,6 +586,10 @@ <h3>
               It is accessed by the <code><a>RTCIceCandidateStats</a></code> for the remote
               candidate.
             </p>
+            <p>
+              A remote candidate is deleted when the RTCIceTransport does an ICE restart, and the
+              candidate is no longer in use.
+            </p>
```
lennart-csio: the candidates are deleted when they are "no longer in use", which afaik is the case after ICE completed and chose an active pair? so, if one queries for stats e.g. only at the end of the connection, then it will never know what other candidates had been available. is this intended?

lennart-csio: i just read the definition of "statslifetimeended" event from webrtc-pc, so i guess my question is void. it will simply emit the event when the candidate(pair) is deleted?

webrtc-stats.html
```diff
@@ -532,12 +510,16 @@ <h3>
...
-              media-level metrics. It is accessed by either
+              Contains statistics related to a specific MediaStreamTrack's connection to this
+              PeerConnection and the corresponding media-level metrics. It is accessed by either
               <code><a>RTCLocalVideoStreamTrackStats</a></code> or
               <code><a>RTCLocalAudioStreamTrackStats</a></code>, both inherited from
               <code><a>RTCMediaStreamTrackStats</a></code>.
             </p>
+            <p>
+              The monitored "track" object is deleted when the track is disconnected from the
+              PeerConnection.
```
jan-ivar: When exactly is that? Is it in the in-parallel steps of [replaceTrack](http://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-replacetrack)?

I'd love to nail down when the event fires. Could we say it MUST fire before replaceTrack's promise is resolved?

webrtc-stats.html
```diff
@@ -576,6 +558,10 @@ <h3>
               ICE candidate pair statistics related to the <code>RTCIceTransport</code> objects. It
               is accessed by the <code><a>RTCIceCandidatePairStats</a></code>.
             </p>
+            <p>
+              A candidate pair is deleted when the RTCIceTransport does an ICE restart, and the
+              candidate pair is no longer in use.
```
jan-ivar: When exactly is that? Is it when the [statechange](http://w3c.github.io/webrtc-pc/#event-icetransport-statechange) event is fired? Is that from SLD?

If so, should we say it MUST fire before SLD's promise is resolved?

webrtc-stats.html
jan-ivar: How do we capture that we want a single event from an ICE restart with one RTCStatsReport containing all the deleted candidates?

webrtc-stats.html
jan-ivar: Or is it right after that ICE changes happen?

webrtc-stats.html
```diff
@@ -303,15 +270,26 @@ <h3>
...
+          timestamp reflecting the time of destruction, using the
+          <code><a>statslifetimeended</a></code> event on the associated PeerConnection. This is
+          important in order to report consistently on short-lived objects and to be able to
+          consistently report totals over the lifetime of a PeerConnection.
+        </p>
+        <p>
+          When an object is destroyed, we can guarantee that no subsequent getStats() call will
+          reference the destroyed object. We also guarantee that the stats id of the destroyed
+          object will never be reused for another object. This ensures that an application that
+          collects deleted objects will always be able to uniquely identify the object pointed to
```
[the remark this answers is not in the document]
alvestrand: I switched to using "deleted" and made it mean "deleted from stats". That seems consistent.

Also qualified a lot of the objects.

webrtc-stats.html
alvestrand: Changed to "when the sender has its track value changed". Probably best to incorporate this in webrtc-pc algorithm too. Also updated the definition - the "this is an attachment" language hadn't made its way down here.

webrtc-stats.html
alvestrand: I think it's when the ICE transport changes state to "new".

webrtc-stats.html
alvestrand: I think I captured that in guidelines.

alvestrand:
Comments adressed, I think. PTAL.

webrtc-stats.html
```diff
@@ -307,15 +277,46 @@ <h3>
...
+          when no lifetime is mentioned in its specification.
+        </p>
+        <p>
+          Objects that might exist in many instances over time SHOULD have a defined time at which
+          they are deleted, at which time a "statsended" event is fired on their behalf. Each
+          event that causes deletions to happen SHOULD fire only one "statsended" event; for instance,
+          an ICE restart SHOULD fire only one event containing the stats for all the discarded candidates
+          and pairs, but may also cause a later event to be fired when the currently-in-use candidate pair
+          and its candidates are discarded.
+        </p>
```
vr000m: trying to parse why this is SHOULD and not MUST.  Because SHOULD needs to also say when it wont happen? otherwise it would be a MUST.

webrtc-stats.html
```diff
@@ -307,15 +277,46 @@ <h3>
...
+          in a few instances over the lifetime of a PeerConnection, it may be simplest to consider it
+          "eternal" and never delete it from the set of objects reported on in stats.
+          This type of object will remain visible until the PeerConnection
+          is no longer available; it is also visible in getStats() after pc.close(). This is the default
+          when no lifetime is mentioned in its specification.
+        </p>
+        <p>
+          Objects that might exist in many instances over time SHOULD have a defined time at which
+          they are deleted, at which time a "statsended" event is fired on their behalf. Each
+          event that causes deletions to happen SHOULD fire only one "statsended" event; for instance,
```
vr000m: link to "statsended" to the appropriate section in webrtc-pc

webrtc-stats.html
```diff
@@ -307,15 +277,46 @@ <h3>
...
+          an ICE restart SHOULD fire only one event containing the stats for all the discarded candidates
+          and pairs, but may also cause a later event to be fired when the currently-in-use candidate pair
+          and its candidates are discarded.
+        </p>
+        <p>
+          When a <a>monitored object</a> is deleted, a final stats object is produced, carrying the
+          values current at the time of deletion. This object will be made available, with a
+          timestamp reflecting the time of deletion, using the
+          <code><a>statslifetimeended</a></code> event on the associated PeerConnection. This is
```
vr000m: statslifetimeended --> statsended

webrtc-stats.html
alvestrand: "There may be reasons this is impractical"  - I think I*m removing the second SHOULD and making the first one ("fire all") a MUST. The example described is that one action (ICE restart) causes multiple deletion events - but each of these MUST only fire one event.

webrtc-stats.html
alvestrand: Done.

webrtc-stats.html
alvestrand: Done.

webrtc-stats.html
```diff
@@ -307,15 +277,47 @@ <h3>
...
-          objects and to be able to consistently report totals over the lifetime of a
-          PeerConnection.
+          The object descriptions will say what the lifetime of a <a>monitored object</a> from the
+          perspective of stats is. When a monitored object is "deleted", it no longer appears in stats;
+          until this happens, it will appear. This may or may not correspond to the actual lifetime
+          of an object in an implementation; what matters for this specification is what appears in stats.
+        </p>
+        <p>
+          If a monitored object can only exist
+          in a few instances over the lifetime of a PeerConnection, it may be simplest to consider it
```
henbos: nit: Here and elsewhere - RTCPeerConnection

webrtc-stats.html
```diff
@@ -307,15 +277,47 @@ <h3>
...
+          event that causes deletions to happen MUST fire only one <a>statsended</a> event, but there are
+          cases where one action causes multiple deletion events; for instance,
+          an ICE restart will fire only one event containing the stats for all the discarded candidates
+          and pairs, but will also cause a later event to be fired when the currently-in-use candidate
+          pair and its candidates are discarded.
+        </p>
+        <p>
+          When a <a>monitored object</a> is deleted, a final stats object is produced, carrying the
+          values current at the time of deletion. This object will be made available, with a
+          timestamp reflecting the time of deletion, using the
```
henbos: Will it always be the case that the timestamp of the stats object reflect the time of deletion?

For example, RTCInboundRtpStreamStats has "The timestamp reported in the statistics object is the time at which the data was sampled." and RTCRemoteInboundRtpStreamStats has "The timestamp reported in the statistics object is the time at which the corresponding RTCP RR was received."
The spec contradicts itself, either we lose the ability to know when the corresponding RR was received or we lose the ability to know when the monitored object was deleted.

Should we add a timestamp to RTCStatsReport or to the statsended event?
I'm fine with discussing this separately if we want to land this.

alvestrand: It's not clear that it makes sense ever to delete remote objects; we don't know if they're gone, or if another RR is imminent some time in the near future. I'd consider this more closely before making changes.

The InboundRtpStreamStats timestamp is consistent with the time of deletion; we sample at the time of deletion.

webrtc-stats.html
alvestrand: The document uses "PeerConnection" all over the place, with and without the "RTC" prefix. 
Filed https://github.com/w3c/webrtc-stats/issues/336.

webrtc-stats.html
alvestrand: Deleted the subsentence. The start says "carrying the values current at the time of deletion"; the timestamp should reflect the time these stats were gathered (whatever that is).

Issue #317, opened 2018-01-31
Typo - RTCLocalAudioStreamTrackStats should inherit from RTCAudioSenderStats


karthikbr82:
In section 7.21 RTCLocalAudioStreamTrackStats should inherit from RTCAudioSenderStats not RTCVideoSenderStats.

The dictionary is empty, any plans to add some stats to it ?

alvestrand:
We can delete those unless we find an use for them Real Soon.

vr000m:
@jan-ivar delete this?

henbos:
I fly-by fixed this in #317, let's see if that gets merged.

henbos:
I mean in #322

PR #322, opened 2018-02-07
Rename sender/receiver/track dictionaries

### LINKED ISSUE #298: Rename sender/receiver/track stats
Largely copied from https://github.com/w3c/webrtc-stats/pull/273.

We have these hierarchies (and similar for audio and receive):

`RTCMediaStreamTrackStats <- RTCVideoStreamTrackStats <- RTCVideoSenderStats <- RTCLocalVideoStreamTrackStats`

In other words

`Track <- [Audio/Video]Track <- [Audio/Video][Sender/Receiver] <- Local[Audio/Video]Track`

It gets a bit confusing about going back and forth between track language and sender/receiver language.

The way I see it, it's sender/receiver all the way down. The RTCLocal[Audio/Video]StreamTrackStats are the only special case where now we're not talking about a sender, now it changes to an attachment of a track (it's not even a track-stat, since attaching the same track multiple times yields different attachment stats). Which really is a subview of the sender stats for the duration a track is attached.

Lacking a better word that can be used to refer to both a sender and a receiver than "SenderReceiver" (and avoiding confusion with RTCRtpTransceiver), I would name them:

```
RTCSenderReceiverStats (previously RTCMediaStreamTrackStats)
RTC[Audio/Video]SenderReceiverStats (RTC[Audio/Video]StreamTrackStats)
RTC[Audio/Video][Sender/Receiver]Stats (same name as before)
RTCSender[Audio/Video]TrackAttachmentStats (RTCLocal[Audio/Video]StreamTrackStats)
```
In other words

`SenderReceiver <- [Audio/Video]SenderReceiver <- [Audio/Video][Sender/Receiver] <- [Audio/Video]TrackAttachment`

Since "objectDeleted" is now a member of all stats and "detached" has been removed, there is no need for a separate dictionary for track attachments, so we can either reuse the same dictionary or keep it around for editorial purposes to make the distinction between sender/receiver stats and track attachment stats.

henbos:
Fixes #298.

Top-level rename (old => new):
```
RTCMediaStreamTrackStats => RTCMediaHandlerStats
```
Audio names:
```
RTCAudioStreamTrackStats => RTCAudioHandlerStats
RTCAudioSenderStats (not renamed)
RTCAudioReceiverStats (not renamed)
RTCLocalAudioStreamTrackStats => RTCSenderAudioTrackAttachmentStats
```
Video names:
```
RTCVideoStreamTrackStats => RTCVideoHandlerStats
RTCVideoSenderStats (not renamed)
RTCVideoReceiverStats (not renamed)
RTCLocalVideoStreamTrackStats => RTCSenderVideoTrackAttachmentStats
```

This forms the following hierarchy:
```
DICTIONARY NAME                          | INSTANCE FOUND BY
-----------------------------------------+--------------------------------------
RTCMediaHandlerStats                     | N/A
  RTCAudioHandlerStats                   | N/A
    RTCAudioSenderStats                  | kind == 'audio' && type == 'sender' 
      RTCSenderAudioTrackAttachmentStats | kind == 'audio' && type == 'track'
    RTCAudioReceiverStats                | kind == 'audio' && type == 'receiver'
  RTCVideoHandlerStats                   | N/A
    RTCVideoSenderStats                  | kind == 'video' && type == 'sender'
      RTCSenderVideoTrackAttachmentStats | kind == 'video' && type == 'track'
    RTCVideoReceiversStats               | kind == 'video' && type == 'receiver'
```
Where 'track' (attachment) stats have the same members as the senders but relative to the attachment period.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -328,10 +328,10 @@
           promise resolving or an event firing, subsequent new getStats() calls must return
           up-to-date dictionaries for the affected objects. For example, if a track is added with
           addTrack() subsequent getStats() calls must resolve with a corresponding
-          RTCMediaStreamTrackStats object. If you call setRemoteDescription() removing a remote
+          RTCMediaHandlerStats object. If you call setRemoteDescription() removing a remote
           track, upon the promise resolving or an associated event (stream's onremovetrack or
           track's onmute) firing, calling getStats() must resolve with an up-to-date
-          RTCMediaStreamTrackStats object.
+          RTCMediaHandlerStats object.
         </p>
       </section>
     </section><!-- VS: getStats() remains in base spec! -->
@@ -534,9 +534,9 @@ enum RTCStatsType {
             <p>
               Contains statistics related to a specific MediaStreamTrack and the corresponding
               media-level metrics. It is accessed by either
-              <code><a>RTCLocalVideoStreamTrackStats</a></code> or
-              <code><a>RTCLocalAudioStreamTrackStats</a></code>, both inherited from
-              <code><a>RTCMediaStreamTrackStats</a></code>.
+              <code><a>RTCSenderVideoTrackAttachmentStats</a></code> or
+              <code><a>RTCSenderAudioTrackAttachmentStats</a></code>, both inherited from
+              <code><a>RTCMediaHandlerStats</a></code>.
             </p>
           </dd>
           <dt>
@@ -1856,10 +1856,10 @@ enum RTCStatsType {
       </section>
       <section id="mststats-dict*">
         <h3>
-          <dfn>RTCMediaStreamTrackStats</dfn> dictionary
+          <dfn>RTCMediaHandlerStats</dfn> dictionary
         </h3>
         <div>
-          <pre class="idl">dictionary RTCMediaStreamTrackStats : RTCStats {
+          <pre class="idl">dictionary RTCMediaHandlerStats : RTCStats {
              DOMString           trackIdentifier;
              boolean             ended;
              DOMString           kind;
@@ -1867,9 +1867,9 @@ enum RTCStatsType {
 };</pre>
           <section>
             <h2>
-              Dictionary <a class="idlType">RTCMediaStreamTrackStats</a> Members
+              Dictionary <a class="idlType">RTCMediaHandlerStats</a> Members
             </h2>
-            <dl data-link-for="RTCMediaStreamTrackStats" data-dfn-for="RTCMediaStreamTrackStats"
+            <dl data-link-for="RTCMediaHandlerStats" data-dfn-for="RTCMediaHandlerStats"
             class="dictionary-members">
               <dt>
                 <dfn><code>trackIdentifier</code></dfn> of type <span class=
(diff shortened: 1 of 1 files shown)


henbos:
PTAL @jan-ivar and @alvestrand

henbos:
This also fixed #317

jan-ivar:
I kinda like the old names TBH. Why "Handler"? I worry "Sender" might be confusing?

henbos:
We needed a name that could mean "Sender or Receiver" for dictionaries with members in common for both senders and receivers; "Handler" because it's more concise than "SenderOrReceiver" (and we can't use a name like "Transceiver" because that already has special meaning and implies "SenderAndReceiver").

We collect stats for what we have. We have senders, receivers and RTP streams. Tracks are the _biproduct_ of this machinery. Not the other way around.

"Track" stats sounds intuitive until you realize these are just a snapshot of sender stats. The current definitions describe a hierarchy in terms of tracks, e.g:
```
RTCMediaStreamTrackStats
RTCVideoStreamTrackStats
RTCVideoSenderStats
RTCLocalVideoStreamTrackStats
```
This is completely backwards. It does not represent the components that the stats are collected from and make it seem like senders are the biproduct of tracks.

I would argue that "track" stats can be deprecated in favor of "sender" stats and that this legacy view of the components should go away. "Sender" is more future-proof than "track" stats.

That said I agree there is utility in true track stats, since often it is this biproduct you are interested in - e.g. What framerate am I getting on this current setup of the machinery? On this _track_?
But stats belonging to tracks should be added to the track dictionary, which is at the bottom of the hierarchy (if we have a hierarchy that represents how this actually works and not the other way around). The current naming scheme would have you believe that track members should be added high up in the hierarchy - this is wrong and would pollute the sender/receiver stats.

Is there something I need to clarify? @alvestrand what's your take?

henbos:
Track stats make sense but remember that current "track" stats are "track attachment stats", that's why I described them as a biproduct above. Of course a track in the sense of a source or a sink has interesting properties on its own, but that should be placed in dictionaries specific to tracks, not dictionaries in common for senders and receivers.

alvestrand:
It's been a few months (many?) since we added the text that pointed out that "track" stats are about the association between the sender/receiver and the track, not about the track itself.

The "handler" came about because we needed a common word for "sender" and "receiver", and "SenderOrReceiver" was just too clumsy.

The only people who will ever see these names are implementors and spec readers. They don't appear on any API surface. So readability for those people trumps (in my mind) other concerns.

I like the proposed names.

jan-ivar:
"SenderOrReceiver" seems concise to me FWIW. :) But fine.

> "Track" stats sounds intuitive until you realize these are just a snapshot of sender stats.

@henbos Is it a snapshot, or do they contain values specific to one track? I.e. do we reset counters when a new track is added?

I'm supportive of simplifying here, just mindful that https://github.com/w3c/webrtc-stats/pull/315 models the API language around stats objects "ending", which fits if "track" stats are deleted. That language might need massaging if we want to get rid of "track" stats and fire events with "sender" snapshots...

henbos:
One track stats object is created for each track attachment, its counters should be reset relative to when it was attached.
It should be: `trackStats = (senderStats - senderStatsFromWhenTrackWasAttached)`

It might be useful to fire an event when something happens like a track is attached or detached containing 'sender' stats snapshots (despite the sender not being deleted) even if we get rid of 'track' stats. I had not thought of that.

PR #326, opened 2018-02-07
Moved trackId from RTCRTPStreamStats to RTC[In/Out]boundRTPStreamStats.

### LINKED ISSUE #325: Track ID doesn't make sense for remote RTCRtpMediaStream stats
We have no idea what track is attached on the other side, so these should be on the local* versions of these ojbects.

henbos:
Fixes #325.

What this does effectively is remove `trackId` from the `RTCRemote[In/Out]boundRTPStreamStats` which don't have remote versions (but do have corresponding local track attachments referenced by the non-remote stream stats' trackId, which are accessed throug the remote's localId, but this is not the same thing as the remote's view of the track attachment which does not exist).

It would make more sense to put the trackId inside of the sender/receiver stats but it is kept here for backwards compatibility reasons and we are likely to want to remove track stats in the future, no need to simplify accessing them from sender/receiver.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -676,7 +676,6 @@ enum RTCStatsType {
           <pre class="idl">dictionary RTCRTPStreamStats : RTCStats {
              unsigned long       ssrc;
              DOMString           mediaType;
-             DOMString           trackId;
              DOMString           transportId;
              DOMString           codecId;
              unsigned long       firCount;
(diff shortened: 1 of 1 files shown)


vr000m:
resolve conflicts and merge

alvestrand:
Henbos to resolve conflicts and merge.

alvestrand:
@henbos will add a reference here to MediaStreamTrack.kind (just like for Handler) and merge.

alvestrand:
Please rebase and merge.

PR #327, opened 2018-02-07
Rename "mediaType" to "kind" in RTCRTPStreamStats.

### LINKED ISSUE #301: terminology: rename mediaType to kind
We use "kind" (from MediaStreamTrack.kind) everywhere to describe the media type and should update
https://w3c.github.io/webrtc-stats/#dom-rtcrtpstreamstats-mediatype for consistency.

This is also mentioned in the main spec once [here](http://w3c.github.io/webrtc-pc/#mandatory-to-implement-stats)

originally added in #9

henbos:
Fixes #301.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1929,7 +1929,7 @@ enum RTCStatsType {
               <dd>
                 <p>
                   Either "<code>audio</code>" or "<code>video</code>". This reflects the "kind"
-                  attribute of the MediaStreamTrack.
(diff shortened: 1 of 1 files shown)


vr000m:
Perhaps was an issue before this change, Is it clear that currently the kind returns the values of "audio" or "video"?

henbos:
We can reference this to say it should match MediaStreamTrack.kind:
https://www.w3.org/TR/mediacapture-streams/#dom-mediastreamtrack-kind

alvestrand:
Henbos will insert a reference and merge.

PR #331, opened 2018-02-21
Add hugeFramesSent to RTCVideoSenderStats

### LINKED ISSUE #330: Add hugeFramesSent counter to RTCVideoSenderStats dictionary
Please add a hugeFramesSent counter to RTCVideoSenderStats dictionary.

This should report the total number of "huge" frames sent since the beginning of the call.
A frame is defined as "huge" if it's encoded size is at least 2.5 times the average encoded frame size for the configured encoder bitrate. In the simulcast case, if any of the substreams produced a "huge" frame, this should be counted as a single "huge" frame sent.

This metric is useful for in case of screensharing a presentation for detecting slide changes.

Without such a metric, only way to do so would be to compare uncompressed captured frames, which is slow and may not even be possible. On the other hand, this metric is fairly easy to implement inside the webrtc. Although it may not be 100% accurate, for detecting the slide changes, it's reasonably efficient.

ilyanikolaevskiy:
Fixes #330

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2058,11 +2058,18 @@ enum RTCStatsType {
                 <p>
                   Represents the total number of huge frames sent by this RTCRtpSender (or for this
                   MediaStreamTrack, if <code>type</code> is <code>"track"</code>). Huge frames, by
-                  definition, are frames with encoded size at least 2.5 times the target bitrate
-                  per second divided by target fps set for the encoder at the time the frame was
-                  encoded. These are usually difficult to encode frames with a lot of changes in the
-                  picture. This can be used to estimate, e.g slide changes in the streamed
-                  presentation.
+                  definition, are frames with encoded size at least 2.5 times the typical frame size,
+                  where the typical frame size is defined as the target bitrate per second divided by
+                  the target fps set for the encoder at the time the frame was encoded. These are
+                  usually difficult to encode frames with a lot of changes in the picture. This can be
+                  used to estimate, e.g slide changes in the streamed presentation.
+                </p>
+                <p>
+                  The multiplier of 2.5 is choosen from analyzing encoded frame sizes for a sample
+                  presentation using webrtc standalone implementation. 2.5 is a reasonably large
+                  multiplier which still caused all slide change events to be identified as a huge
+                  frames. It, however, produced 1.4% of false positive slide change detections which
+                  is deemed reasonable.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2059,9 +2059,10 @@ enum RTCStatsType {
                   Represents the total number of huge frames sent by this RTCRtpSender (or for this
                   MediaStreamTrack, if <code>type</code> is <code>"track"</code>). Huge frames, by
                   definition, are frames with encoded size at least 2.5 times the target bitrate
-                  per second divided by target fps set for the encoder at the time. These are
-                  usually difficult to encode frames with a lot of changes in the picture. This can
-                  be used to estimate, e.g slide changes in the streamed presentation.
+                  per second divided by target fps set for the encoder at the time the frame was
+                  encoded. These are usually difficult to encode frames with a lot of changes in the
+                  picture. This can be used to estimate, e.g slide changes in the streamed
+                  presentation.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2058,11 +2058,13 @@ enum RTCStatsType {
                 <p>
                   Represents the total number of huge frames sent by this RTCRtpSender (or for this
                   MediaStreamTrack, if <code>type</code> is <code>"track"</code>). Huge frames, by
-                  definition, are frames with encoded size at least 2.5 times the typical frame size,
-                  where the typical frame size is defined as the target bitrate per second divided by
-                  the target fps set for the encoder at the time the frame was encoded. These are
-                  usually difficult to encode frames with a lot of changes in the picture. This can be
-                  used to estimate, e.g slide changes in the streamed presentation.
+                  definition, are frames that have an encoded size at least 2.5 times the average
+                  size of the frames. The average size of the frames is defined as the target bitrate
+                  per second divided by the target fps at the time the frame was encoded. These are
+                  usually complex to encode frames with a lot of changes in the picture. This can be
+                  used to estimate, e.g slide changes in the streamed presentation. If a huge frame is
+                  also a key frame, then both counters hugeFramesSent and keyFramesSent are
+                  incremented.
                 </p>
                 <p>
                   The multiplier of 2.5 is choosen from analyzing encoded frame sizes for a sample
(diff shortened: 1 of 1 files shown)


ilyanikolaevskiy:
Something is really wrong with the integration test. It can't start chrome and need a kernel updated or something?

alvestrand:
IPR issue: Please join the webrtc WG on W3C. That's the simplest way to resolve that issue.

webrtc-stats.html
```diff
@@ -2049,6 +2050,20 @@ <h2>
...
+              <dt>
+                <dfn><code>hugeFramesSent</code></dfn> of type <span class="idlMemberType">
+                <a>unsigned long</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the total number of huge frames sent by this RTCRtpSender (or for this
+                  MediaStreamTrack, if <code>type</code> is <code>"track"</code>). Huge frames, by
+                  definition, are frames with encoded size at least 2.5 times the target bitrate
+                  per second divided by target fps set for the encoder at the time. These are
```
alvestrand: targetBitrate is a property of RTCOutboundRTPStreamStats, while this is a property of the RTCVideoHandlerStats.
framesPerSecond is a property of RTCVideoHandlerStats, so that's easily resolvable, but targetBitrate has to be specified as "the targetBitrate of the RTP stream that is currently sending this track" or something like that. They're connected via the trackId variable of the RTCRTPStreamStats.

ilyanikolaevskiy: I don't see the issue here. The exact frame encoded size is not the part of any stats struct, yet it's needed for calculation of hugeFramesSent. So, tying this exact stat to something not present in RTCVideoHandlerStats is not an issue. All that information is  available inside the WebRTC where it computes the new stat.

It's already stated there, that bitrate and fps are that used by the encoder, which produced the encoded frame for that frame.

To reduce confusion, I will shortly rephrase it to: "... target bitrate per second divided by target fps set for the encoder at the time the frame was encoded"

ilyanikolaevskiy:
Ipr issue is now fixed, but I don't think I can fix continous-integration/travis-cl/pr failing. The linux kernel is outdated on the testing machine, so chrome can't run with a sandbox there.

alvestrand:
@vr000m can you take a quick look?
We can ignore the trybot for the moment - this is a widespread crash. Link: https://github.com/w3c/respec/issues/1522

webrtc-stats.html
```diff
@@ -2049,6 +2050,21 @@ <h2>
...
               </dd>
+              <dt>
+                <dfn><code>hugeFramesSent</code></dfn> of type <span class="idlMemberType">
+                <a>unsigned long</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the total number of huge frames sent by this RTCRtpSender (or for this
+                  MediaStreamTrack, if <code>type</code> is <code>"track"</code>). Huge frames, by
+                  definition, are frames with encoded size at least 2.5 times the target bitrate
```
alvestrand: Suggest writing this in two parts: "encoded size at least 2.5 times the typical frame size, where the typical frame size is defined as the targetBitrate of the corresponding RTCOutboundRTPStreamStats divided by the targetBitrate of this RTCVideoSenderStats".

ilyanikolaevskiy: I agree, this is more readable, but I still don't want to tie this to other reported stats, because there's some aggregation going there. I want to keep "the target bitrate divided by the target fps of the encoder at the time the frame was encoded." Is that fine with you?

vr000m:
There is no target Frame rate, although there is target bitrate. https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-targetbitrate. 

We should add targetFrameRate and may be targetFrameSize.

webrtc-stats.html
```diff
@@ -2049,6 +2050,21 @@ <h2>
...
+                  per second divided by target fps set for the encoder at the time the frame was
+                  encoded. These are usually difficult to encode frames with a lot of changes in the
+                  picture. This can be used to estimate, e.g slide changes in the streamed
```
vr000m: These are usually difficult to encode frames with a lot of changes in the picture --> not sure if I understand this correctly, if there are a lot of changes in the picture, would it not generate a Key Frame?

vr000m: This metric should also state if KeyFrames are excluded or included?

webrtc-stats.html
vr000m: 2.5 is a magic number, we should state somewhere that the number comes from an analysis, so that it would be easier to update the number by either re-analysing the data or by a number from another standard.

ilyanikolaevskiy:
@vr000m, I argue that we don't need to add them. The issue is that the metric is tied to the target bitrate/fps **at the time the frame was encoded**.  That fps/bitrate value may be even different to the one, actual at the time getStats is called. Also, most sensible targetFrameRate added to rtcOutboundRtpStreamStats would be aggregated somehow (average value over the window maybe), so, again, it will in no way be tied to the hugeFramesSent.

webrtc-stats.html
ilyanikolaevskiy: I only considered to add this metric here, because this is what reviewers of webrtc changelist requested.

Usually the encoders don't produce keyframes, even if there are a lot of changes in the frame. 

This is why kayFramesSent can't be used for the purpose of detecting slide changes. Thus, the huge frames concept. It's entirely separate from keyframes. While key frames most probably would be counted in the metric, they are not guaranteed to be so.

webrtc-stats.html
alvestrand: No. When you don't tie it to other stats in an understandable way, two implementations will implement it differently, and give different counters for the same bitstream.
If you add "at the time the frame was encoded", that's fine. It's going to give you the same numbers in two implementations (within the variability of encoding times).

ilyanikolaevskiy: Added a new paragraph justifying the magic number.

webrtc-stats.html
```diff
@@ -2049,6 +2050,28 @@ <h2>
...
               </dd>
+              <dt>
+                <dfn><code>hugeFramesSent</code></dfn> of type <span class="idlMemberType">
+                <a>unsigned long</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the total number of huge frames sent by this RTCRtpSender (or for this
+                  MediaStreamTrack, if <code>type</code> is <code>"track"</code>). Huge frames, by
+                  definition, are frames with encoded size at least 2.5 times the typical frame size,
```
vr000m: are frames that have an encoded size of at least 2.5 times the average size of the frames. The average size of the frames is defined as the target bitrate per second divided by the target frame rate at the time when the frame was encoded.

webrtc-stats.html
```diff
@@ -2049,6 +2050,28 @@ <h2>
...
+                  where the typical frame size is defined as the target bitrate per second divided by
+                  the target fps set for the encoder at the time the frame was encoded. These are
+                  usually difficult to encode frames with a lot of changes in the picture. This can be
```
vr000m: complex instead of difficult.

I do not know what difficult means in this case.

vr000m:
Need to also say something like: if the huge frames are also key frames, then both counters hugeFramesSent and keyFramesSent are incremented.

ilyanikolaevskiy:
@vr000m , Thank you for review I have addressed your comments.

ilyanikolaevskiy:
@alvestrand, please take a look. May I merge this now?

PR #352, opened 2018-06-14
Added list of forbidden stats for isolated streams.

alvestrand:
Fixes https://github.com/w3c/webrtc-pc/issues/1613

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -3802,11 +3802,15 @@ function processStats() {
       <p>
         When applied to isolated streams, media metrics may allow an application to infer some
         characteristics of the isolated stream, such as if anyone is speaking (by watching the
-        <code>volume</code> statistic).
+        <code>audioLevel</code> statistic).
       </p>
-      <p class="issue">
-        TODO: Consider whether certain stats need to be not reported on <a data-cite=
-        "WEBRTC#isolated-media-streams">isolated media streams</a>.
+      <p>
+        The following stats are deemed to be sensitive, and MUST NOT be reported for an
+        <a data-cite="WEBRTC#isolated-media-streams">isolated media stream</a>:
+        <ul>
+          <li><a>RTCAudioHandlerStats.audioLevel</a>.</li>
+          <li><a>RTCRtpContributingSourceStats</a> (the whole dictionary).</li>
+        </ul>
       </p>
     </section>
     <section id="change-log*">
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -3802,11 +3802,15 @@ <h3>
...
         When applied to isolated streams, media metrics may allow an application to infer some
         characteristics of the isolated stream, such as if anyone is speaking (by watching the
-        <code>volume</code> statistic).
+        <code>audioLevel</code> statistic).
       </p>
-      <p class="issue">
-        TODO: Consider whether certain stats need to be not reported on <a data-cite=
-        "WEBRTC#isolated-media-streams">isolated media streams</a>.
+      <p>
+        The following stats are deemed to be sensitive, and MUST NOT be reported for an
```
ekr: These seem like a good start, but I don't think they are sufficient. For instance, the last frame height seems sensitive

henbos: What's sensitive about resolution? You could probably infer if there are frames produced or not by looking at bytes sent/received too.

vr000m: webrtc-pc says:
> Applications will be able to observe the parameters of the media that affect session negotiation and conversion into RTP. This includes the codecs that might be supported by the track, the bitrate, the number of packets, and the current settings that are set on the MediaStreamTrack.

I am reading this as being able to observe current settings of the MST. Since the application probably set the constraints in gUM or gDM, with isolated media streams, they will lose the ability to see the outcome of what was set, is that the intention here?

icydragons: If you look at the full paragraph it pretty clearly says that resolution is fine since it's not based on content.
---
Most of these data are exposed to the network when the media is transmitted. Only the settings for the MediaStreamTrack present a new source of information. This can includes the frame rate and resolution of video tracks, the bandwidth of audio tracks, and other information about the source, which would not otherwise be revealed to a network observer. Since settings don't change at a high frequency or in response to changes in media content, settings only reveal limited reveal information about the content of a track. However, any setting that might change dynamically in response to the content of an isolated MediaStreamTrack MUST have changes suppressed.

alvestrand: I also think that we should look at "what's useless to hide". An application is able to play an isolated media stream in a <video> element, and a video element exposes the width and height of its source: https://html.spec.whatwg.org/#dom-video-videowidth
But I may misunderstand what @ekr is referring to.

alvestrand:
We know that the list needs to be there, and this PR at least initializes the list. As we discover more properties that need to be protected, we can add them.

PR #355, opened 2018-06-27
Add TLS version and TLS group to RTCTransportStats

### LINKED ISSUE #279: Add TLS version and EC group id to stats
This is a spin-off from #276.

alvestrand:
Fixes #279


alvestrand:
Specref pull: https://github.com/tobie/specref/pull/471

alvestrand:
@vr000m it's actually 2 bytes, and there's fun and games played with it to avoid the kind of ossification that led to the original "version field" now being stuck at showing 1.2 forever. See @ekr comments on the bug.

vr000m:
Alright, I missed the hex representation part of the explanation.

LGTM.

PR #367, opened 2018-09-12
Rename RTCIceCandidateStats.ip to address.

jan-ivar:
Fix for https://github.com/w3c/webrtc-pc/issues/1913.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2958,7 +2958,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  It is the IP address of the candidate, allowing for IPv4 addresses, IPv6
+                  It is the address of the candidate, allowing for IPv4 addresses, IPv6
                   addresses, and fully qualified domain names (FQDNs). See
                   [[!RFC5245]] section 15.1 for details.
                 </p>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -2953,13 +2953,13 @@ <h2>
...
               <dt>
-                <dfn><code>ip</code></dfn> of type <span class=
+                <dfn><code>address</code></dfn> of type <span class=
                 "idlMemberType"><a>DOMString</a></span>
               </dt>
               <dd>
                 <p>
-                  It is the IP address of the candidate, allowing for IPv4 addresses and IPv6
-                  addresses, but fully qualified domain names (FQDNs) are not allowed. See
+                  It is the IP address of the candidate, allowing for IPv4 addresses, IPv6
```
fippo: is the address of the candidate

fippo: oh that is the prose from 5245... which no longer seems to be in the successor? anyway, :+1: on using that prose for consistency

jan-ivar: As discussed offline, this matches the language in [rfc5245](https://tools.ietf.org/html/rfc5245#section-15.1). I suppose it has a bit of a terminology issue. Anyone feel strongly which way it should be here?

jan-ivar: Turns out others agree with you, so I made the change. Thanks!

vr000m:
LGTM

jan-ivar:
@vr000m Thanks, can we merge this now that https://github.com/w3c/webrtc-pc/pull/1985 has merged?

vr000m:
Done. I was waiting for the webrtc-pc PR to be merged. With that done.

PR #368, opened 2018-09-12
fix rtcicecandidate url description

fippo:
also restrict it to local candidates

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -3016,9 +3016,12 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  The URL of the TURN or STUN server indicated in the that translated this IP
-                  address. It is the URL address surfaced in an
-                  <code>RTCPeerConnectionIceEvent</code>.
+                  For local candidates this is the URL of the ICE server from
+                  which the candidate was obtained. It is the same as the
+                  <a href="https://w3c.github.io/webrtc-pc/#rtcpeerconnectioniceevent">url surfaced in the RTCPeerConnectionIceEvent</a>.
+                </p>
+                <p>
+                  For remote candidates, this property is not applicable.
                 </p>
               </dd>
               <dt>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -3016,9 +3016,13 @@ <h2>
               </dt>
               <dd>
                 <p>
-                  The URL of the TURN or STUN server indicated in the that translated this IP
-                  address. It is the URL address surfaced in an
-                  <code>RTCPeerConnectionIceEvent</code>.
+                  For local candidates this is the URL of the ICE server from
```
jan-ivar: Nice catch. Like relayProtocal, we should probably say "This is only present for local candidates."

webrtc-stats.html
```diff
@@ -3016,9 +3016,13 @@ <h2>
...
+                  which the candidate was obtained. It is the same as the
+                  <code>url</code> surfaced in the
```
jan-ivar: Please link to http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnectioniceevent-url otherwise we get a markup loop.

webrtc-stats.html
fippo: see below where i coped from relayProtocol :-)

webrtc-stats.html
```diff
@@ -3016,9 +3016,12 @@ <h2>
...
                 <p>
-                  The URL of the TURN or STUN server indicated in the that translated this IP
-                  address. It is the URL address surfaced in an
-                  <code>RTCPeerConnectionIceEvent</code>.
+                  For local candidates this is the URL of the ICE server from
+                  which the candidate was obtained. It is the same as the
+                  <a href="https://w3c.github.io/webrtc-pc/#rtcpeerconnectioniceevent">url surfaced in the RTCPeerConnectionIceEvent</a>.
+                </p>
+                <p>
+                  For remote candidates, this property is not applicable.
```
jan-ivar: Can we s/applicable/present/ ?

fippo: will change in both places

fippo: hrm:
```
When a feature is not applicable to an instance of an object (for example audioLevel
          on a video stream), omit the dictionary member. Do NOT report a count of zero, -1 or
          "empty string".
```
We use "applicable" four times (up from 3). 
We do not use it for things like pliCount.

Keep as is an file an issue to be more consistent?

fippo:
changed "not applicable" which is not used in many places. The usage of "applicable" is not consistent anyway, see e.g. pliCount

Issue #339, opened 2018-03-07
Add stat for RTTs between client and STUN/TURN server


taylor-b:
We already have `totalRoundTripTime` in `RTCIceCandidatePairStats`, and `roundTripTime` in `RTCRemoteInboundRtpStreamStats`, but these are both an end-to-end round trip time.

I suggest adding a similar stat to `RTCIceCandidateStats`, which would accumulate round trip times as measured by STUN transactions between client and STUN server (from initial candidate gathering, keepalives for NAT bindings, TURN refreshes, etc.). We've encountered a couple uses for this:

- It can provide some information about the endpoint's Internet connection even before connecting to a remote endpoint.
- Once connected, it can help indicate which side of the connection is contributing more to latency.

So I'd suggest adding something like:

- `totalStunRoundTripTime`
- `totalStunResponsesReceived` (divide by this to get average RTT)
- `totalStunRequestsSent` (allows you to detect packet loss)

Or maybe `totalStunServerRoundTripTime` would be a better name, to make it extra clear that this is between the client and STUN server, and doesn't include connectivity checks.

fippo:
> It can provide some information about the endpoint's Internet connection even before connecting to a remote endpoint.
One can actually get some data on this by looking at the stun and turn candidate gathering time. But having more than two data points (for udp; only one for tcp) is better.

alvestrand:
Should this somehow be reported per STUN server?
If we have two STUN servers configured, it seems wrong somehow to report the average number. But if we need that level of complexity, we should introduce an RTCStunServerStats object.

fippo:
good point... this needs to live on local-candidate then? We *might* drop the "Stun" in the names then since we have no other RTT.

alvestrand:
Not sure localCandidate is the right place. One STUN server can end up giving you > 1 candidate, I think.
(or can it?)

fippo:
hrm... for relay candidates the candidate ip should make the localCandidate unique.

For STUN you're correct... if you ask two different stun servers you *might* get the same srflx address from both (depending on the nat type; see [this which is in vain](https://webrtchacks.com/symmetric-nat/)).
I would hope that the browser actually drops the one with the higher (initial) latency?

taylor-b:
My idea was for these fields to go on `RTCIceCandidateStats`, only for "local-candidate"s.

[...]

PR #372, opened 2018-09-18
add stunserverconnectionstats

lennart-csio:
Adds RTCStunServerConnectionStats as discussed as discussed in issue #339 .
I have added networkType, as I think it might be useful if the server is not responding (can be an access network issue).

Some questions:
1. do we want to have networkType here? if yes, should it remain as a subsection of ICE candidate stats?
2. with url, local address and port, I think we can uniquely link these stats to relevant candidates if wanted, or am I missing something? :)
3. is having url only to describe the server enough? with DNS load balancers it would be possible to have several servers behind a url

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -3639,7 +3639,6 @@ enum RTCNetworkType {
         <div>
           <pre class="idl">dictionary RTCStunServerConnectionStats : RTCStats {
              DOMString url;
-             DOMString address;
              long port;
              DOMString protocol;
              RTCNetworkType networkType;
@@ -3660,17 +3659,6 @@ enum RTCNetworkType {
               <dd>
                 <p>
                   The URL of the TURN or STUN server.
-                </p>
-              </dd>
-              <dt>
-                <dfn><code>address</code></dfn> of type <span class=
-                "idlMemberType"><a>DOMString</a></span>
-              </dt>
-              <dd>
-                <p>
-                  It is the IP address of the local interface, allowing for IPv4 addresses
-                  and IPv6 addresses, and fully qualified domain names (FQDNs).
-                  See [RFC5245] section 15.1 for details.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -3632,109 +3632,107 @@ enum RTCNetworkType {
           </section>
         </div>
       </section>
-
-    <section id="stunserverconnection-dict*">
-      <h3>
-        <dfn>RTCStunServerConnectionStats</dfn> dictionary
-      </h3>
-      <div>
-        <pre class="idl">dictionary RTCStunServerConnectionStats : RTCStats {
-           DOMString url;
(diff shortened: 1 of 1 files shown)


alvestrand:
Needs review.

webrtc-stats.html
```diff
@@ -3622,6 +3632,107 @@ <h2>
...
+        </h3>
+        <div>
+          <pre class="idl">dictionary RTCStunServerConnectionStats : RTCStats {
+             DOMString url;
+             DOMString address;
+             long port;
+             DOMString protocol;
+             RTCNetworkType networkType;
+             unsigned long totalRequestsSent;
+             unsigned long totalResponsesReceived;
```
henbos: I don't know much about STUN/TURN so feel free other people to pitch in, this looks sensible to me though.

Is there something worth adding for indicating whether or not this is still "active", such as the timestamp of the last request/response or some info if we lose connection, etc?

vr000m:
This was discussed at the TPAC18 session, what was the outcome?

webrtc-stats.html
```diff
@@ -3622,6 +3632,107 @@ <h2>
...
+                  The URL of the TURN or STUN server.
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>address</code></dfn> of type <span class=
+                "idlMemberType"><a>DOMString</a></span>
+              </dt>
+              <dd>
+                <p>
+                  It is the IP address of the local interface, allowing for IPv4 addresses
```
alvestrand: Just from the name, one would expect "address" to be the address of the STUN server, but this is a local interface address (sort of). Section 15.1 of RFC 5245 contains two addresses (connection-address and rel-addr), so isn't a precise enough reference.

If it's rel-addr (and rel-port, below), please use those names, suitably camel-cased. Note also that this may be different from the browser's address - see 5245 appendix B.3, which says that "the server reflexive address towards the TURN server is needed".

Revealing rel-addr is also a privacy issue - see JSEP section 3.5.3. This needs mentioning, and what to do if it's not exposed (I would recommend not returning the member rather than returning 0.0.0.0).

lennart-csio:
will address

webrtc-stats.html
lennart-csio: i looked into this a bit, and rel-addr seems not the right fit for this.
the problem is that rel-addr (afaik) relies on a response from the STUN/TURN server. this would mean that a response is required to identify the connection. however, in case of a failed connection to this server, it is valuable information to know which local interface is responsible.
so i'm wondering if it would be possible to actually get the local interface address, i.e. something that identifies the origin of the connection independent of a response?

alvestrand: why not use the URL of the STUN/TURN server? That's what's configured.

henbos:
@lennart-csio to remove address from this PR and we can merge it

alvestrand:
Lennart will remove the "address" field - it raises too many privacy issues. Then we merge.

Issue #376, opened 2018-10-18
[DataChannels] Use RTT from sctp in the stats


shacharz:
SCTP keeps an RTT parameter which is constantly being adjusted as data is being sent through the channel. Using that RTT can be significantly more accurate than using the ICE heartbeeting. Furthermore, browsers (such as Firefox) would find it easier to implement for data channel only peer connections. @nils-ohlmeier

alvestrand:
The SCTP RTT should probably be exposed on an SCTPTransport stats object. we don't have such an object at the moment, but we can add one.

alvestrand:
Is this exposed on the usrsctp API?

lgrahl:
Untested but it should be there: https://tools.ietf.org/html/rfc6458#section-8.2.2 (`spinfo_srtt`).

vr000m:
> SCTP RTT versus Using that RTT can be significantly more accurate than using the ICE heart-beating. 

Is the issue about accuracy a question about how often the RTT is calculated (i.e., SCTP updates per packet or window) versus ICE heartbeats happen at longer intervals (1s or similar)?

shacharz:
> Is the issue about accuracy a question about how often the RTT is calculated by the protocols (i.e., SCTP updates per packet or window) versus ICE heartbeats happen at longer intervals (1s or similar)?

yes, that's my understanding of it

vr000m:
Since, we expose RTT each level, it make sense to expose the round-trip time for SCTP. 
If you can make a PR, I can review it.

nils-ohlmeier:
> Is the issue about accuracy a question about how often the RTT is calculated by the protocols (i.e., SCTP updates per packet or window) versus ICE heartbeats happen at longer intervals (1s or similar)?

You are assuming that ICE measures RTT in the first place. Which Firefox ICE does not. So if we can get the SCTP RTT instead it helps.

vr000m:
@nils-ohlmeier I am not opposed to more accurate measurements from across the pipeline. If someone can make a PR, I am happy to review it.

henbos:
TPAC decision: Add RTCSctpTransportStats with RTT metrics. If there is an accumulative counter that is standardized so we can reference it prefer that, otherwise use [spinfo_srtt in RFC6458](https://tools.ietf.org/html/rfc6458#section-8.2.2) as a "the current smoothed round-trip time" but convert it to seconds.

PR #379, opened 2018-10-23
add rtt to rtcdatachannelsstats

shacharz:
adds `currentRoundTripTime` to `rtcdatachannelstats` as discussed in https://github.com/w3c/webrtc-stats/issues/376

First time PRing here so excuse the fouls if any.


lennart-csio:
is this the measured network RTT or is there some SCTP buffering included? would probably be good to know both?

shacharz:
> is this the measured network RTT or is there some SCTP buffering included? would probably be good to know both?

I meant for the RTT as measured by usrsctp, not sure what do you mean by additional buffering included

lennart-csio:
sorry, im not familiar with SCTP. is the RTT measured between sending a packet to the network and receiving a response?

shacharz:
> sorry, im not familiar with SCTP. is the RTT measured between sending a packet to the network and receiving a response?

on high level it's supposed to be calculated using the data packets sent and acks
https://github.com/sctplab/usrsctp/blob/master/Manual.md#calculate-rtt
I'm sure @lgrahl can share more details

alvestrand:
We should have an RFC reference for this calculation, otherwise implementations based on different SCTP stacks will return different values.

shacharz:
> We should have an RFC reference for this calculation, otherwise implementations based on different SCTP stacks will return different values.

@lgrahl can you help me reference the correct RFC here ?

vr000m:
@shacharz in addition to the clarification of the metric, it would be nice to know if this RTT is calculated per packet/ack. Depending on the freq of acks/packets, the RTT will be updated more often than getStats is called, we would need additional metrics to currentRoundTripTime, for example:

```
totalRoundTripTime
acksReceived
```

archives/20170614/webrtc-stats.html
```diff
@@ -3305,6 +3306,16 @@ <h4 id="dictionary-rtcdatachannelstats-members" resource="#dictionary-rtcdatacha
                   <code>RTCDatachannel</code>, i.e., not including headers or padding.
                 </p>
               </dd>
+
+              <dt>
+                <dfn data-dfn-for="rtcdatachannelstats" data-dfn-type="dfn" id="dom-rtcdatachannelstats-currentRoundTripTime" data-idl="" data-title="currentRoundTripTime"><code>currentRoundTripTime</code></dfn> of type <span class="idlMemberType">double</span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the latest round trip time measured in seconds, exposed by [<cite><a class="bibref" href="">usrsctp - spinfo_srtt</a></cite>].
```
alvestrand: The correct reference for spinfo_srtt is RFC 6458: https://tools.ietf.org/html/rfc6458#section-8.2.2

alvestrand:
Given that this PR is patching the wrong file (archive, not spec), and still hasn't cleared IPR, I'm closing this PR for now. You're very welcome to try again once the account linkage step is done.

Issue #374, opened 2018-09-28
Exposing RTCIceCandidateStats.networkType might trigger fingerprinting


youennf:
Isn't it an issue to expose the fact that a host is vpn for instance?

youennf:
Or being able to know whether someone is connected over ethernet vs. wifi...
I am also wondering what an application might do of this particular information.

vr000m:
This is already exposed for http://wicg.github.io/netinfo/, we should make sure that the stats document references it.

We have found `networkType` to be very helpful for diagnostics. (during a call and also afterwards). At least in the case for diagnostics, knowing if the user was on a WiFi helps a lot. 

Not having done any study, I cannot say if the additional information helps with fingerprinting. Because, in several enterprises/homes, the ethernet and wifi work in the same address range. There are also situations when the networkType is wrong, because someone is tethering (for example: LTE over WiFi).

youennf:
> This is already exposed for http://wicg.github.io/netinfo/, we should make sure that the stats document references it.

http://wicg.github.io/netinfo/ acknowledges that this API causes fingerprinting issues.
And it refers to WebRTC to state that this is already leaking elsewhere.
There would be a circular reference if WebRTC stats would refer to this API.
And the netinfo API is not as widely implemented/endorsed as WebRTC.

User agents that want to prevent fingerprinting in WebRTC will probably also want to prevent fingerprinting in netinfo.

Note also that VPN is not an option in netinfo, and some efforts have been made to prevent VPN detection through WebRTC.

> We have found `networkType` to be very helpful for diagnostics. (during a call and also afterwards). At least in the case for diagnostics, knowing if the user was on a WiFi helps a lot.

For diagnostics, couldn't you infer this in most cases by looking at:
- Network metrics like packet loss
- IP addresses, including private IP addresses if getUserMedia is granted

In general, the more expensive a fingerprinting method, the less attractive and used it will be.
Network info API/WebRTC stats are making it cheap to get this information.

> Not having done any study, I cannot say if the additional information helps with fingerprinting. Because, in several enterprises/homes, the ethernet and wifi work in the same address range. There are also situations when the networkType is wrong, because someone is tethering (for example: LTE over WiFi).

If the info is wrong enough to prevent fingerprinting, it should probably not be exposed as it might prevent good diagnostics.

vr000m:
> For diagnostics, couldn't you infer this in most cases by looking at:
> 
> Network metrics like packet loss

[...]

PR #380, opened 2018-10-31
Add networkType fingerprint warning

youennf:


lennart-csio:
currently the text reads that the implementation might just not implement it, but does that solve the issue? if browsers implement it, then there is still fingerprinting possible.
i know very little about the issue, so correct me if i'm wrong: would it be better to leave control over exposure of this parameter to the UA?

youennf:
This member is optional so it is ok to implement the corresponding stat but not expose that particular field. Implementations may also use 'unknown' but the intent of the first approach is clearer to me.

lennart-csio:
not entirely sure i understand. so lets say the browser implements this but an app or even a user thinks it is important to not expose this information. how does it help that we specify it "may not be implemented", because it is already implemented in the browser?

youennf:
I agree it would be good to reach consensus amongst browsers but we are not yet there.

That said, if an app thinks it is important to not expose this information, it should ignore this information. If some third party script is gathering stats on its behalf, the app should make sure to instruct the third party script to not gather that particular information.

For a user, I guess the only choice would be to use a browser that does not expose that information or provide a way to not expose that information.

youennf:
> so it is not possible to involve the user agent in this? if it can be disabled for this UA, the browser should just report unknown or similar

I am not exactly sure what you would like to see in the PR.
A UA may disable this by either reporting 'unknown' or not exposing the 'networkType' member.

Do you want this PR to require more from the UA?
Something like:
"User agents SHOULD/MUST allow users to disable the exposure of this information."
Or:
"User agents SHOULD NOT/MUST NOT expose this information without user consent."

juberti:
One outcome that I could live with would be to gate this behind gUM permission. (ugly, but perhaps the most pragmatic).

regarding 2119 language, I would say that UAs SHOULD omit this field unless the user has provided some form of consent.

youennf:
> what is the purpose of the last sentence "Implementations may decide to not expose this member."?

Just a way to express the fact that some implementations do not plan to expose this member. There is no constraints whatever the form.
There is some consensus between I and @juberti to make it a SHOULD.
@vr000m, @lennart-csio, is that ok with you? 
If so, I'll update the PR accordingly.

FWIW, I would tend to investigate various approaches:
- Never expose this information when there is no need: no connection or connection is good. It is always possible to abuse this but would require the fingerprinter to do actual network traffic which is costly
- Expose that information as part of UA information gathering, not as part of JS stats
- Enable the possibility for UA to easily generate more detailed reports that a user consciously accept to give to the web application, something like an easily uploadable chrome://webrtc-internals/

lennart-csio:
i think obtaining consent by the user is fine, but then there should be an active mandatory procedure to obtain this consensus.
what i dont want is that this is silently omitted and its just opt-in. i would rather have an opt-out, because imo its a really nice feature that i would like to use without much hassle.

juberti:
My main goal is that it should be possible for legitimate RTC apps to get access to this info. If it's only exposed as part of UA stats, that's not enough.

Exposing only if connection is good is probably not great, this will bias stats toward successful connections. I think that using a non-specific notion of consent as a SHOULD limitation on this feature is probably the least-bad option, that is at least consistent with what we're doing in ip-handling.

webrtc-stats.html
```diff
@@ -2978,6 +2978,11 @@ <h2>
                   relevant candidate would be <code>"wifi"</code>, even when the next hop is over a
                   cellular connection.
                 </div>
+                <p class="fingerprint">
+                  This reveals information that would otherwise not be available to web pages,
```
juberti: TBH I don't think this is a big issue for fingerprinting (one can already look up the service provider based on the IP and make a good guess at connection type), but it might be a more significant issue for people trying to use VPNs.

youennf:
> I think that using a non-specific notion of consent as a SHOULD limitation on this feature is probably the least-bad option, that is at least consistent with what we're doing in ip-handling.

I do not think we can spec a specific notion of consent.
For instance, a UA might have a toggle to enter in mode where more analytic data is given to some/all web apps as a way to improve these apps.

youennf:
> My main goal is that it should be possible for legitimate RTC apps to get access to this info. If it's only exposed as part of UA stats, that's not enough.

This is actually difficult to classify legitimate in the web though.

One use case I heard is that issues were popping here and there and, that after gathering lots of data, the network type info was able to pinpoint VPN as a potential issue.
That is why I am mentioning the case of bad connections.

vr000m:
We run pre-call tests agains customer's or our infrastructure, in this case callstats creates a peerconnection and runs a slew of network tests over data channels. In this case, since there is no gUM prompt, we would collect and run this test over the default interface. 

I would assume that in this case, we would know that the call would go over Wireless, or VPN, or something similar, upon detection callstats throws an event to the app and state clearly this call is going over 3G or VPN or ... and if the user wants to do the call over that particular default network type. If they think that this would go over the wrong network type (they get a chance to fix that before the call starts, i.e., before gUM consent, etc).

youennf:
> We would like to know that the call would go over Wireless, or VPN, or something similar (network type), because upon detection callstats throws an event to the app and state clearly this call is going over 3G or VPN or ... and if the user wants to do the call over that particular default network type. If they think that this would go over the wrong network type (they get a chance to fix that before the call starts, i.e., before gUM consent, etc).

This seems interesting to put in https://github.com/w3c/webrtc-stats/issues/374 as this is a new use case for this info. In case gUM kicks in and browser goes from mode 2 to mode 1, the chosen route might not be VPN anymore but some other route. This might be confusing to users. 

Also, I am not clear what your opinion on the PR is.
Are you pushing against the added SHOULD statement? Do you prefer the initial attempt?

vr000m:
> Are you pushing against the added SHOULD statement? 

Yes, I want to make sure if a UA wants to expose this info before gUM, it should be possible. 

> Do you prefer the initial attempt?

Since all keys are optional, the guidance  "may not expose" is not really helpful, since it should also when it should reveal this info.  Otherwise the may not expose seems like never expose this.

webrtc-stats.html
```diff
@@ -2978,6 +2978,11 @@ <h2>
...
+                  which can increase the fingerprint surface. The user agent SHOULD expose this
+                  field only if the user provides some form of consent.
```
alvestrand: I don't think we should have SHOULD statements inside a "fingerprint" note - it's easy to overlook.
Conditional access to stats is something we need to be pretty explicit about if we're going to have any hope of making consistent behaviour out of it. So I'd recommend just going with the fingerprint warning here.

alvestrand:
@youennf @vr000m says you had discussed some changes to this. Can you update?

youennf:
I think we can scope the PR to just the warning.

As is defined webrtc-stats, there is no requirement to implement any such stat.
If we were to mandate a SHOULD requirement on this value, I guess it would be in webrtc-pc.

webrtc-pc currently makes networkType and implementation (the two stats that have fingerprinting issues) as optional. It is strange that the address field is not marked as a fingerprint risk.

There could be a statement in webrtc-stats acknowledging the issue that some stat values (list could be automatically generated) have fingerprint risks and that specs using stats should provide guidance on how/when to expose those fields.

PR #381, opened 2018-10-31
Add a note controlling exposure of remote candidate addresses

youennf:


webrtc-stats.html
```diff
@@ -2989,6 +2989,17 @@ <h2>
                   addresses, and fully qualified domain names (FQDNs). See
                   [[!RFC5245]] section 15.1 for details.
                 </p>
+                <div class="note">
```
alvestrand: Don't use "note" for a normative statement - this paragraph has a MUST NOT. Just use <p>.

webrtc-stats.html
```diff
@@ -2989,6 +2989,17 @@ <h2>
...
+                  A user agent should make sure to limit exposure of remote candidate addresses to
+                  the ones already known to the web application, in particular peer reflexive remote
+                  candidate addresses. If a user agent cannot tell whether a remote candidate
+                  address is known to the web application, it MUST NOT surface this address value in
+                  any RTCICECandidateStats dictionary. Once a RTCPeerConnection instance is notified
+                  by the web application of an address using <a
+                  href="https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-addicecandidate">addIceCandidate</a>
+                  the user agent can surface the 'address' member in any remote RTCIceCandidateStats
+                  dictionary of the corresponding RTCPeerConnection.
```
alvestrand: I think we need to say something other than "MUST NOT surface".
The choices I see are:
- Omit the whole stat (which would hurt overall clarity of stats)
- Leave out the address string
- Generate a name in the same way we generate MDNS names, and enter that instead, possibly with a prefix that allows identification of "hidden" rather than "mdns".
A quick scan finds an IP address only in the candidate stats - offhand, I'd prefer alternative 3.

youennf: The idea was to not have any 'address' member in RTCICECandidateStats in that case since this is not a required field. I'll clarify the PR.
I can go with alternative 3 but I do not really see the value.
Is the concern that scripts might break if this field does not show up?

youennf:
Uploaded related WPT test PR at https://github.com/web-platform-tests/wpt/pull/13922

webrtc-stats.html
vr000m: If it is known that the address is peer-reflexive, I would go with option 3. Although would that not require a reserved value for hidden? 

However, if we go with option 2, I am assuming omit means null, I want to make sure that we set it to null only when it needs to be hidden. Although it might be set to null when the remote address is unknown?

youennf: Omit means 'undefined'.
I am fine with an explicit 'hidden' if that helps but I am wondering whether that is worth it.
If we want to make it explicit, I would go with null instead of undefined, which would probably require changing the IDL.

youennf:
> Uploaded related WPT test PR at [web-platform-tests/wpt#13922](https://github.com/web-platform-tests/wpt/pull/13922)

Written in terms of 'undefined'

webrtc-stats.html
```diff
@@ -2989,6 +2989,19 @@ <h2>
                   addresses, and fully qualified domain names (FQDNs). See
                   [[!RFC5245]] section 15.1 for details.
                 </p>
+                <p>
+                  A user agent should make sure to limit exposure of remote candidate addresses to
+                  the ones already known to the web application, in particular peer reflexive remote
```
alvestrand: The test actually makes it clear that it's not "the ones already known to the web application", it's "the ones already known to be in use on this peerconnection". In the test, the web application knows all the addresses (from the "other" PC), but it doesn't "know" that one PC's local address is in use by the other PC as a remote address.

Suggested rephrase:

"An user agent should make sure that only remote candidate addresses that the web application has configured on the peerconnection are exposed; in particular peer reflexive remote canididate addresses that don't match a candidate set by addIceCandidate."

alvestrand:
@youennf ping?

webrtc-stats.html
youennf: I initially thought that keeping some flexibility in the way implementations would handle that could be good.
The added test is somehow a degenerated case since the web page already learned the IP address.

I agree though that such smarter implementation would be more complex, will not provide a lot of benefit.
It would also make the test not fully authoritative. Let's mandate it then.

webrtc-stats.html
alvestrand: JS makes a difference between "undefined" and "null". We should have "null" here.

vr000m:
LGTM

Issue #312, opened 2018-01-25
video packet counters should be moved out from RTCRTPStreamStats to appropriate dictionaries


vr000m:
Since the changes in #272 were a lot of work, did we just miss these or was there a reason that we chose to keep them in RTCRTPStreamStats and not move them to appropriate dictionaries? 

```c
    unsigned long      firCount;
    unsigned long      pliCount;
    unsigned long      nackCount;
    unsigned long      sliCount;
    unsigned long long qpSum;
```

vr000m:
@jan-ivar could you comment if these stats are in the right dictionary.

vr000m:
Is the sender reporting these?

alvestrand:
Should be on local sender and local receiver RTP stream, since the counts aren't transferred in RTCP.

henbos:
There are lots of sender/receiver (track) stats that may or may not make sense to move to/from RTP stream stats, related issue: #318

henbos:
@vr000m Can you create a PR?

jan-ivar:
The reason I didn't move them initially is I misunderstood their descriptions—unsure what "sent by receiver" meant in the context of rtp/rtcp. From @alvestrand's https://github.com/w3c/webrtc-stats/issues/312#issuecomment-361946378 it seems:

* firCount: *"only valid for video and is sent by receiver"* -> RTCInboundRtpStreamStats?
* pliCount: *"only valid for video and is sent by receiver"* -> RTCInboundRtpStreamStats?
* sliCount: *"only valid for video and is sent by receiver"* -> RTCInboundRtpStreamStats?
* nackCount: *"Count ... packets received by the sender and is sent by receiver"* -> ???
* qpSum: *"sum of the QP values of ... framesDecoded for inbound ... framesEncoded for outbound"* -> RTCInboundRtpStreamStats and RTCOutboundRtpStreamStats?

Except for *nackCount* they're all video stats AFAICT. I can attempt a PR if someone tells me where *nackCount* goes.

jan-ivar:
*qpSum* is also only for some video codecs, I gather?

PR #384, opened 2019-01-15
Move {fir|pli|nack|sli}Count and qpSum out of baseclass to {in|out}bound (Editorial)

jan-ivar:
Fix for https://github.com/w3c/webrtc-stats/issues/312.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -720,11 +720,6 @@ enum RTCStatsType {
              DOMString           kind;
              DOMString           transportId;
              DOMString           codecId;
-             unsigned long       firCount;
-             unsigned long       pliCount;
-             unsigned long       nackCount;
-             unsigned long       sliCount;
-             unsigned long long  qpSum;
 };</pre>
           <section>
             <h2>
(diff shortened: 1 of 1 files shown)


jan-ivar:
[Preview](https://rawgit.com/w3c/webrtc-stats/4809e859b8f23b180f39aa13a433eb8c7ccb6a11/webrtc-stats.html#inboundrtpstats-dict*)

alvestrand:
This means duplication of the stats into two dictionaries. Can we avoid the duplication of the descriptions? (ie replace one set with "see <other> for definition")?

alvestrand:
QP is the one that has a long description that might get out of sync. The other ones are short.

alvestrand:
Jan-Ivar will merge when the definition of qp is deduplicated.

alvestrand:
@vr000m would like to review this prior to merge - discussing whether they should be nacksReceived and nacksSent rather than nackCount.

vr000m:
When it was in the base class, the ambigious *Count name made sense because the text described how the counters were updated at the sender and receiver. Now that these metrics are being moved to the appropriate dictionary, we can add the suffix -Sent/-Received qualifier. 

If we are going down this path, I would encourage the -suffix, but that would require retiring the -Count metrics.

Browsers that have implemented these -Count metrics should also chime in? @henbos @aboba @jan-ivar @youennf

na-g:
@vr000m, Firefox implements most of those counts, and I don't foresee being able to remove support for the *Count names any time soon. If the goal in renaming them is to remain consistent with other packet counters, should they be *PacketsSent/*PacketsReceived instead?

jan-ivar:
@vr000m This PR is mostly editorial, shuffling around dictionaries in a manner hardly observable by JS. It is 99.9% web compatible (only minor reordering of members by name).

It therefore does not represent a renaming opportunity the way it might in other languages.

Renaming members would have significant web compat impact and should be considered independent of this editorial PR IMHO.

na-g:
@jan-ivar and @vr000m, is there anything holding this PR request back that shouldn't be addressed in another issue?

jan-ivar:
@vr000m @alvestrand In the interest of merging this, I've opened a separate issue https://github.com/w3c/webrtc-stats/issues/390 on `nacksReceived` and `nacksSent`.

OK to merge?

PR #387, opened 2019-02-07
initial commit for webrtc-stats-testing

### LINKED ISSUE #346: Add a "panel test" to track implementation status
If we're going to advance to PR, we need to know what the implementation status is.
Having a "panel" based on KITE or WPT should give us a list of stats and who implements them.

vr000m:
Fixes #346

### DIFF
--- a/webrtc-stats-testing.md
+++ b/webrtc-stats-testing.md
@@ -10,7 +10,7 @@ Tests are split in three parts: API semantics, availability of metrics, and corr
 
 Tests if the getStats API returns the correct stats object depending on the API input (i.e., filter by senders, etc). Also these tests should check if the reported statistics conform to the peerconnection state machine, i.e., check when obsolete stats are removed when the conditions are met, etc. The test should be carried out in a way that all the RTCStatsType dictionaries are covered, however, in some cases it may make sense to test parts of the media pipeline as a whole, for example sender and receiver transport and media (audio and video) stats.
 
-Web Platform Tests [WPT] currently runs these tests, available at: https://wpt.fyi/results/webrtc-stats
+Web Platform Tests [WPT] currently runs these tests, available at: https://wpt.fyi/results/webrtc-stats. There are also browser specific tests [Chrome-Tests].
 
 ## Availability of metrics
 
@@ -23,7 +23,7 @@ Tests if all metrics specified as mandatory are returned in the stats. This shou
 + When the peer connection is in failed state, this is when a call is dropped 
 + When the peer connection is in closed state, this is after call has ended.
 
-In every case, the test should produce a connection between two PeerConnections, and call getStats()on them. In these tests, we are mainly verifying three aspects:
+In every case, the test should produce a connection between two PeerConnections, and call getStats() on them. In these tests, we are mainly verifying three aspects:
 
 1. The stats objects and the corresponding metrics are created in the correct peerconnection phases. 
 
@@ -43,3 +43,4 @@ Additionally, the validation should take into account when various state changes
 [KITE] KITE is a test engine designed to test WebRTC interoperability across browsers https://github.com/webrtc/KITE
 [WPT] Web Platform Test for webrtc-stats https://wpt.fyi/results/webrtc-stats
 [VERIFY] Verify the getStats() implementation status for different browsers  https://webrtc-stats.callstats.io/verify/
+[Chrome-tests] Chrome's Web Platform Tests for WebRTC https://cs.chromium.org/chromium/src/third_party/blink/web_tests/external/wpt/webrtc/
--- a/webrtc-stats-testing.md
+++ b/webrtc-stats-testing.md
@@ -10,7 +10,7 @@ Tests are split in three parts: API semantics, availability of metrics, and corr
 
 Tests if the getStats API returns the correct stats object depending on the API input (i.e., filter by senders, etc). Also these tests should check if the reported statistics conform to the peerconnection state machine, i.e., check when obsolete stats are removed when the conditions are met, etc. The test should be carried out in a way that all the RTCStatsType dictionaries are covered, however, in some cases it may make sense to test parts of the media pipeline as a whole, for example sender and receiver transport and media (audio and video) stats.
 
-Web Platform Tests [WPT] currently runs these tests, available at: https://wpt.fyi/results/webrtc-stats. There are also browser specific tests [Chrome-Tests].
+Web Platform Tests [WPT] currently runs these tests, available at: https://wpt.fyi/results/webrtc-stats. There are also browser specific tests [WPT-CHROME].
 
 ## Availability of metrics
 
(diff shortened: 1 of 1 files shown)


webrtc-stats-testing.md
```diff
@@ -0,0 +1,45 @@
...
++ Correctness and Validation of metrics: Using KITE with a test harness for media and network, checks the correctness of the reported stats.
+
+Tests are split in three parts: API semantics, availability of metrics, and correctness of metrics.
+
+## API Semantics
+
+Tests if the getStats API returns the correct stats object depending on the API input (i.e., filter by senders, etc). Also these tests should check if the reported statistics conform to the peerconnection state machine, i.e., check when obsolete stats are removed when the conditions are met, etc. The test should be carried out in a way that all the RTCStatsType dictionaries are covered, however, in some cases it may make sense to test parts of the media pipeline as a whole, for example sender and receiver transport and media (audio and video) stats.
+
+Web Platform Tests [WPT] currently runs these tests, available at: https://wpt.fyi/results/webrtc-stats
```
henbos: There are several getStats tests at https://cs.chromium.org/chromium/src/third_party/blink/web_tests/external/wpt/webrtc/ too. Should we link to it or move stuff around? The webrtc directory relies on RTCPeerConnection-helper.js

webrtc-stats-testing.md
```diff
@@ -0,0 +1,45 @@
...
+Tests if all metrics specified as mandatory are returned in the stats. This should be verified against the phases of the peerconnection: 
+
++ When the peer connection is in new state, this is before the call is setup
++ When the peer connection is in connecting state, this is while setting up a call, 
++ When the peer connection is in connected state, this is when a call is established 
++ When the peer connection is in disconnected state, this is when a call has an intermittent connection issue
++ When the peer connection is in failed state, this is when a call is dropped 
++ When the peer connection is in closed state, this is after call has ended.
+
+In every case, the test should produce a connection between two PeerConnections, and call getStats()on them. In these tests, we are mainly verifying three aspects:
```
henbos: nit: getStats() [space] on

webrtc-stats-testing.md
vr000m: If you know which level of testing they apply to, then make a proposal, I will add it to the doc.

vr000m:
@dontcallmedom @henbos any feedback, should I go ahead and merge, and continue improving the document via new issues?

webrtc-stats-testing.md
```diff
@@ -0,0 +1,46 @@
...
++ Conformance across browsers: Using KITE [KITE], which is a test engine designed to test WebRTC interoperability across browsers. This is especially important to verify if the metrics reported in the RTP Control Protocol (RTCP) Sender Report (SR), Receiver Report (RR), or Extended Reports (XR) are reported correctly. 
++ Correctness and Validation of metrics: Using KITE with a test harness for media and network, checks the correctness of the reported stats.
+
+Tests are split in three parts: API semantics, availability of metrics, and correctness of metrics.
+
+## API Semantics
+
+Tests if the getStats API returns the correct stats object depending on the API input (i.e., filter by senders, etc). Also these tests should check if the reported statistics conform to the peerconnection state machine, i.e., check when obsolete stats are removed when the conditions are met, etc. The test should be carried out in a way that all the RTCStatsType dictionaries are covered, however, in some cases it may make sense to test parts of the media pipeline as a whole, for example sender and receiver transport and media (audio and video) stats.
```
alvestrand: I'm not sure what the second sentence (from "The test should be carried out....") adds. If something needs to be said, I think you may want to say "Each test should verify the content of a single dictionary. However, some tests need to cover multiple parts of the media pipeline, for instance checking that the stats in an RTPSender are consistent with the stats in the corresponding RTPReceiver."

webrtc-stats-testing.md
```diff
@@ -0,0 +1,46 @@
...
+
+Web Platform Tests [WPT] currently runs these tests, available at: https://wpt.fyi/results/webrtc-stats. There are also browser specific tests [WPT-CHROME].
```
alvestrand: Add some words about the current test coverage of WPT (good, bad or ugly, not percentages).

webrtc-stats-testing.md
```diff
@@ -0,0 +1,46 @@
...
+
+Verification for availability of metrics for Chrome, Firefox, and Safari is available at: https://webrtc-stats.callstats.io/verify/
+
+## Correctness of metrics
+These set of tests are supposed to validate that the implementation returns valid (close to correct values). For example, that the sent packet counters are increasing and the corresponding received packet counters are less than equal to the sent counters.
+
+Additionally, the validation should take into account when various state changes take place, i.e.,  RTCSignalingState, RTCIceGatheringState, RTCIceConnectionState, RTCPeerConnectionState.
+
+
+### References
```
alvestrand: This isn't a subsection of "Correctness", so bump it up a level.

webrtc-stats-testing.md
```diff
@@ -0,0 +1,46 @@
...
++ When the peer connection is in failed state, this is when a call is dropped 
++ When the peer connection is in closed state, this is after call has ended.
+
+In every case, the test should produce a connection between two PeerConnections, and call the getStats API on those peerconnections. We are mainly verifying three aspects with these tests, they are:
+
+1. The stats objects and the corresponding metrics are created in the correct peerconnection phases. 
+
+2. The stats objects and fields are returned in the correct format or type (integer, double, String, etc). Additionally, we are verifying that a value is returned.
+
+It should be noted that if the getStats API is called at short intervals, they might return cached values. These tests need to make sure that the measurement results are not incorrectly returned because of browser internals caching the result. If specification is unclear on any of these aspects, the specification needs to be updated.
```
alvestrand: Add "The spec specifies transitions for which caching is not allowed. The tests need to verify that this restriction is obeyed."

webrtc-stats-testing.md
```diff
@@ -0,0 +1,46 @@
...
+
+Verification for availability of metrics for Chrome, Firefox, and Safari is available at: https://webrtc-stats.callstats.io/verify/
+
+## Correctness of metrics
+These set of tests are supposed to validate that the implementation returns valid (close to correct values). For example, that the sent packet counters are increasing and the corresponding received packet counters are less than equal to the sent counters.
```
alvestrand: Since you have WPT tests and KITE tests as two separate sections up above, you need to be clear about whether these are WPT tests, KITE tests or both. (I think we'll need both.)

webrtc-stats-testing.md
henbos: Perhaps the only thing we should say is that some metrics have relationships between peer connections and should be tested together. But even that ("should be tested together") I'm not sure is true, you can verify send counters going up independently of receive counters going up. I am generally in favour of splitting up tests by behaviour, and doing as little in a single test as possible. E.g. I shouldn't fail a test about bytesSent because I don't implement bytesReceived. As for comparing exact numbers, KITE may be good enough. End-to-end tests are probably useful too, but I see them as a bonus rather than a requirement of testing.

webrtc-stats-testing.md
henbos: "The tests must be agnostic towards which caching mechanisms may or may not be in place by the implementation. In particular, the spec defines under which circumstances caching is not allowed. In any other circumstances, our tests must not assume non-cached values in calls to getStats()."

PR #406, opened 2019-03-18
Clarify concealedSamples

### LINKED ISSUE #405: Clarify concealedSamples
concealedSamples is said to be "data that was synthesized to conceal packet loss". But the packet might not have been lost, it may arrive late. Let's add a clarification.

henbos:
Fixes #405.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1948,7 +1948,9 @@ enum RTCStatsType {
                 <p>
                   Only present for inbound audio tracks. The total number of inbound audio samples
                   that are concealed samples. A concealed sample is a sample that is based on data
-                  that was synthesized to conceal packet loss and does not represent incoming data.
+                  that was synthesized to conceal packet loss or packets arriving late and does not
+                  represent incoming data. These samples are generated when we don't have the
+                  necessary non-concealed samples in the buffer.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2694,10 +2694,10 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  The total number of inbound audio samples that are concealed samples. A concealed
-                  sample is a sample that is based on data that was synthesized to conceal packet
-                  loss or packets arriving late and does not represent incoming data. These samples
-                  are generated when we don't have the necessary non-concealed samples in the buffer.
+                  The total number of audio samples that are concealed samples. A concealed sample is a
+                  sample that is based on data that was synthesized to conceal packet loss or packets
+                  arriving late and does not represent incoming data. These samples are generated when
+                  we don't have the necessary non-concealed samples in the buffer.
                 </p>
               </dd>
               <dt>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -2695,8 +2695,9 @@ <h2>
               <dd>
                 <p>
                   The total number of samples that are concealed samples. A concealed sample is a
-                  sample that is based on data that was synthesized to conceal packet loss and does
-                  not represent incoming data.
+                  sample that is based on data that was synthesized to conceal packet loss or
+                  packets arriving late and does not represent incoming data. These samples are
+                  generated when we don't have the necessary non-concealed samples in the buffer.
```
vr000m: The total number of samples that conceal samples. A conceal sample is a sample that is based on data that was synthesized to **conceal samples from a missing packet during playout**. Concealment samples may be inserted when a packet is lost (reported in <a>packetsLost</a>) or the packet arrives to late/early to be played out (reported in <a>packetsDiscarded</a>).

vr000m: I dont quite remember what this means --- "does not represent incoming data" --> is this to disambiguate against datachannel?

vr000m: > These samples are generated when we don't have the necessary non-concealed samples in the buffer.

I dont think this is accurate, if playing out timestamp `t` and is missing packets, and the packets in the Jitter Buffer are already present at (t+200ms), would the application not add concealment samples??

ivocreusen:
I think "does not represent incoming data" means that the concealment audio
does not come from incoming audio packets, but is made up by the jitter
buffer to conceal the fact that a packet was missing (either lost or
delayed).

If we are playing out at timestamp t and don't have the necessary packet
available, but we do have a packet at t+200ms available the jitter buffer
should indeed play out concealment samples. I don't think that is
contradictory with the text though.

webrtc-stats.html
henbos: See ivoc's reply.

webrtc-stats.html
```diff
@@ -2695,8 +2695,10 @@ <h2>
               <dd>
                 <p>
                   The total number of samples that are concealed samples. A concealed sample is a
-                  sample that is based on data that was synthesized to conceal packet loss and does
-                  not represent incoming data.
+                  sample that is based on data that was synthesized to conceal missing samples
+                  during playout. Concealed samples may be inserted when a packet is lost
+                  (reported in <a>packetsLost</a>) or the packet arrives too late to be played out
+                  (reported in <a>packetsDiscarded</a>).
```
alvestrand: Is it too late to change the name? I find the name "concealed" very confusing - since the samples we're counting is samples that were synthesized, not the (unknown) samples they were synthesized to conceal.

If we want to keep the "concealed" name, rewrite as: "A concealed sample is a sample that was replaced with synthesized samples generated locally before being played out. Examples of samples that have to be concealed are samples from lost packets (reported in packetsLost) or samples from packets that arrive too late to be played out (reported in packetsDiscarded).

The "examples of" language allows us to use the counter for samples that are removed for other reasons, such as being identifies as Doritos munching, keyboard clicks or sounds too loud to be played out (if that makes sense).

henbos: It is too late to change the name (this is already shipped in Chrome). Update the definition as you suggested, making it more flexible, though if we had different reasons to conceal we might need more stats to help distinguish between "concealing" and other reasons in the future.

PR #410, opened 2019-03-18
Add silentConcealedSamples

### LINKED ISSUE #409: Add counters for the subset of concealedSamples that are not silent
There is a qualitative difference between concealed samples that were produced as silence and concealed samples that were produced as non-silence. Having counters for this is important in order to measure audio quality.

Proposal:
- audibleConcealedSamples: The total number of concealed samples inserted that are non-silent. This is a subset of concealedSamples.

Products that are relying on Chrome's non-standard googSpeechExpandRate could use this metric instead. This unblocks migration efforts on such products.

henbos:
Fixes #409.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2711,7 +2711,6 @@ enum RTCStatsType {
                 </p>
               </dd>
               <dt>
-              <dt>
                 <dfn><code>concealmentEvents</code></dfn> of type <span class=
                 "idlMemberType"><a>unsigned long long</a></span>
               </dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1594,6 +1594,7 @@ enum RTCStatsType {
              unsigned long long  totalSamplesReceived;
              double              totalSamplesDuration;
              unsigned long long  concealedSamples;
+             unsigned long long  audibleConcealedSamples;
              unsigned long long  concealmentEvents;
              double              jitterBufferDelay;
              RTCPriorityType     priority;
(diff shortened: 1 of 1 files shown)


henbos:
Please take another look @vr000m, I defined these as silentConcealedSamples instead. Anyone interested in audibleConcealedSamples can calculate it as concealedSamples - silentConcealedSamples. I defined a silent samples as "Playing out silent samples result in no audio".

> In some cases, concealedSamples may be equal to silentSamples, because no non-silent samples were added.

Indeed, I don't think this needs to be called out though? (It's still a subset.)

webrtc-stats.html
```diff
@@ -2699,6 +2700,16 @@ <h2>
...
                 </p>
               </dd>
+              <dt>
+                <dfn><code>silentConcealedSamples</code></dfn> of type <span class=
+                "idlMemberType"><a>unsigned long long</a></span>
+              </dt>
+              <dd>
+                <p>
+                  The total number of concealed samples inserted that are silent. Playing out
+                  silent samples result in no audio. This is a subset of <a>concealedSamples</a>.
```
alvestrand: So if the samples are generated as comfort noise with a non-zero volume, they are not counted as silent?
Nit: result in no audio -> results in no audio. Or perhaps "results in silence", since that's what's being played out (but makes it sound like a tautology).

henbos: Still waiting to see what Chrome's implementation does, but this makes sense for now and PR updated, since what you are about to distinguish is between "no noise" and "noise or someone talking", the latter may result in robot voices, etc.

henbos: I checked with Chrome - we count comfort noise as silence, and the spec is now clear on that point.

PR #412, opened 2019-03-18
Add fecPacketsDiscarded

### LINKED ISSUE #411: Add error correction metrics
Some packets contain contain payload for error correction, which may end up getting used or it may end up getting discarded. Having metrics for this indicate the health of the connection, and because error correction payloads may not result in as high quality as the "real" payloads, this is also good to help diagnose quality issues.

Proposals:
- errorCorrectionPacketsReceived: The total number of packets received that contain error correction payload. These are packets conveying information of previous packets, used to conceal packet loss. This is a subset of packetsReceived. To calculate the total number of error correction packets that was decoded, subtract errorCorrectionPacketsDiscarded from errorCorrectionPacketsReceived.
- errorCorrectionPacketsDiscarded: The total number of packets received that contain error correction payload, but the error correction payload was discarded due to having received the original packet. This is a subset of errorCorrectionPacketsReceived.

These may help replace Chrome's non-standard googSecondaryDecodedRate and googSecondaryDiscardedRate, unblocking products that rely on these for migrating to the standardised getStats.

henbos:
Fixes #411.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1127,6 +1127,7 @@ enum RTCStatsType {
              DOMHighResTimeStamp  lastPacketReceivedTimestamp;
              double               averageRtcpInterval;
              unsigned long        fecPacketsReceived;
+             unsigned long        fecPacketsDiscarded;
              unsigned long long   bytesReceived;
              unsigned long        packetsFailedDecryption;
              unsigned long        packetsDuplicated;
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1234,6 +1235,17 @@ <h2>
...
                 </p>
               </dd>
+              <dt>
+                <dfn><code>fecPacketsDiscarded</code></dfn> of type <span class=
+                "idlMemberType"><a>unsigned long</a></span>
+              </dt>
+              <dd>
+                <p>
+                  Total number of RTP FEC packets received for this SSRC where the error correction
+                  payload was discarded due to having received the original packet. This is a
```
vr000m: Discard can also happen FEC packets arriving after the recovery window has expired.

vr000m: Total number of RTP FEC packets received for this SSRC where the error correction payload was discarded by the application. This may happen 1. if all the source packets protected by the FEC packet were received, or 2. if the FEC packet arrived late, i.e., outside the recovery window, and the lost RTP packets have already been skipped during playout.

henbos:
Thanks, please take another look, I updated it to use your more complete definition.

vr000m:
> 1. if all the source packets protected by the FEC packet were received

Would this be better:

1. if all the source packets protected by the FEC packet were received or already recovered by a separate FEC packet.

henbos:
Thanks, I incorporated your last comment as well.

Issue #414, opened 2019-03-24
References to undefined RTCReceiverAudioTrackAttachmentStats and RTCReceiverVideoTrackAttachmentStats


dontcallmedom:
The definition of `trackId` in`RTCInboundRtpStreamStats` refers to `RTCReceiverAudioTrackAttachmentStats` and `RTCReceiverVideoTrackAttachmentStats`, but no such dictionaries exist.

vr000m:
This may depend on the Issue opened by @henbos on #403.

alvestrand:
RTCSenderAudioTrackAttachmentStats exists. I think we're missing the other 3 (empty) dictionaries.

henbos:
I filed #424 which covers this one, merging this into that.

Issue #403, opened 2019-03-06
Impact on Track Stats because of moving stats from sender/receiver to appropriate in/outbound rtp


vr000m:
Related #402.

vr000m:
@pallab-gain @karthikbr82 could you make sure that the tracks have the right stats once @henbos makes the appropriate changes to RTC[In/Out]boundRtpStreamStats

vr000m:
We probably to discuss replaceTrack and onstatsended but also to discuss if the trackattachment stats  are needed and instead rely on track stats. 

cc: @henbos @pallab-gain @karthikbr82

henbos:
If "onstatsended" fires for the track attachment we could do getStats() to get outbound-rtp values at the time of the event, though it wouldn't be a perfect snapshot because the counters in the outbound-rtp might keep increasing.

henbos:
Let's discuss further when #463 has landed

henbos:
#463 has landed

henbos:
I just filed #479 arguing for the removal of these dictionaries. I think "what to do about onstatsended?" should be discussed in a separate issue, so I filed #480.

With those two issues, I think we can close this issue as having been split up into two issues. Feel free to object.

PR #415, opened 2019-03-24
Fix warnings from respec on broken automatic links

dontcallmedom:
Still 2 warnings due to #414


vr000m:
@dontcallmedom Merge this and keep #414 open until #403 is done?

dontcallmedom:
Sgtm

PR #417, opened 2019-04-03
totalEncodedBytesTarget added

### LINKED ISSUE #416: Rethinking RTCOutboundRtpStreamStats.targetBitrate
Knowing something about the target encode bitrate is important, for example when debugging issues related to too large frames being encoded.

The current definition of [targetBitrate](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-targetbitrate) is:

> It is the current target bitrate configured for this particular SSRC and is the Transport Independent Application Specific (TIAS) bitrate [RFC3890]. Typically, the target bitrate is a configuration parameter provided to the codec's encoder and does not count the size of the IP or other transport layers like TCP or UDP. It is measured in bits per second and the bitrate is calculated over a 1 second window.

Two problems:
- Measuring the bitrate over a 1 second window is not what implementations necessarily should do. Chrome doesn't. We should not force implementation strategies about when this bitrate can and cannot update the target.
- This is a bitrate. If an application would like calculate the difference between target and actual bytes produced, there is no way to do this. You would have to estimate target total bytes based on bitrate and compare to total bytes sent. Why doesn't the stat provide accumulative counters instead?

**Proposal**
Replace targetBitrate with targetTotalBytesEncoded, to be defined as:

> The total number of bytes the encoder would have produced if it exactly matched the configured target. In reality, the actual number of bytes may be higher or lower on a frame-by-frame basis. This value is updated each time a frame has been encoded.

This definition is agnostic towards when or how the target is updated, the counter would go up with whatever the target was for each frame, which makes it directly comparable to bytesSent.

henbos:
Fixes #416.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1501,7 +1501,7 @@ enum RTCStatsType {
              DOMString            remoteId;
              DOMHighResTimeStamp  lastPacketSentTimestamp;
              double               targetBitrate;
-             unsigned long long   totalTargetBytesEncoded;
+             unsigned long long   totalEncodedBytesTarget;
              unsigned long        framesEncoded;
              unsigned long long   qpSum;
              double               totalEncodeTime;
@@ -1570,20 +1570,19 @@ enum RTCStatsType {
                   Transport Independent Application Specific (TIAS) bitrate [[!RFC3890]].
                   Typically, the target bitrate is a configuration parameter provided to the
                   codec's encoder and does not count the size of the IP or other transport layers
-                  like TCP or UDP. It is measured in bits per second. This statistic gives the
-                  configured target bitrate at the time of the stats timestamp.
+                  like TCP or UDP. It is measured in bits per second and the bitrate is calculated
+                  over a 1 second window.
                 </p>
               </dd>
               <dt>
-                <dfn><code>totalTargetBytesEncoded</code></dfn> of type <span class=
+                <dfn><code>totalEncodedBytesTarget</code></dfn> of type <span class=
                 "idlMemberType">unsigned long long</span>
               </dt>
               <dd>
                 <p>
-                  The total number of bytes the encoder would have produced if it exactly matched
-                  the configured target. In reality, the actual number of bytes may be higher or
-                  lower on a frame-by-frame basis. This value is updated each time a frame has
-                  been encoded by incrementing its value by target bytes for that frame.
+                  This value is increased by the target frame size in bytes every time a frame
+                  has been encoded. The actual frame size may be bigger or smaller than this
+                  number. This value goes up every time <code>framesEncoded</code> goes up.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1501,6 +1501,7 @@ enum RTCStatsType {
              DOMString            remoteId;
              DOMHighResTimeStamp  lastPacketSentTimestamp;
              double               targetBitrate;
+             unsigned long long   totalTargetBytesEncoded;
              unsigned long        framesEncoded;
              unsigned long long   qpSum;
              double               totalEncodeTime;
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1569,8 +1570,20 @@ <h2>
...
+              <dt>
+                <dfn><code>totalTargetBytesEncoded</code></dfn> of type <span class=
+                "idlMemberType">unsigned long long</span>
+              </dt>
+              <dd>
+                <p>
+                  The total number of bytes the encoder would have produced if it exactly matched
+                  the configured target. In reality, the actual number of bytes may be higher or
+                  lower on a frame-by-frame basis. This value is updated each time a frame has
+                  been encoded by incrementing its value by target bytes for that frame.
```
alvestrand: Suggested formulation: "This value is increased by the target frame size in bytes every time a frame has been encoded. The actual frame size may be bigger or smaller than this number."

I'm not sure the target frame size is constant. Do keyframes get a larger target size than non-keyframes?

webrtc-stats.html
```diff
@@ -1569,8 +1570,20 @@ <h2>
...
                   Typically, the target bitrate is a configuration parameter provided to the
                   codec's encoder and does not count the size of the IP or other transport layers
-                  like TCP or UDP. It is measured in bits per second and the bitrate is calculated
-                  over a 1 second window.
+                  like TCP or UDP. It is measured in bits per second. This statistic gives the
+                  configured target bitrate at the time of the stats timestamp.
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>totalTargetBytesEncoded</code></dfn> of type <span class=
```
alvestrand: totalEncodedBytesTarget?

webrtc-stats.html
vr000m: Target Size is for the frame to be encoded, right?

Would changing "has been encoded" --> "is to be encoded" mislead implementors into counting target bytes when no frame was generated?

webrtc-stats.html
vr000m: do we need Frames?  because the bytes belong to a target frame size.
do we need total? although we have not used total for other metrics. Perhaps in this case we do need it.

webrtc-stats.html
alvestrand: Yes, I think it would mislead people.
OTOH, if we drop a frame because we have overshot the target bitrate in the previous frame, then it makes sense to increase the totalEncodedBytesTarget, so perhaps "is to be encoded" is the right formulation.

vr000m: want to make sure a non-encoded frame should not be counted towards the byte.

henbos: Clarify: only goes up when framesEncoded goes up.
Split up PR into new stat and clarifying old stat

vr000m: btw what is the difference between the bytesSent (payload only) and totalTargetBytesEncoded?

One difference that I foresee is that the two metrics are updated at different moments in time.

vr000m:
Split the discussion on 1-second window... (We are undoing #13). At the time we thought TIAS was needed for the bitrate definition and TIAS seemed appropriate and perhaps at the time, rates needed a window to be defined. 

This brings in if TIAS is really appropriate for target bitrate, since TIAS is the maximum session bitrate. 

> The Transport Independent Application Specific Maximum (TIAS)
   bandwidth modifier has an integer bit-rate value in bits per second.
   A fractional bandwidth value SHALL always be rounded up to the next
   integer.  The bandwidth value is the maximum needed by the
   application (SDP session level) or media stream (SDP media level)
   without counting IP or other transport layers like TCP or UDP. 

In this case, 1-second window is perhaps not needed because it is possible that the encoded may never achieve this rate. We should also point out that the target bit rate can change often, as often as generating a new frame or upon receiving a packet.

PR #419, opened 2019-04-03
Add retransmittedPacketsSent and retransmittedBytesSent

### LINKED ISSUE #418: RTCOutboundRtpStreamStats.retransmitted[Bytes/Packets]Sent
In the case of RTX this happens on separate SSRCs and this separate RTCOutboundRtpStreamStats with separate bytesSent counters, but for good ol' fashioned retransmissions we don't have counters. The existing [bytesSent](https://w3c.github.io/webrtc-stats/#dom-rtcsentrtpstreamstats-bytessent) is "Total number of bytes sent for this SSRC".

Being able to tell how many bytes/packets were retransmission is useful for network diagnostic, and being able to tell how much of the bytes/packets was not retransmitted is useful for comparing that total with the expected output from the encoder (finding out of the encoder overshoots etc).

**Proposal**
Add retransmittedBytesSent and retransmittedPacketsSent as the "Total number of retransmitted bytes/packets sent for this SSRC."

henbos:
Fixes #418.


webrtc-stats.html
```diff
@@ -1559,6 +1561,24 @@ <h2>
...
+                  The total number of packets that were retransmitted for this SSRC.
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>retransmittedBytesSent</code></dfn> of type <span class=
+                "idlMemberType">unsigned long long</span>
+              </dt>
+              <dd>
+                <p>
+                  The total number of bytes that were retransmitted for this SSRC.
```
alvestrand: include "payload bytes only"

henbos: Clarify that this only means payload bytes.

vr000m:
No issues, please merge when the bytes to use payload bytes is clarified.

webrtc-stats.html

PR #421, opened 2019-04-03
totalPacketSendDelay added

### LINKED ISSUE #420: Stats for the delay of packets being sent
Proposal: A metric that measures the time between each packet being produced and it being sent onto the network.
This is similar to jitter buffer delay but for the sender.

henbos:
Fixes #420.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1504,6 +1504,7 @@ enum RTCStatsType {
              unsigned long        framesEncoded;
              unsigned long long   qpSum;
              double               totalEncodeTime;
+             double               totalPacketSendDelay;
              double               averageRtcpInterval;
              RTCQualityLimitationReason                 qualityLimitationReason;
              record&lt;DOMString, double&gt; qualityLimitationDurations;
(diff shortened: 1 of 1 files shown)


vr000m:
This makes sense. And we should do it.

However, I think we need to decide what we mean "produced". Since we are talking about packets and not Frames. Packets are produced when a video frame is fragmented or audio samples are concatenated. 

So in the case of a video frame, all fragmented packets will have the produced timer start at the same time and the timer will expire when they are sent, which would mean that we could observe a send delay spike for large frames.

webrtc-stats.html
```diff
@@ -1617,6 +1618,18 @@ <h2>
...
               </dd>
+              <dt>
+                <dfn><code>totalPacketSendDelay</code></dfn> of type <span class=
+                "idlMemberType">double</span>
+              </dt>
+              <dd>
+                <p>
+                  The total number of seconds that packets have spent buffered locally before being
+                  transmitted onto the network. Measurement starts when a packet has been produced
+                  and ends when the packet is sent. <code>totalPacketSendDelay</code> is incremented
```
alvestrand: Needs to define what "is sent" means - probably it is when the send() call is executed on the actual socket - which is after the webrtc congestion control, but before the tcp congestion control when the socket is a TCP socket. "last possible time to report before the clock is sent" would do.

webrtc-stats.html
```diff
@@ -1617,6 +1618,18 @@ <h2>
...
                 </p>
               </dd>
+              <dt>
+                <dfn><code>totalPacketSendDelay</code></dfn> of type <span class=
+                "idlMemberType">double</span>
+              </dt>
+              <dd>
+                <p>
+                  The total number of seconds that packets have spent buffered locally before being
+                  transmitted onto the network. Measurement starts when a packet has been produced
```
alvestrand: Define "has been produced" if possible.

webrtc-stats.html
```diff
@@ -1617,6 +1618,18 @@ <h2>
...
+                  and ends when the packet is sent. <code>totalPacketSendDelay</code> is incremented
+                  for each packet sent by this measurement for that packet.
```
henbos: Clarify: When is the clock started, e.g. when a large frame is split up into multiple packets.
Clarify: When the clock stops, i.e. does this include delays for passing this over to the network?

henbos: Will clarify to say something like...
Start time: When the packet was emitted from the RTP packetizer.
End time: When the packet was handed to the OS network socket.

henbos:
Travis failing: It looks unrelated, it says "ImportError: No module named html5lib"

Issue #402, opened 2019-03-06
Move all stream stats from the "sender"/"receiver" stats


henbos:
Any RTP media related metrics should be placed on "outbound-rtp" or "inbound-rtp" accordingly. One "sender" could have multiple "outbound-rtp" due to simulcast, so having counters on a per-"sender" basis is wrong. For symmetry, conceptual correctness, and for more future-proof way of thinking, we should do the same thing for "receiver" and "inbound-rtp", even though today a "receiver" only ever has a single "inbound-rtp" object.

vr000m:
This may have an impact on track stats...

henbos:
See also #392

vr000m:
How do we resolve this one, @henbos -- quite a few stats depend on this issue.

henbos:
I might be able to take a stab at it before next meeting, but I can't promise anything, I'm trying to ramp up on a new project. Otherwise do we have anyone on your side that can take it?

henbos:
With #463 and #466, this is now fixed.

PR #435, opened 2019-05-15
Add RTCMediaSourceStats dictionaries for pre-encoding metrics

### LINKED ISSUE #400: Add RTCMediaSourceStats
Stats for a source (track) like captured resolution, framerate, whether or not the track is ended, etc.

### LINKED ISSUE #422: Capture vs encode frameWidth, frameHeight and framesPerSecond
These currently exist in [RTCVideoHandlerStats](https://w3c.github.io/webrtc-stats/#dom-rtcvideohandlerstats). For frameWidth/Height, "width/height of the last processed frame for this track" it is not clear if "processed frame" of outbound media refers to frames before encoding (captured frames) or frames after encoding (i.e. frames sent). But framesPerSecond is clear: "For sending tracks it is the current captured FPS and for the receiving tracks it is the current decoding framerate."

We should have metrics for both captured and encoded frames. Related issues: [RTCMediaSourceStats](https://github.com/w3c/webrtc-stats/issues/400) and [Move all stream stats from the "sender"/"receiver" stats](https://github.com/w3c/webrtc-stats/issues/402).

**Proposal**
- For frames sent and received, we should have these metrics on a per-RTP stream bases (simulcast compatible).
- For frames captured, we should have these metrics on a "source stats" object.

henbos:
Fixes #400. Fixes #422.
Related work not done in this PR: [Move all stream stats from the "sender"/"receiver" stats](https://github.com/w3c/webrtc-stats/issues/402).

The parent dictionary is called RTCMediaSourceStats with type "media-source" (rather than RTCMediaStreamTrackStats or type "track" because those names are already in use by current/legacy dictionaries) and it has two subdictionaries: RTCAudioSourceStats and RTCVideoSourceStats.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -432,6 +432,7 @@ enum RTCStatsType {
 "outbound-rtp",
 "remote-inbound-rtp",
 "remote-outbound-rtp",
+"media-source",
 "csrc",
 "peer-connection",
 "data-channel",
@@ -513,6 +514,15 @@ enum RTCStatsType {
               stream that is currently received with this <code>RTCPeerConnection</code> object. It
               is measured at the remote endpoint and reported in an RTCP Sender Report (SR). It is
               accessed by the <code><a>RTCRemoteOutboundRtpStreamStats</a></code>.
+            </p>
+          </dd>
+          <dt>
+            <dfn><code>media-source</code></dfn>
+          </dt>
+          <dd>
+            <p>
+              Statistics for the media produced by a local <code>MediaStreamTrack</code> that is
+              currently attached to an <code>RTCRtpSender</code>.
             </p>
           </dd>
           <dt>
@@ -1497,6 +1507,7 @@ enum RTCStatsType {
         <div>
           <pre class="idl">dictionary RTCOutboundRtpStreamStats : RTCSentRtpStreamStats {
              DOMString            trackId;
+             DOMString            mediaSourceId;
              DOMString            senderId;
              DOMString            remoteId;
              DOMHighResTimeStamp  lastPacketSentTimestamp;
@@ -1530,6 +1541,14 @@ enum RTCStatsType {
                 The identifier of the stats object representing the current track attachment to the
                 sender of this stream, an <code><a>RTCSenderAudioTrackAttachmentStats</a></code> or
                 <code><a>RTCSenderVideoTrackAttachmentStats</a></code>.
+              </dd>
+              <dt>
+                <dfn><code>mediaSourceId</code></dfn> of type <span class=
+                "idlMemberType">DOMString</span>
+              </dt>
+              <dd>
+                The identifier of the stats object representing the track currently attached to the
+                sender of this stream, an <code><a>RTCMediaSourceStats</a></code>.
               </dd>
               <dt>
                 <dfn><code>senderId</code></dfn> of type <span class=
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1497,6 +1507,7 @@ <h3>
         <div>
           <pre class="idl">dictionary RTCOutboundRtpStreamStats : RTCSentRtpStreamStats {
              DOMString            trackId;
+             DOMString            mediaSourceId;
```
henbos: Note: It is enough to have the mediaSourceId on the "sender" object but because currently browsers don't implement "sender" we place the ID here. This is the same reason that it already has "trackId".

webrtc-stats.html
```diff
@@ -515,6 +516,15 @@ <h3>
...
             </p>
           </dd>
+          <dt>
+            <dfn><code>media-source</code></dfn>
+          </dt>
+          <dd>
+            <p>
+              Statistics for the media produced by a local <code>MediaStreamTrack</code> that is
+              currently attached to an <code>RTCRtpSender</code>.
+            </p>
```
alvestrand: Add what the type of the dictionary is - "it is accessed by the <code>DictionaryName</code>" is the standard incantation above, even though the words don't make much sense.

webrtc-stats.html
```diff
@@ -1531,6 +1542,14 @@ <h2>
                 sender of this stream, an <code><a>RTCSenderAudioTrackAttachmentStats</a></code> or
                 <code><a>RTCSenderVideoTrackAttachmentStats</a></code>.
               </dd>
+              <dt>
+                <dfn><code>mediaSourceId</code></dfn> of type <span class=
+                "idlMemberType">DOMString</span>
+              </dt>
+              <dd>
+                The identifier of the stats object representing the track currently attached to the
+                sender of this stream, an <code><a>RTCMediaSourceStats</a></code>.
```
alvestrand: I'm worried about having stats defined because of the present state of our implementation, instead of where they logically belong. How many people will implement source stats before sender stats?
And what about track stats? If we're putting this in for backwards compatibility, shouldn't the track get one too?

webrtc-stats.html
```diff
@@ -1860,6 +1879,84 @@ <h2>
...
         </div>
       </section>
+      <section id="mediasourcestats-dict*">
+        <h3>
+          RTCMediaSourceStats dictionary
+        </h3>
+        <p>
+          The <dfn>RTCMediaSourceStats</dfn> dictionary represents a local track that is attached
+          to a sender. It contains information about media sources such as frame rate and
+          resolution pre-encoding. This is in contrast to <code>RTCOutboundRtpStreamStats</code>
```
alvestrand: "pre-encoding" is not grammatical. Say "... and resolution prior to encoding".

webrtc-stats.html
```diff
@@ -1860,6 +1879,84 @@ <h2>
...
+          whose members describe transmission and post-encoding metrics. For example, a track may
```
alvestrand: "describe metrics as measured after the encoding step".

webrtc-stats.html
```diff
@@ -1860,6 +1879,84 @@ <h2>
...
+          be captured from a high-resolution camera but be transmitted in low resolution due to
+          CPU and network conditions.
```
alvestrand: ... and track constraints?

webrtc-stats.html
```diff
@@ -1860,6 +1879,84 @@ <h2>
...
+            class="dictionary-members">
+              <dt>
+                <dfn><code>width</code></dfn> of type <span class=
+                "idlMemberType">unsigned long</span>
+              </dt>
+              <dd>
+                <p>
+                  Only applicable to video sources. The width, in pixels, of the
+                  last frame originating from this source that was passed to the
+                  sender for further processing. Before the first frame has been
```
henbos: I just realized one "media source" (track) can be used by multiple senders. I'll update all of these "passed to the sender" to say something like "is ready for further processing". Alternative suggestions are welcomed.

webrtc-stats.html
```diff
@@ -1860,6 +1879,84 @@ <h2>
...
+            </h2>
+            <dl data-link-for="RTCMediaSourceStats" data-dfn-for="RTCMediaSourceStats"
+            class="dictionary-members">
+              <dt>
+                <dfn><code>width</code></dfn> of type <span class=
+                "idlMemberType">unsigned long</span>
+              </dt>
+              <dd>
+                <p>
+                  Only applicable to video sources. The width, in pixels, of the
```
alvestrand: Given that every single object here has "Only applicable to video sources", should we rename the dictionary "RTCVideoSourceStats"?

webrtc-stats.html
henbos: I clarified the life cycle at the top and here just said "originating from the source"

webrtc-stats.html
henbos: I was planning to implement source stats before sender stats in Chrome. Having access to resolution prior to encoding is a "migration blocker" that I want done for M76. Moving "track" stats to "sender" is less important right now. (And moving "track" stats to "outbound-rtp" is mostly important for simulcast, not migration either)

The "track" stats get mediaSourceId already in this PR, they're inherited from RTCVideoSenderStats and RTCAudioSenderStats to RTCSenderVideoTrackAttachmentStats and RTCSenderAudioTrackAttachmentStats.

webrtc-stats.html

henbos:
> Query: If two tracks are sourced from the same source, set up with different constraints, and attached to the same PeerConnection, will there be one or two source objects?

The "source" in the context of WebRTC refers to tracks (would have called them "track" stats if that wasn't already in-use). Asking "where does the track get its media from?" is leaving the scope of WebRTC.

Current proposal:
- Two tracks with different constraints will show up as two different media sources even if they originate from the same camera.
- A track that is attached to multiple senders will show up as a single media source. This is relevant for the total frame counter.

henbos:
@vr000m to take a look

alvestrand:
Editor meeting decision: Varun to review, looks close, but need time to review.

webrtc-stats.html
```diff
@@ -515,6 +516,17 @@ <h3>
               accessed by the <code><a>RTCRemoteOutboundRtpStreamStats</a></code>.
             </p>
           </dd>
+          <dt>
+            <dfn><code>media-source</code></dfn>
+          </dt>
+          <dd>
+            <p>
+              Statistics for the media produced by a <code>MediaStreamTrack</code> that is currently
```
vr000m: It should be clear that the data here is input to the encoder after gUM constraints are applied and these are not raw media source metrics.

webrtc-stats.html
```diff
@@ -1860,6 +1881,168 @@ <h2>
...
+        </h3>
+        <p>
+          The <dfn>RTCAudioSourceStats</dfn> dictionary represents an audio
+          track that is attached to one or more senders. It is an
+          <code><a>RTCMediaStreamSource</a></code> whose <code>kind</code> is
+          <code>"audio"</code>.
+        </p>
+        <div>
+          <pre class="idl">dictionary RTCAudioSourceStats : RTCMediaSourceStats {
+};</pre>
```
vr000m: confirming we currently have no stats for audio source.

webrtc-stats.html

webrtc-stats.html
henbos: We may want to put stuff like input energy levels here, but no, currently there's nothing here.

henbos:
Travis is broken for unrelated reasons; merging.

Issue #399, opened 2019-03-06
Remove RTC[Audio/Video][Sender/Receiver]Stats.remoteSource


henbos:
This information is already available based on dictionary type ("sender" means not remote, "receiver" means remote). It is still needed for ye olde "track" stats though.

henbos:
Move to deprecated

jan-ivar:
I agree with this. Removal also solves the ambiguity of its language: *"True if the source is remote, for instance if it is sourced from another host via an RTCPeerConnection"*

In the [forwarding case](https://webrtc.github.io/samples/src/content/peerconnection/multiple-relay/), where a track from one peer connection is feed into another, this would return `true`, confusing code relying on `remoteSource` to deduce whether it is sending or receiving.

> It is still needed for ye olde "track" stats though.

`"track"` stats in the spec are never remote.

Since I'm doing an audit of mandatory stats, I've taken the liberty of providing a PR. @henbos PTAL.

vr000m:
For an application layer multicast, each node in the tree is forwarding media from one hop-to-another. In this case, a peerconnection will take the output from one peerconnection and attach it to the other peerconnection. 

In this case, we would like to keep the `remoteSource`.

henbos:
I don't understand what multicast or piping peer connections have to do with remoteSource. It should have no relevance? Can you explain what you would expect to happen?

vr000m:
I dont see an issue with remoteSource as defined: "True if the source is remote, for instance if it is sourced from another host via an RTCPeerConnection. False otherwise."

vr000m:
https://github.com/w3c/webrtc-stats/pull/148

PR #442, opened 2019-06-06
Deprecate remoteSource

jan-ivar:
Fixes https://github.com/w3c/webrtc-stats/issues/399.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2263,7 +2263,6 @@ enum RTCStatsType {
         <div>
           <pre class="idl">dictionary RTCMediaHandlerStats : RTCStats {
              DOMString           trackIdentifier;
-             boolean             remoteSource;
              boolean             ended;
              DOMString           kind;
              RTCPriorityType     priority;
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -4249,6 +4240,25 @@ <h2>Obsolete RTCInboundRtpStreamStats members</h2>
...
+          <h2>Obsolete RTCMediaHandlerStats members</h2>
+          <dl data-dfn-for="RTCMediaHandlerStats">
+            <dt>
+              <dfn><code>remoteSource</code></dfn> of type <span class=
+                "idlMemberType">boolean</span>
+            </dt>
+              <dd>
+                This field got removed in June 2019, as it was redundant and
+                sometimes ambiguous (when a track from one peer connection is
+                fed into another). A <code>type</code> of
```
alvestrand: I don't see how it's ambiguous. A track fed from one PC to another PC would have remoteSource = true; remoteSource says nothing about its destination.

henbos: remoteSource means the stats are "receiver" stats (as opposed to "sender" stats). This is needed for legacy implementations (like current Chrome) that have "track" stats in both sending and receiving directions. Can we add a receiver track attachment dictionary that is inheriting from the receiver stats dictionaries that only exists in the obsolete section? And have the only new member to this dictionary the remoteSource?

In this case "remoteSource" is not redundant, because the type is "track" for both sending and receiving directions. This is how Chrome works today, so we can't remove remoteSource.

As for "ambiguous (when a track from one peer connection fed into another)" I don't understand this. remoteSource means its either a sender track attachment or a receiver track attachment. The "track" stats object _does not_ refer to the track; only to the attachment. It matters not whether the track came from getUserMedia() or from another peer connection; the attachment is "remoteSource == false" if the attachment refers to an RTCRtpSender, and "remoteSource == true" if the attachment refers to an RTCRtpReceiver.

henbos: Another way to think of remoteSource is "did the MediaStreamTrack that the trackIdentifier refers to get created by this RTCPeerConnection or by other means?"

vr000m:
The origin is in https://github.com/w3c/webrtc-stats/pull/148, lets keep discussing in #399

PR #450, opened 2019-06-26
Add RTCInboundRtpStreamStats.totalDecodeTime

### LINKED ISSUE #434: Add RTCInboundRtpStreamStats.totalDecodeTime
In #150 we added totalEncodeTime. Should we add totalDecodeTime too?
Arguments for and against would be similar as for totalEncodeTime. This could replace googDecodeMs in Chrome's legacy getStats API.

drkron:
Fixes #434

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1231,9 +1231,9 @@ enum RTCStatsType {
               </dt>
               <dd>
               <p>
-                 Total number of seconds that have been spent decoding the framesDecoded frames
+                 Total number of seconds that have been spent decoding the <a>framesDecoded</a> frames
                  of this stream. The average decode time can be calculated by dividing this
-                 value with framesDecoded. The time it takes to decode one frame is the time
+                 value with <a>framesDecoded</a>. The time it takes to decode one frame is the time
                  passed between feeding the decoder a frame and the decoder returning decoded
                  data for that frame.
               </dd>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1147,6 +1147,7 @@ enum RTCStatsType {
              DOMString            remoteId;
              unsigned long        framesDecoded;
              unsigned long long   qpSum;
+             double               totalDecodeTime;
              DOMHighResTimeStamp  lastPacketReceivedTimestamp;
              double               averageRtcpInterval;
              unsigned long        fecPacketsReceived;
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1224,6 +1225,18 @@ <h2>
...
+                <dfn><code>totalDecodeTime</code></dfn> of type
+                <span class="idlMemberType">double</span>
+              </dt>
+              <dd>
+              <p>
+                 Total number of seconds that have been spent decoding the framesDecoded frames
+                 of this stream. The average decode time can be calculated by dividing this
+                 value with framesDecoded. The time it takes to decode one frame is the time
+                 passed between feeding the decoder a frame and the decoder returning decoded
+                 data for that frame.
```
alvestrand: This definition uses wall clock time as the basis for deciding how much time was spent.
curiosity: Is there any case where decoding would pause while waiting for something to happen, so that the time between "decode start" and "decode ends" is not an accurate metric of resource (CPU) usage?
If this is a sufficiently weird case that it's not worth mentioning, we can just disregard this comment.

drkron: I have reused the text from totalEncodeTime.

I've been told that there exist HW encoder/decoders that use a pipeline in such a way that there's a one-frame delay. In this case the decode time would correctly be higher than 1/fps even though it's not a meaningful metric of the resource usage. It's actually the overuse case of the decoder that I'm interested in, I'll create an issue in WebRTC PC for overuse detection.

Looking a bit more closely at this: Should totalDecodeTime be here or in RTCVideoReceiverStats?

webrtc-stats.html
```diff
@@ -1224,6 +1225,18 @@ <h2>
                   formats have ways to vary the quantizer value within the frame.
                 </p>
               </dd>
+              <dt>
+                <dfn><code>totalDecodeTime</code></dfn> of type
+                <span class="idlMemberType">double</span>
+              </dt>
+              <dd>
+              <p>
+                 Total number of seconds that have been spent decoding the framesDecoded frames
```
henbos: nit: &lt;a&gt;framesEncoded&lt;/a&gt;

webrtc-stats.html
henbos: I think it should be in "inbound-rtp" (RTCInboundRtpStreamStats) where you have currently placed it, we are trying to move things around, see #402.

This is most important on the sending side, since one "sender" can have multiple "outbound-rtp" due to multiple simulcast layers. Whereas on the receiving side, simulcast is not supported, so we only ever have one "inbound-rtp" per "receiver". But conceptually putting RTP stream metrics on the "receiver" rather than the "inbound-rtp" is wrong, and if a "receiver" were ever to support receiving multiple "inbound-rtp" for whatever reason, we would be in trouble if all the metrics were placed in the "receiver" dictionary.

webrtc-stats.html
drkron: Fixed!

webrtc-stats.html
drkron: Thanks for the explanation, the difference between "inbound-rtp" and "receiver" wasn't clear to me.

vr000m:
I like this. 

To know the total time a frame spent before decoding, I can add jitterBufferDelay + totalDecodeTime, assuming jitterBufferEmittedCount and framesDecoded line up.

vr000m:
LGTM, can be merged

PR #451, opened 2019-06-26
Move audio levels to "media-source" and "receiver"

### LINKED ISSUE #449: Outbound audio levels
audioLevel, totalAudioEnergy and totalSamplesDurations are currently on the "track" or "sender" stats objects. This raises several questions in my head:
- They are tied to samples sent, should all of these really be moved to "outbound-rtp"?
- Hang on, if these are samples sent, that means that using these for audio indication of the microphone is broken if you are not connected?

Instead, should we move these to "media-source"? Or have similar metrics in both "media-source" (the levels of the local track) and audio levels sent over RTP ("outbound-rtp")?

henbos:
Fixes #449.

Before this PR:
- audioLevel, totalAudioEnergy and totalSamplesDuration were present in both "sender" and "receiver" dictionaries.

After this PR:
- audioLevel, totalAudioEnergy and totalSamplesDuration are present in "receiver" and "media-source" instead.

The difference is even if we are not sending samples (we are not connected), the "media-source" can provide energy levels based on samples produced by the media source (the MediaStreamTrack).

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1979,6 +1979,9 @@ enum RTCStatsType {
         </p>
         <div>
           <pre class="idl">dictionary RTCAudioSourceStats : RTCMediaSourceStats {
+              double              audioLevel;
+              double              totalAudioEnergy;
+              double              totalSamplesDuration;
 };</pre>
           <section>
             <h2>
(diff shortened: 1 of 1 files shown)


henbos:
Currently in chrome, the "receiver" stats (receiving "track") are implemented, but the "sender" stats (sending "track") always return 0. In fixing this bug I realized that the spec says only to calculate on samples sent, but the use case of local levels is usually to show how the microphone is operating. If network problems cause the audio meter to go quiet that would be confusing to debug.

henbos:
By having this on the "media-source" instead it is enough to attach a track to the PC, and we get audio levels regardless of connection status.

vr000m:
I like that the metrics are on the media source. However, having it on the sent is also meaningful to see if the encoder is doing something funny with the source input.

henbos:
Travis errors are unrelated, to be fixed here: https://github.com/w3c/webrtc-stats/issues/455

PR #459, opened 2019-07-18
Remove references to the notion of objectDeleted in tracks

youennf:
objectDeleted is not really defined.
This wording is probably redundant with https://w3c.github.io/webrtc-stats/#dom-rtcstatstype-track, which refers to RTCRtpSender track change.


vr000m:
I thought we were going to keep the replaced tracks dangling for a while, because there is no event that indicates that getStats needs to be called when the track is being replaced.

Issue #356, opened 2018-07-10
RTCIceCandidateStats.deleted can be removed


na-g:
This field was made redundant by the addition of the onstatsended event, though the accompanying prose may still be useful.

vr000m:
@youennf could you PTAL at this issue and create a PR?

youennf:
Will do, thanks for the reminder!

youennf:
There are potential slight differences between the deleted candidate value definition and the definition of when a candidate is deleted as per https://w3c.github.io/webrtc-stats/#dom-rtcstatstype-local-candidate.

In theory, an ICE transport could free the candidate when it is not used by the selected candidate pair.
The same could apply to a candidate pair that 'disappears' before an ICE restart is done or the connection closed.

I guess for now we can remove the deleted attribute and decide whether we want/need to precise https://w3c.github.io/webrtc-stats/#dom-rtcstatstype-local-candidate and https://w3c.github.io/webrtc-stats/#dom-rtcstatstype-candidate-pair later on.

vr000m:
@youennf Are you asking how to mark unselected by successful ice-candidate-pairs?

henbos:
This was fixed in #462

PR #462, opened 2019-07-18
Remove RTCIceCandidateStats deleted property

youennf:
Related to https://github.com/w3c/webrtc-stats/issues/356


vr000m:
I wonder if the text should be retained somewhere, to explain that this stat may not appear anymore because the candidate or pair was freed/deleted:
> For host candidates, this means that
                  any network resources (typically a socket) associated with the candidate have
                  been released. For TURN candidates, this means the TURN allocation is no longer
                  active.

vr000m:
@youennf Could you add this to the obsolete section

PR #463, opened 2019-07-24
Move "sender" stats to "outbound-rtp", "inbound-rtp" and "media-source"

henbos:
Partial fix for #402.

The following stats have been moved:

**RTCVideoHandlerStats**
- frameWidth, frameHeight: Moved to RTCOutboundRtpStreamStats and RTCInboundRtpStreamStats represending encoded and decoded frames.
- framesPerSecond: For source stats, we already have RTCVideoSourceStats.framesPerSecond (i.e. camera FPS). In this PR framesPerSecond was added to RTCOutboundRtpStreamStats, but representing the encoding frame rate, which previously did not exist. For the receiving case, this was moved to RTCInboundRtpStreamStats representing the decoding frame rate.

**RTCVideoSenderStats**
- framesCaptured: Obsolete. We already have RTCMediaSourceStats.frames.
- framesSent, hugeFramesSent: Moved to RTCOutboundRtpStreamStats.

**RTCAudioHandlerStats**
- voiceActivityFlag: This was moved to RTCOutboundRtpStreamStats and RTCInboundRtpStreamStats. The definition was clarified to be the same as for RTCRtpReceiver.getSynchronizationSources() for the inbound case. For the outbound case, there is no webrtc-pc equivalent way of obtaining this information, so it is defined as the V-bit of the last sent packet.

**RTCAudioSenderStats**
- totalSamplesSent: Moved to RTCOutboundRtpStreamStats.
- echoReturnLoss, echoReturnLossEnhancement: Moved to RTCAudioSourceStats.

This PR moves metrics that that apply to both sending and receiving, and moves all the sending metrics. A follow-up PR will take care of moving receiving-only metrics to RTCInboundRtpStreamStats.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1150,6 +1150,9 @@ enum RTCStatsType {
              DOMString            remoteId;
              unsigned long        framesDecoded;
              unsigned long        keyFramesDecoded;
+             unsigned long        frameWidth;
+             unsigned long        frameHeight;
+             double               framesPerSecond;
              unsigned long long   qpSum;
              double               totalDecodeTime;
              DOMHighResTimeStamp  lastPacketReceivedTimestamp;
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -2751,7 +2778,7 @@ <h3>
         </h3>
         <div>
           <pre class="idl">dictionary RTCAudioHandlerStats : RTCMediaHandlerStats {
-             boolean             voiceActivityFlag;
+              boolean              voiceActivityFlag;
```
vr000m: Do we keep this here or to the RTP stats...

vr000m:
Make clear that the stats are appropriate to the dictionary that they are in.

webrtc-stats.html
henbos: Let's move it to RTP. In that case it should definitely be in RTCInboundRtpStreamStats (reflecting the getSSRCs() which are for receivers), but I question whether it should be in RTCOutboundRtpStreamStats (there is no getSSRCs() on the senders). In this PR, to not introduce any changes other than moving, I moved it to both,  but we nobody has ever implemented sender-side voiceActivityFlag. (Although we haven't implemented the VAD in getSSRCs() either so probably feature at risk regardless)

Issue #468, opened 2019-08-14
New Respec issues


vr000m:
Failed to parse WebIDL: Bodyless dictionary. at: 1.
Details
Failed to parse WebIDL: Bodyless dictionary. at: 1.
Details
The term "boolean" is defined in infra, webidl in multiple ways, so it's ambiguous. See how to cite to resolve the error at: 1, 2.
The term "boolean" is defined in infra, webidl in multiple ways, so it's ambiguous. See how to cite to resolve the error at: 1.
The term "boolean" is defined in infra, webidl in multiple ways, so it's ambiguous. See how to cite to resolve the error at: 1, 2.
The term "record" is defined in html, webidl in multiple ways, so it's ambiguous. See how to cite to resolve the error at: 1, 2, 3.

henbos:
I merged the other PR and get this on tip-of-tree, oops:

😱 ReSpec error: Multiple `<dfn>`s for `framesDecoded` in `RTCInboundRtpStreamStats`
😱 ReSpec error: Multiple `<dfn>`s for `framesDecoded` in `RTCInboundRtpStreamStats`
😱 ReSpec error: Failed to parse WebIDL: Bodyless dictionary. at: [1](#respec-offender-bodyless-dictionary).

<details><pre>Syntax error at line 1 in 40, since `partial dictionary RTCAudioSenderStats`:
partial dictionary RTCAudioSenderStats : RTCAudioHandlerStats {
                                       ^</pre></details>
😱 ReSpec error: Failed to parse WebIDL: Bodyless dictionary. [object HTMLPreElement]
😱 ReSpec error: Failed to parse WebIDL: Bodyless dictionary. at: [1](#respec-offender-bodyless-dictionary-0).

<details><pre>Syntax error at line 1 in 41, since `partial dictionary RTCAudioReceiverStats`:
partial dictionary RTCAudioReceiverStats : RTCAudioHandlerStats {
                                         ^</pre></details>
😱 ReSpec error: Failed to parse WebIDL: Bodyless dictionary. [object HTMLPreElement]
😱 ReSpec error: Failed to parse WebIDL: Bodyless dictionary. at: [1](#respec-offender-bodyless-dictionary-1).

<details><pre>Syntax error at line 1 in 42, since `partial dictionary RTCVideoHandlerStats`:
partial dictionary RTCVideoHandlerStats : RTCMediaHandlerStats {
                                        ^</pre></details>
[...]

PR #466, opened 2019-08-13
Move "receiver" stats to "inbound-rtp"

henbos:
This is a follow-up to #463 that fixes the remaining issues of #402 .

This PR moves all RTCAudioReceiverStats and RTCVideoReceiverStats to RTCInboundRtpStreamStats. This means the following metrics:

**RTCAudioReceiverStats**
- estimatedPlayoutTimestamp
- jitterBufferDelay
- jitterBufferEmittedCount
- totalSamplesReceived
- concealedSamples
- silentConcealedSamples
- concealmentEvents
- insertedSamplesForDeceleration
- removedSamplesForAcceleration
- audioLevel
- totalAudioEnergy
- totalSamplesDuration

**RTCVideoReceiverStats**
- estimatedPlayoutTimestamp
- jitterBufferDelay
- jitterBufferEmittedCount
- framesReceived
- framesDecoded
- framesDropped
- partialFramesLost
- fullFramesLost;

PR #469, opened 2019-08-14
Fix respec issues

henbos:
Fixes some Travis errors introduced by #466 and #463.
See #468 

Removed duplicate framesDecoded and make partial dictionaries not repeat the inheritance


henbos:
I'll just go ahead and merge, Travis passes

PR #473, opened 2019-08-16
Move RTCCodecStats.implementation to RTC[In/Out]boundRtpStreamStats.[decoder/encoder]Implementation

### LINKED ISSUE #445: We should move RTCCodecStats.implementation to RTC[In/Out]boundRtpStreamStats
The set of codec objects is a list based on capabilities and what is negotiated in the SDP.
Which encoder is instantiated for a given codec could theoretically be different for different RTP streams, and is a decision that could be made on-the-fly. When we create codec objects we don't know the implementation. In Chrome, one could start with a HW encoder, but then due to HW encoder errors, it falls back on the fly to a SW implementation.

The current state of the spec is unable to express codecs with multiple implementations, since they are tightly coupled with the SDP, which says nothing about implementation.

The list of codec objects is already huge, I propose we do not duplicate every codec object to have a "implementation: SW" and "implementation: HW" version of the same thing.

To minimize bloat, and to make implementation simpler (at least for Chrome that already has a googCodecImplementation), I propose we move "RTCCodecStats.implementation" to "RTC[In/Out]boundRtpStreamStats.codecImplementation" (or encoderImplementation/decoderImplementation).

This would make shipping "implementation" in Chrome require just a few lines of code, versus refactoring or special logic for the edge-case of implementation changing.

henbos:
Fixes #445 

Which encoder or decoder is used is a property of the RTP stream, not a property of the codec. For example, it is possible that one stream uses an H264 hardware decoder but another stream, having experienced decoding errors and fallen back to software, uses an H264 software decoder. Both RTCInboundRtpStream objects would reference the same RTCCodecStats; the codec stats are tightly coupled with the list if supported/negotiated codecs in the SDP.

This PR allows Chrome to ship this in M78: https://webrtc-review.googlesource.com/c/src/+/149168

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -825,7 +825,6 @@ enum RTCStatsType {
              unsigned long clockRate;
              unsigned long channels;
              DOMString     sdpFmtpLine;
-             DOMString     implementation;
 };</pre>
           <section>
             <h2>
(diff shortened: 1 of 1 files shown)


henbos:
@vr000m Please take a look, Varun.

PR #475, opened 2019-08-19
Add RTCTransportStats.selectedCandidatePairChanges

### LINKED ISSUE #474: Add metric for ICE reroutes
High number of ICE restarts positively correlate with bad network/quality, and would be a useful metric to have.

An ICE reroute is when the ICE candidate pair changes. I propose adding:

> unsigned long RTCTransportStats.selectedCandidatePairChanges:
> The number of times that the selected candidate pair of this transport has changed.

henbos:
Fixes #474

This can happen due to ICE restarts or it can happen spontaneously if continuous renomination (extension spec) is supported. A higher count in selected candidate pair changes (ICE reroutes) may correlate with worse quality.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -3211,6 +3211,7 @@ enum RTCStatsType {
              DOMString             dtlsCipher;
              DOMString             srtpCipher;
              DOMString             tlsGroup;
+             unsigned long         selectedCandidatePairChanges;
 };</pre>
           <section>
             <h2>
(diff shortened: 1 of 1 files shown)


henbos:
Please take a look @vr000m

webrtc-stats.html
```diff
@@ -3362,6 +3363,16 @@ <h2>
...
                 </p>
               </dd>
+              <dt>
+                <dfn><code>selectedCandidatePairChanges</code></dfn> of type <span class=
+                "idlMemberType">unsigned long</span>
+              </dt>
+              <dd>
+                <p>
+                  The number of times that the selected candidate pair of this transport has
+                  changed.
```
alvestrand: Need to specify whether this is 0 or 1 after the initial selection of a candidate pair.

jonasoreland:
the value should be 0 initially, and 1 after the first candidate pair has been selected.

webrtc-stats.html

vr000m:
I believe this is a good metric. However, I have a few clarifications.

Depending on regular or aggressive nomination, a candidate pair can be selected before the checks against all candidates is complete. In the case of aggressive nomination, the selected pair might change, would this metric be updated in those cases?

The difference being that this is measured only after the RTCPeerConnectionState is in "completed" state.

henbos:
If the candidate pair changes the counter goes up, there's no special casing of ignoring changes before being in a completed state, which is why the PR does not include any ifs-or-buts. @vr000m feel free to suggest wording if you think the PR needs to be clarified and I can update the PR per your liking before the meeting.

PR #477, opened 2019-08-27
RTCOutboundRtpStreamStats.qualityLimitationResolutionChanges added.

### LINKED ISSUE #476: Add counter for number of times resolution changes
The spec allows us to see if and why we are resolution limited ([qualityLimitationReason](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-qualitylimitationreason)) as well as how long we have been limited ([qualityLimitationDurations](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-qualitylimitationdurations)). We can also poll for resolution ([frameWidth/frameHeight](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-framewidth)) which gives us a good idea about quality.

However, every time the resolution changes because of being quality limited that will have a noticeable effect to quality for the user. It will also have an impact on the bandwidth usage because it necessarily creates additional key frames. [keyFramesEncoded](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-keyframesencoded) does not tell us why it went up, and estimating resolution changes requires polling getStats a lot. There could be an easier way to measure this.

I propose adding RTCOutboundRtpStreamStats.qualityLimitationResolutionChanges:
> Number of times that the resolution has changed because of we are quality limited (qualityLimitationReason is not "none"). The counter is initially zero, and increases when the resolution goes up or down. For example, a 720p track being sent as 640x480 for some time and then recovers to 720p, qualityLimitationResolutionChanges will have the value 2.

Note that this is a separate discussion than when we discussed number of times the quality limitation reason changed. We are not counting flip-flopping going between "cpu" and "bandwidth" for example; we could be "bandwidth" limited for 100% of the call and the resolution could still vary during the call. This counter is about resolution changes, not reason changes, and complements qualityLimitationDurations nicely without adding additional polling logic to the application.

henbos:
Fixes #476.

Please take a look @vr000m and @alvestrand

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1867,7 +1867,7 @@ enum RTCStatsType {
              double               averageRtcpInterval;
              RTCQualityLimitationReason                 qualityLimitationReason;
              record&lt;DOMString, double&gt; qualityLimitationDurations;
-             unsigned long        qualityLimitationResolutionChanges.
+             unsigned long        qualityLimitationResolutionChanges;
              record&lt;USVString, unsigned long&gt; perDscpPacketsSent;
              unsigned long        nackCount;
              unsigned long        firCount;
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2173,8 +2173,8 @@ enum RTCStatsType {
                   Only valid for video. The number of times that the resolution has changed because
                   we are quality limited (<a>qualityLimitationReason</a> has a value other than
                   <code>"none"</code>). The counter is initially zero and increases when the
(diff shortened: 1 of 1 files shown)


henbos:
FYI @Orphis

henbos:
See motivation in the spec issue

henbos:
Approved by both of you, @alvestrand is on vacation, and there's no meeting this week - I'll merge this then.

PR #482, opened 2019-09-09
Adds RTCOutboundRtpStreamStats.rid

### LINKED ISSUE #395: Add RTCOutboundRtpStreamStats.rid

henbos:
Fixes #395

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1847,6 +1847,7 @@ enum RTCStatsType {
              DOMString            mediaSourceId;
              DOMString            senderId;
              DOMString            remoteId;
+             DOMString            rid;
              DOMHighResTimeStamp  lastPacketSentTimestamp;
              unsigned long long   retransmittedPacketsSent;
              unsigned long long   retransmittedBytesSent;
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1918,6 +1919,17 @@ <h2>
...
                 </p>
               </dd>
+              <dt>
+                <dfn><code>rid</code></dfn> of type <span class="idlMemberType">DOMString</span>
+              </dt>
+              <dd>
+                <p>
+                  Exposes the <a href="https://w3c.github.io/webrtc-pc/#dom-rtcrtpcodingparameters-rid">rid</a>
+                  encoding parameter of this RTP stream if it has been set, otherwise it is
+                  undefined. If set this value will be present regardless if the RID RTP header
```
vr000m: The value will be present even if the RID RTP header extension has not been negotiated.

henbos:
Before merging we need to decide if we want encoding-parameters as a separate object referenced by the outbound-rtp or if we want all per-layer metrics to be part of the RTCOutboundRtpStreamStats

henbos:
This will probably become obsolete by #495

vr000m:
OBE by #495

PR #488, opened 2019-09-16
packet vs report block

### LINKED ISSUE #433: The spec has incorrect assumptions abound SR and RR
The spec seems to assume that RTCRemoteInboundRtpStreamStats map 1:1 with a Receiver Report and RTCRemoteOutboundRtpStreamStats map 1:1 with a Sender Report. This may be true when there is only one stream in one direction, but talking with the network folks this does not seem to be true in general.

For example, when both sending and receiving, the "remote-inbound-rtp" objects correspond to _Report Blocks_, but the blocks could have been carried over either in a Sender Report or a Receiver Report.

The reports are just a list of blocks. Also one report may contain blocks for multiple different streams.

The spec is probably correct if we replace the "Sender Report" and "Receiver Report" language by "Report Block" instead.

vr000m:
fixes #433

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -694,6 +694,12 @@ enum RTCStatsType {
           that make sense in many different contexts are represented just once in IDL.
         </p>
         <p>
+          The metrics exposed here correspond to local measurements and those reported by RTCP packets. 
+          Compound RTCP packets contain multiple RTCP report blocks, such as Sender Report (SR) and 
+          Receiver Report (RR). Wheras, a non-compound RTCP packets may contain just a single 
+          RTCP SR or RR block.
+        </p>
+        <p>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -693,6 +693,12 @@ <h3>
           The dictionaries for RTP statistics are structured as a hierarchy, so that those stats
           that make sense in many different contexts are represented just once in IDL.
         </p>
+        <p>
+          The metrics exposed here correspond to local measurements and those reported by RTCP packets. 
+          Compound RTCP packets contain multiple RTCP report blocks, such as Sender Report (SR) and 
+          Receiver Report (RR). Wheras, a non-compound RTCP packets may contain just a single
```
fippo: grammar nit: s/wheras/whereas/

Issue #458, opened 2019-07-02
Clarify period for framesDropped / FramesLost


stefhak:
The spec text talks about `total number of frames` and `cumulative number of partial/full frames`; total/cumulative hints measuring period would be 'since receiver was created', however RFC7004 appendix A is referenced, and RFC7004 references RFC3611 for `begin_seq` and `end_seq`.

I think this should be clarified.

vr000m:
We should reference Sections A.g, h, i, j..

Basically,
```
 Number of duplicate frames (dup_frames): 32 bits

      Number of duplicate frames received in the above sequence number
      interval.

   Number of full lost frames (full_lost_frames): 32 bits

      A frame is either split across multiple packets or carried in only
      one packet.  If the whole frame or all the packets of the frame
      are lost, this frame is regarded as one full_lost_frame.  The
      full_lost_frames can be inferred from packet(s) that comprise the
      frame.  The full_lost_frames is equivalent to the number of full
      lost frames in the above sequence number interval.

   Number of partial lost frames (partial_lost_frames): 32 bits

      When a frame is split across multiple packets and some packets of
      the frame are lost, this frame is regarded as one
      partial_lost_frame.  The partial_lost_frames can be inferred from
      packets that comprise the frame.  The value of the
      partial_lost_frames field is equivalent to the number of partial
      lost frames in the above sequence number interval.
``

**Comment by vr000m** [comment]
In RTP the frames can be identified from the fragemented packets.
All packets have the same RTP Timestamp, they should be contiguous sequence numbers, the last packet in the frame has the market bit set. 

IIRC, the begin_seq and end_seq in RFC7004 are more important for the other metrics being reported in the packet (namely burst/gap metrics).

**Comment by stefhak** [comment]
I did parse the definitions, but I find them hard to use since I don't know what `the above sequence number interval` is. Basically I had expected x_lost_frames to be stable or increasing over a session but was surprised to find out that they can go up and down (e.g. getting a sequence like 1, 0, 1, 0 , 2, ... when sampling every n seconds) during a session when I did a small test.

**Comment by vr000m** [comment]
I agree that we should make clear that this is not an interval metric but the start from the appearance of the Receiver.

@stefhak: For the XR in RFC7004 what was the setting for the I (interval flag).

```
This field is used to indicate whether the burst/gap loss summary
      statistics metrics are Sampled, Interval, or Cumulative metrics:

         I=10: Interval Duration - the reported value applies to the
[...]

PR #490, opened 2019-09-16
Clarify Cumulative metric for XR blocks and not interval and move to right dictionary

vr000m:
If these are reported in XR blocks, these metrics should be in RTCReceivedRtpStreamStats and not in RTCInboundRtpStreamStats


vr000m:
Fixes #458

PR #491, opened 2019-09-16
adding totalRoundTripTime, reportsReceived, reportsSent

### LINKED ISSUE #426: RTCRemoteInboundRtpStreamStats: total round trip time
[RTCRemoteInboundRtpStreamStats.roundTripTime](https://w3c.github.io/webrtc-stats/#dom-rtcremoteinboundrtpstreamstats-roundtriptime) is only the [latest estimate](https://tools.ietf.org/html/rfc3550#section-6.4.1) (scroll down to "delay since last SR (DLSR): 32 bits").

To allow calculating averages we should do the same thing that we did with regards to RTCIceCandidatePairStats - have a `totalRoundTripTime` and a counter that gets up every time you calculate a sample, i.e. every time you get a Receiver Report... or more accurately, every time you get a Report Block; `reportsReceived`.

vr000m:
Fixes #426

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1714,6 +1714,7 @@ enum RTCStatsType {
              double               totalRoundTripTime;
              double               fractionLost;
              unsigned long long   reportsReceived;
+             unsigned long long   roundTripTimeMeasurements;
 };</pre>
           <section>
             <h2>
@@ -1753,7 +1754,8 @@ enum RTCStatsType {
                   since the beginning of the session. The individual round trip time is calculated 
                   based on the RTCP timestamps in the RTCP Receiver Report (RR) [[!RFC3550]], hence
                   undefined roundtrip are not added. The average round trip time can be computed 
-                  from <code>totalRoundTripTime</code> by  dividing it by <code>reportsReceived</code>.
+                  from <code>totalRoundTripTime</code> by  dividing it by 
+                  <code>roundTripTimeMeasurements</code>.
                 </p>
               </dd>
               <dt>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1741,6 +1743,19 @@ <h2>
...
+                <dfn><code>totalRoundTripTime</code></dfn> of type <span class=
+                "idlMemberType">double</span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the cumulative sum of all round trip time measurements in seconds 
+                  since the beginning of the session. The individual round trip time is calculated 
+                  based on the RTCP timestamps in the RTCP Receiver Report (RR) [[!RFC3550]], hence
+                  undefined roundtrip are not added. The average round trip time can be computed 
+                  from <code>totalRoundTripTime</code> by  dividing it by <code>reportsReceived</code>.
```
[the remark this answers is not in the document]
henbos: I spoke too soon: You've defined sent/received as "contain a valid round trip time", so my question now is, does it make sense for a metric called "reportsSent/reportsReceived" to not increase if we get a report because something about that report does not fulfill a criteria? Should we rename the metric or have multiple metrics?

Why would timestamp not be present?

webrtc-stats.html
```diff
@@ -1709,9 +1709,11 @@ <h3>
...
         <div>
           <pre class="idl">dictionary RTCRemoteInboundRtpStreamStats : RTCReceivedRtpStreamStats {
-             DOMString          localId;
-             double             roundTripTime;
-             double             fractionLost;
+             DOMString            localId;
+             double               roundTripTime;
+             double               totalRoundTripTime;
+             double               fractionLost;
+             unsigned long long   reportsReceived;
```
henbos: Let's call it roundTripTimeMeasurements (or something like that)

webrtc-stats.html
```diff
@@ -2331,6 +2356,7 @@ <h3>
           <pre class="idl">dictionary RTCRemoteOutboundRtpStreamStats : RTCSentRtpStreamStats {
              DOMString           localId;
              DOMHighResTimeStamp remoteTimestamp;
+             unsigned long long  reportsSent;
```
henbos: remoteTimestampsSent

webrtc-stats.html
```diff
@@ -2359,10 +2385,20 @@ <h2>
...
                 </p>
               </dd>
+              <dt>
+                <dfn><code>reportsSent</code></dfn> of type <span class=
+                "idlMemberType">unsigned long long</span>
+              </dt>
+              <dd>
+                <p>
+                  Represents the total number of RTCP SR blocks sent for this SSRC that contain 
+                  a valid remoteTimestamp.
```
vr000m: remove the caveat.

PR #492, opened 2019-09-16
SILK and CELT mode sample counters

### LINKED ISSUE #432: OPUS -- Exposing SILK or CELT mode in getStats
We've found that Opus performance can vary depending on what mode it is in. Especially, when it transitions from one mode to another, and it is currently impossible to know when that transition occurs or what mode the encoder is in to begin with.

I am not sure how this information could be exposed in getStats. Any thoughts or proposals?
For example -- counters for each mode, and start-mode and current-mode might make sense.

vr000m:
Fixes #432

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1157,6 +1157,8 @@ enum RTCStatsType {
              double               jitterBufferDelay;
              unsigned long long   jitterBufferEmittedCount;
              unsigned long long   totalSamplesReceived;
+             unsigned long long   samplesDecodedWithSilk;
+             unsigned long long   samplesDecodedWithCelt;
              unsigned long long   concealedSamples;
              unsigned long long   silentConcealedSamples;
              unsigned long long   concealmentEvents;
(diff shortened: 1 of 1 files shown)


aboba:
@vr00m It's not just Silk or CELT; Opus has hybrid modes as well:
https://wiki.hydrogenaud.io/index.php?title=Opus

This makes me wonder whether we shouldn't look at target bitrates instead.

vr000m:
If I remember correctly, if the bitrate is going up or down the OPUS may chose SILK, CELT MODE. That was one of the reasons for the original request, ie not have to look up the table because the implementation might be doing something different.

What is the hybrid mode in Opus?

Founder, CEO, callstats.io ( http://callstats.io )
http://www.callstats.io

Interested in networking, media quality, and diagnostics.
We are hiring!: www.callstats.io/jobs/ ( http://www.callstats.io/jobs/ )

henbos:
To be discussed at TPAC

vr000m:
@aboba any thoughts on this.

henbos:
Needs input from implementors: "What is the hybrid mode?"

PR #494, opened 2019-09-16
making sure that unsigned long long is used for packet counters

### LINKED ISSUE #425: Byte/packet counter type consistency (retransmittedPacketsSent is unsigned long long)
Generally speaking, currently in the spec byte counters are "unsigned long long" but packet counters are "unsigned long". Should we continue this pattern or should we use "unsigned long long" for everything?

A recent PR added both retransmittedBytesSent and retransmittedPacketsSent. Both were added as "unsigned long long". Should retransmitedPacketsSent be "unsigned long"?

karthikbr82:
Fixes #425

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -952,19 +952,19 @@ enum RTCStatsType {
         </h3>
         <div>
           <pre class="idl">dictionary RTCReceivedRtpStreamStats : RTCRtpStreamStats {
-             unsigned long      packetsReceived;
-             long               packetsLost;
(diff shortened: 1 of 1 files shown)


henbos:
@vr000m  @karthikbr82 TODO to look up if there are problems with overflowing. The local side should always report the right metric, but on the receiver side it should be either undefined...?

henbos:
Or create a separate issue for the wrap around problems

vr000m:
Fixes issue #425

PR #497, opened 2019-09-21
Remove all references to onstatsended

### LINKED ISSUE #472: Issues with replaceTrack, will statsended fire or give me what I want.
With all the changes to track attachment stats, we are close to deprecating it. May be we should not. 

My use-case of replaceTrack switching from camera to screenshare, this does not fire an event on the peerconnection. We kind of solved this issue with `onstatsended` event returning the stats of track that was being replaced. however, with the recent changes, if we get rid of the empty dictionaries there will be no stats for the track that got replaced, that means the onstatsended will fire but return an empty dictionary. We would then need to call getStats() and get the latest stats from there, albeit with a bit of delay.

### LINKED ISSUE #480: What should we do about onstatsended?
Now that there is little point to having track/sender/receiver stats dictionaries around because all their members are made obsolete (#479), it seems that the only use case for them are having objects that would trigger `onstatsended`.

So what is the use case for `onstatsended`? Well, it's nice to know if replaceTrack() was called so that you can correlate stats changing with that point in time.

I think we can do better and not mandate people implement track attachment stats anymore. So what do we do about `onstatsended`?

henbos:
Fixes #480.
Fixes #472.

Also need to create a webrtc-pc PR.


webrtc-stats.html
```diff
@@ -4346,7 +4330,10 @@ <h2>
             </dt>
             <dd>
               <p>
-                This field is obsoleted by the <a>statsended</a> event.
+                This field was obsoleted by the <a>statsended</a> event, which
+                has now also been obsolete. <code>deleted</code> is no longer
+                applicable beacause if tthe ice candidate is deleted it no
```
vr000m: double -- "tt"

vr000m:
synchronize fate with webrtc-pc issue

PR #498, opened 2019-09-21
RTCRtpTransceiverStats added.

### LINKED ISSUE #396: RTC[Audio/Video]SenderStats should have mid

### LINKED ISSUE #478: Should we have transceiver stats?
We've had `sender` and `receiver` stats but no `transceiver` stats (and now almost everything has been moved to to the `outbound-rtp` and `inbound-rtp`).

Should we have `transceiver` stats?

Possible uses:
- `mid` is a property of the `RTCRtpTransceiver`
- Codecs that have been negotiated are on a per-m= section, i.e. per transceiver basis. `RTCCodecStats` could reference transceivers.

Both of these use cases would work by using `sender` and `receiver` instead of a `transceiver` though.

henbos:
Fixes #478.
Fixes #396 (by making it obsolete).

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -3046,7 +3046,9 @@ enum RTCStatsType {
               <dd>
                 <p>
                   If the RTCRtpTransceiver that this stats object represents has
-                  a <code>mid</code> value that is not null, this is that value,
+                  a <code><a href=
+                  "https://w3c.github.io/webrtc-pc/#dom-rtptransceiver-mid">
+                  mid</a></code> value that is not null, this is that value,
                   otherwise this value is undefined.
                 </p>
               </dd>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -2884,6 +2894,70 @@ <h2>
...
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>mid</code></dfn> of type <span class=
+                "idlMemberType">DOMString</span>
+              </dt>
+              <dd>
+                <p>
+                  If the RTCRtpTransceiver that this stats object represents has
+                  a <code>mid</code> value that is not null, this is that value,
```
vr000m: should we say that the mid is extracted from the SDP?

henbos: I'll point to [[WEBRTC-PC]]

PR #499, opened 2019-09-21
Make RTCMediaStreamStats obsolete

### LINKED ISSUE #470: RTCMediaStreamStats can be made obsolete
Now that things have been moved to "outbound-rtp" and "inbound-rtp", "track" stats makes less sense, and "stream" stats less so since its a list of trackIds.

We could replace the dictionary by streamIdentifiers on the "sender" and "receiver"

henbos:
Fixes #470


vr000m:
@karthikbr82 Please take a look -- reviewe

henbos:
@karthikbr82 "track" stats are going away, this dictionary is not providing any useful information. One of the proposals at TPAC was to add streamIdentifiers to "sender" and "receiver" stats to make up for it (in a way that would also model the RTP Media API), but opinions were expressed not to add this to getStats() unless we have good reason. We already have made a TPAC decision to remove this dictionary so if you want to review or make a counter argument then speak up

vr000m:
@pallab-gain @karthikbr82 could you please take a look and respond by end of this week, so that we can land this merge.

PR #501, opened 2019-09-21
Added note about channels for all sample based metrics

### LINKED ISSUE #448: Audio samples and channels
totalAudioEnergy is based on totalSamplesSent or totalSamplesReceived, as is totalSamplesDurations.
These are all based on number of samples.

The spec never mentions audio channels though. The current implementation in Chrome calculates all of these things without taking the number of channels in to account. For example, assuming no packet loss, the totalSamplesDuration increase by 1 second per second. If we took in to account the fact that the audio has 2 channels, we would expect the true samples count to be twice as high, and as a consequence that the totalSamplesDurations increased by 2 seconds per second.

Is Chrome doing the right thing or the wrong thing?
Should we update the spec to say "samples per channel" everywhere it currently says "samples"?

henbos:
Fixes #448.

Alternatively we could avoid copy-pasting the same paragraph all over the places by defining this once in the beginning of the document, perhaps that is preferred, what do you think?

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1519,6 +1519,12 @@ enum RTCStatsType {
                   delay can be calculated by dividing the <code>jitterBufferDelay</code> with the
                   <code>jitterBufferEmittedCount</code>.
                 </p>
+                <p>
+                  Note: If multiple audio channels are used, metrics based on
+                  samples do not increment at a higher rate. "A sample" refers
+                  to having a sample in either channel - simultaneously having
+                  samples in multiple channels counts as a single sample.
+                </p>
               </dd>
               <dt>
                 <dfn><code>jitterBufferEmittedCount</code></dfn> of type <span class=
@@ -1529,6 +1535,12 @@ enum RTCStatsType {
                   The total number of audio samples or video frames that have come out of the
                   jitter buffer (increasing <code>jitterBufferDelay</code>).
                 </p>
+                <p>
+                  Note: If multiple audio channels are used, metrics based on
+                  samples do not increment at a higher rate. "A sample" refers
+                  to having a sample in either channel - simultaneously having
+                  samples in multiple channels counts as a single sample.
+                </p>
               </dd>
               <dt>
                 <dfn><code>totalSamplesReceived</code></dfn> of type <span class=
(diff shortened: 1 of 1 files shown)


alvestrand:
Repeating a paragraph is usually a Bad Idea.
I'd suggest putting it in when we first introduce the concept of "audio" stats.

webrtc-stats.html
```diff
@@ -2607,6 +2689,14 @@ <h2>
                   echo cancellation is applied. Calculated in decibels, as defined in [[!ECHO]]
                   (2012) section 3.14.
                 </p>
+                <p>
+                  Note: If multiple audio channels are used, metrics based on
+                  samples do not increment at a higher rate. "A sample" refers
+                  to having a sample in either channel - simultaneously having
+                  samples in multiple channels counts as a single sample. When
+                  calculating echo return loss, the channel of the least audio
+                  energy is considered for any sample.
```
vr000m: confirming if "least audio energy" was intended, we should explain what is the consequence or why we picked least. I am assuming this is not documented in any other standards.

webrtc-stats.html
```diff
@@ -2618,6 +2708,14 @@ <h2>
                   echo cancellation is applied. Calculated in decibels, as defined in [[!ECHO]]
                   (2012) section 3.15.
                 </p>
+                <p>
+                  Note: If multiple audio channels are used, metrics based on
+                  samples do not increment at a higher rate. "A sample" refers
+                  to having a sample in either channel - simultaneously having
+                  samples in multiple channels counts as a single sample. When
+                  calculating echo return loss enhancement, the channel of the
+                  least audio energy is considered for any sample.
```
vr000m: confirming if "least audio energy" was intended, we should explain what is the consequence or why we picked least. I am assuming this is not documented in any other standards.

webrtc-stats.html
```diff
@@ -2580,6 +2648,14 @@ <h2>
                   <code>Math.sqrt(0.0026/0.02) = 0.36</code>, which is the same value that would be
                   obtained by doing an RMS calculation over the contiguous 20ms segment of audio.
                 </p>
+                <p>
+                  Note: If multiple audio channels are used, metrics based on
+                  samples do not increment at a higher rate. "A sample" refers
+                  to having a sample in either channel - simultaneously having
+                  samples in multiple channels counts as a single sample. The
+                  "audio energy of a sample" refers to the highest energy of any
+                  channel.
```
vr000m: confirming if "highest audio energy" was intended, we should explain what is the consequence or why we picked the highest. I am assuming this is not documented in any other standards.

webrtc-stats.html
```diff
@@ -1660,6 +1708,14 @@ <h2>
...
```
vr000m: confirming if "highest audio energy" was intended, we should explain what is the consequence or why we picked the highest. I am assuming this is not documented in any other standards.

webrtc-stats.html
henbos: Correction: The greatest ERL value

webrtc-stats.html
henbos: Correction: The greatest ERLE value

PR #503, opened 2019-09-22
RTCSctpTransportStats with smoothedRoundTripTime added.

### LINKED ISSUE #376: [DataChannels] Use RTT from sctp in the stats

henbos:
Fixes #376.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -448,6 +448,7 @@ enum RTCStatsType {
 "sender",
 "receiver",
 "transport",
+"sctp-transport",
 "candidate-pair",
 "local-candidate",
 "remote-candidate",
@@ -613,6 +614,16 @@ enum RTCStatsType {
             <p>
               Transport statistics related to the <code>RTCPeerConnection</code> object. It is
               accessed by the <code><a>RTCTransportStats</a></code>.
+            </p>
+          </dd>
+          <dt>
+            <dfn><code>sctp-transport</code></dfn>
+          </dt>
+          <dd>
+            <p>
+              SCTP transport statistics related to an
+              <code>RTCSctpTransport</code> object. It is accessed by the
+              <code><a>RTCSctpTransportStats</a></code> dictionary.
             </p>
           </dd>
           <dt>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -3463,6 +3474,40 @@ <h2>
...
+            <dl data-link-for="RTCSctpTransportStats" data-dfn-for="RTCSctpTransportStats" class=
+            "dictionary-members">
+              <dt>
+                <dfn><code>smoothedRoundTripTime</code></dfn> of type
+                <span class="idlMemberType">double</span>
+              </dt>
+              <dd>
+                <p>
+                  The latest smoothed round-trip time value, corresponding to
+                  spinfo_srtt defined in [[RFC6458]] but converted to seconds.
```
henbos: @vr000m When you review can you take a look at the relevant RFCs if there is something else that we could reference from the public SCTP API?

alvestrand: https://tools.ietf.org/html/rfc4960#section-6.3.1 gives the calculation of SRTT.

alvestrand:
Is there implementor interest to add this?
If someone's committed to implement it, we should merge it.

henbos:
@shacharz and @lgrahl can you review and comment?

webrtc-stats.html
vr000m: This will report the instantaneous value of SRTT when getStats is called.

lgrahl:
I don't really understand why exactly SRTT is being chosen and none of the other stats but I don't mind.

> Is there implementor interest to add this?

That would be my question, too. /cc @steely-glint

alvestrand:
As to "why SRTT": Bug #376 was specifically about SRTT, and discussion of "what is useful to expose" has mostly been saying "RTT is useful". More stats have costs (slower getStats calls, more code to maintain), so we'd like to make sure there's a benefit to adding a stat before we add it. Separate bugs are good.

vr000m:
Since there was interest, we will go ahead an merge this. We do not foresee implementation issues with this. If there may be a need for more guidance, we can open a new ticket.

webrtc-stats.html
shacharz: @henbos 
sorry missed the mention.
converted to seconds means it'll be a fracture?
as in `25`ms will be converted to `0.025`s ?

PR #504, opened 2019-09-22
Clarify qualityLimitationReason

### LINKED ISSUE #440: Clarify qualityLimitationReason when limited by multiple reasons
qualityLimitationReason can be "cpu" or "bandwidth", not both. We should clarify what to do in this case, such as to prioritize "bandwidth" over "cpu".

henbos:
Fixes #440

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2206,6 +2206,16 @@ enum RTCStatsType {
                   Only valid for video. The current reason for limiting the resolution and/or
                   framerate, or "none" if not limited.
                 </p>
+                <p>
+                  If there are multiple limitation reasons at the same time, the
+                  implementation is encouraged to attempt to report the most
+                  limiting reason. If the implementation does not know how to
+                  determine this it SHOULD prefer to report reasons in the
+                  following order of priority: "bandwidth", "cpu", "other". In
+                  other words, an implementation that is able to tell we are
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -2206,6 +2206,16 @@ <h2>
...
                 </p>
+                <p>
+                  If there are multiple limitation reasons at the same time, the
+                  implementation is encouraged to attempt to report the most
+                  limiting reason. If the implementation does not know how to
+                  determine this it SHOULD prefer to report reasons in the
+                  following order of priority: "bandwidth", "cpu", "other". In
+                  other words, an implementation that is able to tell we are
+                  both "bandwidth" and "cpu" limited but cannot tell which
+                  factor is the most limiting factor will report "bandwidth".
```
vr000m: First thing, SHOULD always follow when not to apply the preference order. In this case the sentence is trying to say MUST use this order if the endpoint does not know.

Secondly, we are not explaining the rationale why this preference order is chosen as the default.

henbos: I rephrased things. But because what the "most limiting factor" is, even under ideal circumstances: a guestimate - I couldn't faithfully include the word "MUST". I used "SHOULD" in both places. The changes are uploaded in a new PR: #526. Closing this one.

PR #506, opened 2019-10-02
RTC[Out/In]boundRtpStreamStats.headerBytes[Sent/Received] added.

### LINKED ISSUE #505: Metrics for header+padding bytes sent and received
bytesSent and bytesReceived is payload only.
A bug in Chrome is to implement them as payload + header and padding. It would be useful to be able to get both pieces of information (total network throughput), and having both would make it easier to roll out the fix in Chrome to update its counters to be payload bytes only, as that would allow applications relying on them to calculate the total themselves and not see a dip when the fix is rolled out.

henbos:
Fixes #505


webrtc-stats.html
```diff
@@ -1368,6 +1369,15 @@ <h2>
                   CNAME item.
                 </p>
               </dd>
+              <dt>
+                <dfn><code>headerBytesReceived</code></dfn> of type <span class=
+                "idlMemberType">unsigned long</span>
+              </dt>
+              <dd>
+                <p>
+                  Total number of RTP header and padding bytes received for this SSRC.
```
henbos: Clarify that this excludes payload and does not include IP, UDP, etc headers

webrtc-stats.html
```diff
@@ -1940,6 +1951,15 @@ <h2>
                   statistics were generated by the local endpoint.
                 </p>
               </dd>
+              <dt>
+                <dfn><code>headerBytesSent</code></dfn> of type <span class=
+                "idlMemberType">unsigned long</span>
+              </dt>
+              <dd>
+                <p>
+                  Total number of RTP header and padding bytes sent for this SSRC.
```
alvestrand: Want to add a comment saying that this + bytesSent = the number of bytes sent as UDP payload (that is, this doesn't include UDP headers or IP headers)

vr000m: There could be another alternative: totalBytesSent, and to get headers and padding you'd subtract bytesSent (which is payload).

I prefer the headerBytes, though.

henbos: Suggestion: "Total number of RTP header and padding bytes sent for this SSRC. This does not include the size of transport layer headers such as IP or UDP. headerBytesSent + bytesSent equals the number of bytes sent as payload over the transport."

PR #511, opened 2019-10-17
Add totalInterFrameDelaySquared to RTCInboundRtpStreamStats

### LINKED ISSUE #443: Detecting Video Playback glitches
Chrome's legacy stats has "googInterframeDelayMax".
This can help detect if any time during the call there was a large delay between frames such as a glitch.

The standard API has stats useful for detecting long freezes, such as by looking at framesPerSecond, framesReceived and lastPacketReceivedTimestamp.

But if a minor freeze/glitch occurs of say 100 ms this might not be noticed by an application that polls getStats() every one or several seconds.

Using framesReceived on a per-second basis, a 100 ms loss of frames would certainly be noticeable, but it would look the same as if the frame rate was slightly reduced; not as bad for the user experience as a glitch.

Should we add a new metric for this? Can it be done without a "max" metric? (Problem with "max" is it says nothing about recent stuff, it speaks of the entire history of the call.)

drkron:
Fix #443

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1188,7 +1188,7 @@ enum RTCStatsType {
              double               framesPerSecond;
              unsigned long long   qpSum;
              double               totalDecodeTime;
-             unsigned long long   totalInterFrameDelaySquared;
+             double               totalInterFrameDelaySquared;
              boolean              voiceActivityFlag;
              DOMHighResTimeStamp  lastPacketReceivedTimestamp;
              double               averageRtcpInterval;
@@ -1341,13 +1341,14 @@ enum RTCStatsType {
               </dd>
               <dt>
                 <dfn><code>totalInterFrameDelaySquared</code></dfn> of type <span class=
-                "idlMemberType">unsigned long long</span>
+                "idlMemberType">double</span>
               </dt>
               <dd>
                 <p>
-                  Sum of the squared interframe delay in ms between consecutive frames. The
-                  interframe variance can be calculated from this value and <a>framesDecoded</a>
-                  at two timestamps <code>t0</code> and <code>t1</code>.
+                  Sum of the squared interframe delay in seconds between consecutive frames. The
+                  interframe variance can be calculated from <a>totalInterFrameDelaySquared</a>,
+                  <a>framesDecoded</a>, and <a>Timestamp</a> by reading out these values at two
+                  timestamps <code>t0</code> and <code>t1</code>.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1348,10 +1348,12 @@ enum RTCStatsType {
                   Sum of the squared interframe delay in seconds between consecutive decoded frames,
                   recorded just after a frame has been decoded. The interframe variance can be
                   approximately calculated from <a>totalInterFrameDelaySquared</a>,
-                  <a>framesDecoded</a>, and <a>timestamp</a> by reading out these values at two
-                  timestamps <code>t0</code> and <code>t1</code>. This will be an approximation
-                  because <code>t0</code> and <code>t1</code> are timestamps of when the statistics
-                  are collected, which is only an approximation of when the last frame was decoded.
+                  <a>framesDecoded</a>, and
+                  <a href="https://www.w3.org/TR/webrtc/#dom-rtcstats-timestamp">RTCStats.timestamp</a>
+                  by reading out these values at two timestamps <code>t0</code> and <code>t1</code>.
+                  This will be an approximation because <code>t0</code> and <code>t1</code> are
+                  timestamps of when the statistics are collected, which is only an approximation of
+                  when the last frame was decoded.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1188,7 +1188,8 @@ enum RTCStatsType {
              double               framesPerSecond;
              unsigned long long   qpSum;
              double               totalDecodeTime;
-             double               totalInterFrameDelaySquared;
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1338,6 +1339,17 @@ <h2>
...
               </dd>
+              <dt>
+                <dfn><code>totalInterFrameDelaySquared</code></dfn> of type <span class=
+                "idlMemberType">unsigned long long</span>
+              </dt>
+              <dd>
+                <p>
+                  Sum of the squared interframe delay in ms between consecutive frames. The
+                  interframe variance can be calculated from this value and <a>framesDecoded</a>
+                  at two timestamps <code>t0</code> and <code>t1</code>.
```
henbos: Can you clarify that t0 and t1 are measured at the same time that framesDecoded goes up, and that the measurements are in seconds?

drkron: Is this better?

webrtc-stats.html
```diff
@@ -1341,13 +1341,14 @@ <h2>
...
+                "idlMemberType">double</span>
               </dt>
               <dd>
                 <p>
-                  Sum of the squared interframe delay in ms between consecutive frames. The
-                  interframe variance can be calculated from this value and <a>framesDecoded</a>
-                  at two timestamps <code>t0</code> and <code>t1</code>.
+                  Sum of the squared interframe delay in seconds between consecutive frames. The
+                  interframe variance can be calculated from <a>totalInterFrameDelaySquared</a>,
+                  <a>framesDecoded</a>, and <a>Timestamp</a> by reading out these values at two
```
henbos: nit: "timestamp", not "Timestamp"

webrtc-stats.html
```diff
@@ -1341,13 +1341,14 @@ <h2>
...
                 <dfn><code>totalInterFrameDelaySquared</code></dfn> of type <span class=
-                "idlMemberType">unsigned long long</span>
+                "idlMemberType">double</span>
               </dt>
               <dd>
                 <p>
-                  Sum of the squared interframe delay in ms between consecutive frames. The
-                  interframe variance can be calculated from this value and <a>framesDecoded</a>
-                  at two timestamps <code>t0</code> and <code>t1</code>.
+                  Sum of the squared interframe delay in seconds between consecutive frames. The
```
henbos: When you say "between two frames", what time are you referring to with regards to a frame? Make it clear that "a frame" means when a frame is decoded, i.e. at the same time that framesDecoded increments.

I.e. if the measurement was performed just _before_ decoding then the calculations dividing by framesDecoded would not work.

webrtc-stats.html
drkron: Thanks, I was fooled by my test script that capitalized Timestamp and a few other fields.

webrtc-stats.html
drkron: Good point, will fix.

drkron: The more I think of it, it's a bit tricky to do the conversion to standard deviation since the timestamp of when the last frame was decoded is not available.
I made an attempt to include the text and still be correct about it being an approximation.

webrtc-stats.html
```diff
@@ -1338,6 +1339,23 @@ <h2>
...
+                <p>
+                  Sum of the squared interframe delay in seconds between consecutive decoded frames,
+                  recorded just after a frame has been decoded. The interframe variance can be
+                  approximately calculated from <a>totalInterFrameDelaySquared</a>,
+                  <a>framesDecoded</a>, and
+                  <a href="https://www.w3.org/TR/webrtc/#dom-rtcstats-timestamp">RTCStats.timestamp</a>
+                  by reading out these values at two timestamps <code>t0</code> and <code>t1</code>.
+                  This will be an approximation because <code>t0</code> and <code>t1</code> are
+                  timestamps of when the statistics are collected, which is only an approximation of
+                  when the last frame was decoded.
```
henbos: I'm having trouble understanding if t0 and t1 need to be approximations of then the last frame was decoded, or if this is just a comment about the fact that we are (as always) working with averages. But I'll let other reviewers decide if this sentence is good or not. I'm happy with this PR now. Approved.

drkron: One alternative would be to change _calculated_ to _estimated_ and remove the text about approximation.

webrtc-stats.html
vr000m: What you are computing is: (currentInterFrameDelay - averageInterFrameDelay)^2  or (currentInterFrameDelay)^2

henbos: Are we trying to use this forumula? https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Na%C3%AFve_algorithm

If so...
1: Can you say which formula we are expecting to use?
2: If the missing term is (sum of interframe delays)^2, and we are estimating this based on (timestampSecond-timestampFirst)^2, but that this breaks down when you stop receiving frames - how about we add a second metric which is totalInterframeDelay? That way we have everything :))))))))

vr000m: I think we are then going for -- (totalInterFrameDelaySquared - totalInterframeDelay^2/framesDecoded)/framesDecoded

drkron: That's a great idea to also add totalInterframeDelay, that removes all approximations on the way to get the variance and also makes the calculation more self-contained.

drkron: Done, see the latest commit.

webrtc-stats.html
drkron: Adding totalInterframeDelay has made the complicated text unnecessary.

webrtc-stats.html
drkron: Not needed anymore, see the latest commit.

PR #517, opened 2019-11-08
Add frameBitDepth to RTCInboundRtpStreamStats and RTCOutboundRtpStreamStats

### LINKED ISSUE #516: Add frameBitDepth to RTCInboundRtpStreamStats and RTCOutbountRtpStreamStats
Add the video frame bit depth to RTCInboundRtpStreamStats and RTCOutbountRtpStreamStats.

This is useful in the same way as frameWidth and frameHeight to quickly determine the characteristics of the incoming/outcoming video stream.

drkron:
Fix #516

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1185,6 +1185,7 @@ enum RTCStatsType {
              unsigned long        keyFramesDecoded;
              unsigned long        frameWidth;
              unsigned long        frameHeight;
+             unsigned long        frameBitDepth;
              double               framesPerSecond;
              unsigned long long   qpSum;
              double               totalDecodeTime;
@@ -1296,6 +1297,17 @@ enum RTCStatsType {
                 <p>
                   Only valid for video. Represents the height of the last decoded frame. Before
                   the first frame is decoded this attribute is missing.
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>frameBitDepth</code></dfn> of type <span class="idlMemberType">unsigned
+                long</span>
+              </dt>
+              <dd>
+                <p>
+                  Only valid for video. Represents bit depth per channel of the raw video frame.
+                  Typical values are 8, 10, or 12 bits.
+                  Before the first frame is decoded this attribute is missing.
                 </p>
               </dd>
               <dt>
@@ -1931,6 +1943,7 @@ enum RTCStatsType {
              unsigned long long   totalEncodedBytesTarget;
              unsigned long        frameWidth;
              unsigned long        frameHeight;
+             unsigned long        frameBitDepth;
              double               framesPerSecond;
              unsigned long        framesSent;
              unsigned long        hugeFramesSent;
@@ -2131,6 +2144,17 @@ enum RTCStatsType {
                 <p>
                   Only valid for video. Represents the height of the last encoded frame. The resolution
                   of the encoded frame may be lower than the media source (see RTCVideoSourceStats.height).
+                  Before the first frame is encoded this attribute is missing.
+                </p>
+              </dd>
+              <dt>
+                <dfn><code>frameBitDepth</code></dfn> of type <span class="idlMemberType">unsigned
+                long</span>
+              </dt>
+              <dd>
+                <p>
+                  Only valid for video. Represents bit depth per channel of the raw video frame.
+                  Typical values are 8, 10, or 12 bits.
                   Before the first frame is encoded this attribute is missing.
                 </p>
               </dd>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1305,8 +1305,8 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Only valid for video. Represents bit depth per channel of the raw video frame.
-                  Typical values are 8, 10, or 12 bits.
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1298,6 +1299,17 @@ <h2>
...
               </dd>
+              <dt>
+                <dfn><code>frameBitDepth</code></dfn> of type <span class="idlMemberType">unsigned
+                long</span>
+              </dt>
+              <dd>
+                <p>
+                  Only valid for video. Represents bit depth per channel of the raw video frame.
+                  Typical values are 8, 10, or 12 bits.
+                  Before the first frame is decoded this attribute is missing.
```
henbos: What's the benefit of "per channel" versus "per pixel"? Maybe same same but different but it makes me wonder, what if the format is no traditional 3 channel of same depth

drkron: Yes, that's probably better. I've just pushed a change to per pixel.

webrtc-stats.html
```diff
@@ -1298,6 +1299,17 @@ <h2>
                   the first frame is decoded this attribute is missing.
                 </p>
               </dd>
+              <dt>
+                <dfn><code>frameBitDepth</code></dfn> of type <span class="idlMemberType">unsigned
+                long</span>
+              </dt>
+              <dd>
+                <p>
+                  Only valid for video. Represents the bit depth per pixel of the raw video frame.
```
vr000m: Change this to of the encoded frame instead of raw. And add another metric on the media-source for the raw bit depth.

vr000m: The idea would be if 888 is the raw, the encoder encodes it to 444 and sends it, then 444 is received and renders it to 888 or 444. in this case, I'd like to see the original 888, the encoded 444, and the final decoded 888 or 444

vr000m:
LGTM, modulo adding bit depth to media-source.

webrtc-stats.html
```diff
@@ -2134,6 +2147,17 @@ <h2>
                   Before the first frame is encoded this attribute is missing.
                 </p>
               </dd>
+              <dt>
+                <dfn><code>frameBitDepth</code></dfn> of type <span class="idlMemberType">unsigned
+                long</span>
+              </dt>
+              <dd>
+                <p>
+                  Only valid for video. Represents the bit depth per pixel of the raw video frame.
```
henbos: Copy this to RTCVideoSourceStats but rephrase this sentence:
"Represents the bit depth per pixel of the last frame originating from this source" (i.e. before encoding)

And keep it here as well but rephrase:
"Represents the bit depth per pixel of the last encoded frame" (i.e. after encoding)

webrtc-stats.html
henbos: For consistency with my other comment, I suggest rephrasing it:
"Represents the bit depth per pixel of the last decoded frame"

alvestrand: Just for fun ... are you reporting the bit depth for one color, or the bit depth for a pixel?
in 4.2.0 the colors have different bit depths - I suppose this would be the "encoded" bit depth.

drkron: @alvestrand Thanks for the correction! I just saw your first message in my mail and have been thinking about whether this was the case or not.

drkron: @vr000m Thanks for the suggestions!
However, I'm a bit confused here, what do you mean by bit depth of the encoded video frame? To me it only makes sense to talk about bit depth when the frame is not encoded, i.e., raw. When the frame is encoded, the bits are stored in a buffer that only has a meaning to the decoder. The only meaning of a bit depth field for the encoded video frame would be to determine the bit depth of the frame after it's been decoded.
What do you think of adding frameBitDepth to RTCVideoSourceStats as suggested, but removing it completely from RTCOutboundRtpStreamStats?

vr000m: I was trying to figure out where or which dictionary the metric should be reported on. The discussion with @henbos was if this is before encoding and after. Also if the bit depth can change. 

If this is a value before encoding then outbound would be the wrong place to put it.

drkron: Thanks, then I understand.
I've removed it from outbound and added it to RTCVideoSourceStats.

webrtc-stats.html
drkron: Moved from outbound to RTCVideoSourceStats.

webrtc-stats.html
```diff
@@ -2777,6 +2790,17 @@ <h2>
...
                 </p>
               </dd>
+              <dt>
+                <dfn><code>bitDepth</code></dfn> of type <span class="idlMemberType">unsigned
+                long</span>
+              </dt>
+              <dd>
+                <p>
+                  The bit depth per pixel of the last frame originating from this source. Before a
+                  frame has been produced this attribute is missing.
```
henbos: Is this enough or do we also want frameBitDepth on the RTCOutboundRtpStreamStats representing the bit depth *after* encoding? If the bit depth is higher from the camera than what we encode.

drkron: You're right, that could happen. Added it back.
Thanks for the comment!

vr000m:
Yes, I guess we have all three
1. source, before encoding
2. Outbound, after encoding
3. Inbound, at the time of decoding.

Issue #512, opened 2019-10-18
wrong definition of RTCIceServerStats.protocol?


fippo:
The definition of [RTCIceServerStats.protocol](https://w3c.github.io/webrtc-stats/#dom-rtciceserverstats-protocol) seems to be copied from another place, most likely [this one](https://w3c.github.io/webrtc-stats/#dom-rtcicecandidatestats-protocol):
```
    Valid values for transport is one of udp and tcp. 
    Based on the "transport" defined in [RFC5245] section 15.1.
```
which is the protocol used by the candidate. We're rather interested in the protocol used to connect to the TURN server to say "how long is the RTT for this udp turn server", no?
I.e. we want [relayProtocol](https://w3c.github.io/webrtc-stats/#dom-rtcicecandidatestats-relayprotocol):
```
It is the protocol used by the endpoint to communicate with the TURN server.
This is only present for local candidates. Valid values are udp, tcp, or tls. 
```
We could obviously extract that from the url but...

chore task: all the references to 5245 need to be updated :-|

vr000m:
@lennart-csio Please take a look

lennart-csio:
the intention is indeed to have the protocol to connect to the server, i will clarify.
@fippo what do you mean by the core task? at least in the RtcIceServerStats dict there is only this one reference. should it be a task in this issue to go through all references to the RFC and check if its accurate?

fippo:
@lennart-csio, chore, not core. RFC 5245 got obsoleted by 8445 so all references to the old one need to be updated eventually. Which is incredibly boring ;-)

lennart-csio:
thanks for clarifying :)
i will do the fix first as a PR, then we can see if we want to have another issue for updating the RFC reference or not

PR #519, opened 2019-11-22
RTCIceServerStats relayProtocol

lennart-csio:
for issue #512 
clarifying the RTCIceServerStats protocol by renaming to relayProtocol and using similar text and naming than in RTCIceCandidateStats

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -4483,7 +4483,7 @@ enum RTCNetworkType {
           <pre class="idl">dictionary RTCIceServerStats : RTCStats {
              DOMString url;
              long port;
-             DOMString protocol;
+             DOMString relayProtocol;
              RTCNetworkType networkType;
              unsigned long totalRequestsSent;
              unsigned long totalResponsesReceived;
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -4512,13 +4512,14 @@ <h2>
...
+                <dfn><code>relayProtocol</code></dfn> of type <span class=
                 "idlMemberType">DOMString</span>
               </dt>
               <dd>
                 <p>
-                  Valid values for transport is one of <code>udp</code> and <code>tcp</code>. Based
-                  on the "transport" defined in [RFC5245] section 15.1.
+                  It is the protocol used by the endpoint to communicate with the ICE server.
+                  Valid values are <code>udp</code>, <code>tcp</code>, or <code>tls</code>.
+                  This is the same value that is used for the relay protocol of local ICE candidates.
```
alvestrand: What is the reference for "the relay protocol of local ICE candidates"?
Do we have a field in [[WEBRTC]] for it? Or is it an ICE-defined property? If so, which RFC?

vr000m: @fippo could you weigh in...

henbos:
@vr000m to rebase and add reference

henbos:
Ping @vr000m

webrtc-stats.html
fippo: missed that This is an internal reference to https://w3c.github.io/webrtc-stats/#dom-rtcicecandidatestats-relayprotocol

vr000m: Fixed

vr000m:
@alvestrand @henbos Please take a look

PR #526, opened 2020-01-15
Clarify qualityLimitationReason priority order

### LINKED ISSUE #440: Clarify qualityLimitationReason when limited by multiple reasons

henbos:
Fixes #440.

(Makes #504 obsolete)

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2405,11 +2405,11 @@ enum RTCStatsType {
                   framerate, or "none" if not limited.
                 </p>
                 <p>
-                  If there are multiple limitation reasons at the same time, the
-                  implementation SHOULD report the most limiting factor. If the
-                  implementation is not able to determine this, the reasons
-                  SHOULD be reported in the following order of priority:
-                  "bandwidth", "cpu", "other".
+                  The implementation reports the most limiting factor. If the
+                  implementation is not able to determine the most limiting
+                  factor because multiple may exist, the reasons MUST be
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -2399,6 +2399,21 @@ <h2>
                   Only valid for video. The current reason for limiting the resolution and/or
                   framerate, or "none" if not limited.
                 </p>
+                <p>
+                  If there are multiple limitation reasons at the same time, the
+                  implementation SHOULD report the most limiting factor. If the
+                  implementation is not able to determine this, the reasons
+                  SHOULD be reported in the following order of priority:
+                  "bandwidth", "cpu", "other".
```
henbos: I don't think I can use the word "MUST" with regards to the order, because the any attempt to determine the "most limiting factor" is an estimate. We can't encourage using an estimate but then say that if the estimate is not correct you MUST report something else - this "MUST" would require the estimate to not be an estimate but to be the true answer. This is not implementable. So both uses remaing "SHOULD"

vr000m: The implementation reports the most limiting factor. If the implementation is not able to determine the most limiting factor because multiple may exist, the reasons MUST  be reported in the following order of priority: "bandwidth", "cpu", "other".

PR #546, opened 2020-02-04
Mark some stats as only 'present' rather than 'valid' for some type o…

dontcallmedom:
…f stream

Presumably the intent is that if the stream is not of the right kind, the data should not be included ('valid' makes it seem like it would be ok to send garbage data in other cases)


alvestrand:
Looks good to me. It was intended as "this definition is only valid for that type", but "present" is less prone to misunderstanding.

Issue #548, opened 2020-02-06
Expose SCTP stats in RTCSctpTransportStats


shacharz:
Applications that use data channels intensively would find it useful to have data from SCTP layer:
- **congestion window size** (spinfo_cwnd)
- **receiver window size** (sstat_rwnd)
- **MTU** (spinfo_mtu)
- **packet loss ratio** (not exposed directly in sctp as far as I can tell, but might be able to calculate from sstat_unackdata)

For example:
- congestion, receiver, MTU and rtt together can predict the throughput of the channel, and help prioritize different channels.
- packet loss can help identifying a problem somewhere in the network path.

vr000m:
Make it into PR and we can discuss it there.. I think there was https://github.com/w3c/webrtc-stats/issues/377 and concerns raised in that issue.

alvestrand:
the first use case is a particular way of achieving user-defined congestion control, which is a can of worms for which I think stats is not appropriate.
The second use case seems reasonable on the surface - but it's not exposing a stat that is predefined in SCTP. Exposing scstat_unackdata might be more interesting.

shacharz:
Corresponding bugs tracked on implementors
- [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=1088402)
- [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1642286)
- [Webkit](https://bugs.webkit.org/show_bug.cgi?id=212589)

Issue #377, opened 2018-10-18
[DataChannels] Expose bandwidth or congestion window from the SCTP lib


shacharz:
It can be very helpful for applications to get the bandwidth/c_wnd from the SCTP layer, so it can adjust business logic accordingly.
@lgrahl

alvestrand:
How is this supposed to work when SCTP runs on top of DTLS which runs on top of ICE?

lgrahl:
For SCTP, UDP/DTLS looks like a raw socket (and AFAIK pretty much behaves like that).

vr000m:
I am curious what things can the application can do in response to the c_wnd stats?

lgrahl:
That is indeed an interesting question since we already have a (quirky) flow control with `.bufferedAmount` and friends.

shacharz:
> I am curious what things can the application can do in response to the c_wnd stats?

decide whether to send data to this client or another (each client is connected to multiple others)

vr000m:
Since, we are on the topic of this, what other SCTP stats would applications want to use? also are there any other SCTP metrics that would be interesting for the app

shacharz:
> That's a bug.

AFAIK it's not a bug, since bufferedAmount don't include the sctp buffer.


> Since, we are on the topic of this, what other SCTP stats would applications want to use? also are there any other SCTP metrics that would be interesting for the app

- opened a different issue for RTT
- packet loss would also be interesting.
- also interesting but more futuristic, for us at least: maximum size for data payload before packet will be split to 2 (mtu - all the headers calculation)

lgrahl:
> > That's a bug.
> 
> AFAIK it's not a bug, since bufferedAmount don't include the sctp buffer.

It is definitely a bug. See the [corresponding spec section](https://w3c.github.io/webrtc-pc/#dom-rtcdatachannel-bufferedamount). The only reason why all implementations are affected by this is because they all use usrsctp which currently does not expose that information.

> Since, we are on the topic of this, what other SCTP stats would applications want to use? also are there any other SCTP metrics that would be interesting for the app

I'm not seeing further items that should be in the stats (there are a couple that could be added to `RTCSctpTransport` at some point, such as MTU, partial delivery point, RT max, RTO initial/min/max, heartbeat interval, ...). Arguably, some of these *could* also be in the stats but I'm not entirely convinced about that so far.

[...]

PR #551, opened 2020-02-20
RTCSctpTransportStats additional stats

shacharz:
added congestionWindow, receiverWindow, mtu,  unackdata
fixes https://github.com/w3c/webrtc-stats/issues/548
fixes https://github.com/w3c/webrtc-stats/issues/377

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -3615,7 +3615,11 @@ enum RTCStatsType {
         </p>
         <div>
           <pre class="idl">dictionary RTCSctpTransportStats : RTCStats {
-    double smoothedRoundTripTime;
+            double smoothedRoundTripTime;
+            unsigned long congestionWindow;
+            unsigned long receiverWindow;
+            unsigned long mtu;
+            unsigned long unackData;
 };</pre>
           <section>
             <h2>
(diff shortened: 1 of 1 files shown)


vr000m:
Is this a snapshot of the metrics at the point getStats is called? And if there is a concern that the application will poll it several times a second, which may not be ideal.

The other question is if we should go for cumulative counters, which would need to be implemented in sctp-lib, is there any commitment from anyone to do this?

henbos:
How often would you intent to poll getStats as a user of this information?
The SCTP RFC defines a polling mechanism to obtain this which I assume is efficient. getStats() is heavier.

alvestrand:
Can you clean up the PR so that it doesn't have random whiltespace changes all over the place?

henbos:
Despite the polling concerns the editors are happy to merge this if its cleaned up

henbos:
Ping @shacharz please clean up the PR so we can merge it

shacharz:
I'm not sure why the build is failing, would love some help with that.

henbos:
Could be unrelated, re-running travis...

dontcallmedom:
The Travis failure is not linked to this particular pull request, but to a [recent change in WebIDL](https://github.com/heycam/webidl/commit/8d4e7ca67bcb4de476aab0411531dc97f6d17e0d) that deprecates the term "present". Feel free to ignore the Travis failure, I'll send a separate PR to fix that problem

henbos:
Thanks Dom. then let's merge this!

PR #554, opened 2020-03-18
Clarify how RTX works and adds the rtxSsrc stat

### LINKED ISSUE #553: Standard getStats() and RTX streams
The spec has this to say about RTX:

> When there are multiple RTP streams connected to the same sender, such as when using simulcast or RTX, there will be one RTCOutboundRtpStreamStats per RTP stream, with distinct values of the ssrc member

The RTCOutboundRtpStreamStats dictionary contains bytesSent (all payload bytes, including retransmissions) and retransmittedBytesSent (the subset of bytesSent that are retransmissions).

In the non-RTX case it is clear what value is what.

In the RTX case, the spec suggests there would be a separate outbound-rtp stream stats object. But this does not align with implementations. Implementing the spec raises several questions:
- How do I know which stream is the normal one and which one is the RTX one?
- Would the RTX stream have any metrics other than byte counters defined? For example resolution, frame rate and encode time seems rather strange for an RTX stream?
- Would bytesSent always equal to retransmittedBytesSent? Or would we only define one of them?

But even more importantly: Is there any value in creating a separate stats object? At least in the webrtc.org implementation, if I understand correctly, there is always one RTX stream per RTP stream when RTX is used. So why not just add those bytes to the original outbound-rtp stats object?

**Proposal A**
Add a metric like "isRTX" so you know which object is which, and a reference to the original outbound-rtp object.

**Proposal B**
Merge the RTX into the original outbound-rtp objects' retransmittedBytesSent.

henbos:
Fixes #553.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -446,13 +446,17 @@ enum RTCStatsType {
               {{RTCOutboundRtpStreamStats}}.
             </p>
             <p>
-              When there are multiple <a>RTP stream</a>s connected to the same sender, such as when using
-              simulcast or RTX, there will be one {{RTCOutboundRtpStreamStats}} per <a>RTP stream</a>,
+              When there are multiple <a>RTP stream</a>s connected to the same sender due to using
+              simulcast, there will be one {{RTCOutboundRtpStreamStats}} per <a>RTP stream</a>,
               with distinct values of the {{RTCRtpStreamStats/ssrc}} member, and all these senders will have a
               reference to the same {{RTCStatsType/"sender"}} object (of type {{RTCAudioSenderStats}} or
               {{RTCVideoSenderStats}}) and {{RTCStatsType/"track"}} object (of type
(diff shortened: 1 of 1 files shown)


henbos:
I think this is what we want the spec to say. It's confusing if bytes and retransmissions are put in separate counters and have to manually be added together to get something sensible - it matters little whether they are sent on ssrc or rtxSsrc.

However, the fact that bytesSent includes retransmissions means that "Total number of RTP packets sent for this &lt;a&gt;SSRC&lt;/a&gt;" is actually referring to the bytes sent of both the {{RTCRtpStreamStats/ssrc}} and the {{RTCOutboundRtpStreamStats/rtxSsrc}}.

This matches current implementation though.

henbos:
This PR was reviewed by ilnik@webrtc.org (Google developer) who knows the RTX code, he says LGTM.

henbos:
@vr000m Can you merge this?

PR #557, opened 2020-03-31
Adapt references to dictionary member existence

dontcallmedom:
[WebIDL no longer defines a dictionary member as present](https://github.com/heycam/webidl/commit/8d4e7ca67bcb4de476aab0411531dc97f6d17e0d) but instead makes use of [map existence](https://infra.spec.whatwg.org/#map-exists) as terminology. The spec is updated to make a similar usage.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1118,7 +1118,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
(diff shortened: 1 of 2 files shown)


dontcallmedom:
merging this will make Travis happy again (cf https://github.com/w3c/webrtc-stats/pull/551#issuecomment-606583919)

PR #559, opened 2020-04-21
RTCScalableVideoCodecLayerStats added

### LINKED ISSUE #401: Add SVC stats in RTC[In/Out]boundRtpStreamStats
perResolutionFramesSent/Received gives the frame counters for each resolution

distributionOfTimeSinceLastFrameSent/Received gives the amount of time between frames, so !% FPS (66.66ms), 30 FPS (33.33ms) and 60 FPS (16.66ms)

henbos:
Fixes #401

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -97,6 +97,9 @@
       <p>
         The term <dfn><a href="https://tools.ietf.org/html/rfc7656#section-2.1.10">RTP stream</a></dfn> is defined in [[RFC7656]].
       </p>
+      <p>
+        The term Scalable Video Coding (<dfn>SVC</dfn>) is defined in [[RFC6190]].
+      </p>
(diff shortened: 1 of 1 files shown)


vr000m:
Is this only for the sent SVC? what about received svc?

henbos:
Only sent for now, let's discuss if receiving metrics are useful

alvestrand:
WEBRTC-SVC doesn't define access to layers yet. Let's not do anything with it before the definition there lands.

webrtc-stats.html
```diff
@@ -392,6 +395,7 @@ <h3>
 "codec",
 "inbound-rtp",
 "outbound-rtp",
+"svc-layer",
```
alvestrand, line 398: we should also add an "SVC mode" value to the video RTP sender, I think.

henbos:
> WEBRTC-SVC doesn't define access to layers yet. Let's not do anything with it before the definition there lands.

The per-layer access is discussed here: https://github.com/w3c/webrtc-svc/issues/14#issuecomment-519631248

Tag @Orphis if we want SVC stats this is blocked on the referenced SVC API. This stats PR is iceboxed in the meantime.

webrtc-stats.html
henbos: Yes that makes sense!

PR #563, opened 2020-04-21
Update targetBitrate definition

### LINKED ISSUE #560: targetBitrate should be defined as current encoder target settings
In #423 I proposed marking this metric obsolete in favor of [totalEncodedBytesTarget](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-totalencodedbytestarget) but there may be [issues with totalEncodedBytesTarget](https://bugs.chromium.org/p/webrtc/issues/detail?id=11443). For the time being, let's keep targetBitrate.

Its definition doesn't make sense. Especially "It is measured in bits per second and the bitrate is calculated over a 1 second window."
- It's an instantaneous target reflecting the encoder setting, not something you measure over a period of time.
- I'm not sure what the TIAS reference is accomplishing? Looks like a bandwidth modifier in SDP?

I propose just saying this is the current encoder target settings, which is a target for the payload bytes excluding retransmissions.

henbos:
Fixes #560.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2122,12 +2122,12 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  It is the current target bitrate configured for this particular <a>SSRC</a> and is the
-                  Transport Independent Application Specific (TIAS) bitrate [[!RFC3890]].
-                  Typically, the target bitrate is a configuration parameter provided to the
-                  codec's encoder and does not count the size of the IP or other transport layers
-                  like TCP or UDP. It is measured in bits per second and the bitrate is calculated
-                  over a 1 second window.
+                  Reflects the current encoder target in bits per second. The
+                  target is an instantanous value reflecting the encoder's
+                  settings, but the resulting payload bytes sent per second,
+                  excluding retransmissions, SHOULD closely correlate to the
+                  target. See also {{RTCSentRtpStreamStats/bytesSent}} and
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -2122,12 +2122,12 @@ <h2>
...
-                  Typically, the target bitrate is a configuration parameter provided to the
-                  codec's encoder and does not count the size of the IP or other transport layers
-                  like TCP or UDP. It is measured in bits per second and the bitrate is calculated
-                  over a 1 second window.
+                  Reflects the current encoder target in bits per second. The
+                  target is an instantanous value reflecting the encoder's
+                  settings, but the resulting payload bytes sent per second,
+                  excluding retransmissions, SHOULD closely correlate to the
+                  target. See also {{RTCSentRtpStreamStats/bytesSent}} and
+                  {{RTCOutboundRtpStreamStats/retransmittedBytesSent}}.
```
alvestrand, line 2130: The targetBitrate is defined in the same way as the Transport Independent Application Specific (TIAS) bitrate [[!RFC3890]].

vr000m:
Do the implementations really exclude retransmittedBytes, I would have imagined that RTX,FEC would be included. I dont have a strong opinion on this

henbos:
I believe current implementations refer to what the encoder tries to produce and that RTX would be separate from this

PR #566, opened 2020-04-21
remoteSource obsolete, relayedSource added

### LINKED ISSUE #453: Add remoteSource to "media-source"
There were two meansings of remoteSource discussed:
1. Was the track sourced from a peer connection or not?
2. Was the track on a sender or a receiver?

Part of #452 should be to cover 2). But @vr000m wants 1) to be an available stat; let's add this version of the metric to "media-source"

### LINKED ISSUE #528: Move RTCMediaHandlerStats::remoteSource to the obsolete section
With "sender" and "receiver" stats, you can already tell if it's remote or not. (Not to be confused with Varun's conception of "remoteSource": #453)

henbos:
Fixes #528.
Fixes #453.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2639,6 +2639,7 @@ enum RTCStatsType {
           <pre class="idl">dictionary RTCMediaSourceStats : RTCStats {
              DOMString       trackIdentifier;
              DOMString       kind;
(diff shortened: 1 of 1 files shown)


vr000m:
relayed source sounds good to me.

Issue #549, opened 2020-02-07
Define reception time for jitterBufferDelay stat


murillo128:
Currently the `jitterBufferDelay` stat is defined as:
```
It is the sum of the time, in seconds, each audio sample or video frame takes from the time 
it is received and to the time it exits the jitter buffer.
```
When tried to land a patch for libwebrtc to correct the current implementation (https://webrtc-review.googlesource.com/c/src/+/168042), it has been risen the concern that it is not clear what the reception time really mean:

1. The reception time of the RTP packet as close to the network layer as possible
2. The reception time of the RTP packet by the jitter buffer code
3. The time when the audio samples have been already decoded and ready to be passed to the jitter buffer internal queues

In case of (2) or (3), we would need a new stat defined for (1)  (playoutDelay?)

henbos:
@alvestrand @ivocreusen Do you have an opinion?

I think the original wording of the spec was supposed to describe the implementation, or something close enough to be standardizable, but I don't know the details well enough to have an authority here.

Questions:
- What does the current implementation measure?
- What would be the most useful thing to measure?

I think we can either update the implementation or update the spec. Or update the spec to match the implementation and then add another metric to give us the remaining information, if the current implementation is still useful on its own.

murillo128:
Currently, libwebrtc implementation sets the reception time as when the rtp packet is inserted in NetEq, my patch uses RTP network reception time instead.

As an app developer, I don't really care about how much delay does the NetEq introduce but what is the total playout delay since the rtp packet was received from the network. 

In normal conditions, the processing time before NetEq would be minimum, but I have seen it going up to several 10ms in cases of high cpu loads.

henbos:
Taking a closer look at the spec definition and discussing with libwebrtc colleagues:

I think the intent of the spec is that it is the time spent in the jitter buffer, i.e. the time it is inserted until the time it is exited. I think the spec should be clarified to say "time it enters" rather than "is received". This is well defined enough that different webrtc implementations could be confident that they're measuring the same part of the pipeline, and simple enough to implement.

As I understand things, there isn't additional _buffering_ going on prior to the jitter buffer other than perhaps network card and OS level buffering before the packet reaches webrtc, which we don't have a metric for, probably don't want to have a metric for, and isn't what your CL attempts to calculate anyway.

[...]

PR #567, opened 2020-04-21
clarifies JB metric measurement

### LINKED ISSUE #441: Clarify jitterBufferDelay/jitterBufferEmittedCount for video
I'm not sure "jitter buffer" is well defined for video. There are various buffers, so for example, does this include decoding time or not?

vr000m:
Fixes #441 and partially #549

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1557,14 +1557,15 @@ enum RTCStatsType {
                   The purpose of the jitter buffer is to recombine RTP packets into frames (in the case of video)
                   and have smooth playout. The model described here assumes that the samples or frames are 
                   still compressed and have not yet been decoded.
-                  It is the sum of the time, in seconds, each <a>audio sample</a> or video frame takes from
-                  the time it the first packet received by the jitter buffer (ingest timestamp) and to the 
+                  It is the sum of the time, in seconds, each <a>audio sample</a> or a video frame takes from
+                  the time the first packet is received by the jitter buffer (ingest timestamp) to the 
                   time it exits the jitter buffer (emit timestamp). 
                   In the case of audio, several samples belong to the same RTP packet, hence they will have the same 
                   ingest timestamp but different jitter buffer emit timestamps. 
-                  In the case of video, the frame is received over several RTP packets, hence the ingest timestamp
-                  is the earliest packet of the frame that entered the jitter buffer.
-                  This increases upon samples or frames exiting, having completed their time in the buffer (and
+                  In the case of video, the frame maybe is received over several RTP packets, hence the ingest timestamp
+                  is the earliest packet of the frame that entered the jitter buffer and the emit timestamp is 
+                  when the whole frame exits the jitter buffer.
+                  This metric increases upon samples or frames exiting, having completed their time in the buffer (and
                   incrementing {{jitterBufferEmittedCount}}). The average jitter buffer
                   delay can be calculated by dividing the {{jitterBufferDelay}} with the
                   {{jitterBufferEmittedCount}}.
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1555,9 +1555,15 @@ enum RTCStatsType {
               <dd>
                 <p>
                   The purpose of the jitter buffer is to recombine RTP packets into frames (in the case of video)
-                  and have smooth playout.  
+                  and have smooth playout. The model described here assumes that the samples or frames are 
+                  still compressed and have not yet been decoded.
                   It is the sum of the time, in seconds, each <a>audio sample</a> or video frame takes from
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1554,9 +1554,11 @@ <h2>
               </dt>
               <dd>
                 <p>
+                  The purpose of the jitter buffer is to recombine RTP packets into frames (in the case of video)
+                  and have smooth playout.  
                   It is the sum of the time, in seconds, each <a>audio sample</a> or video frame takes from
-                  the time it is received and to the time it exits the jitter buffer. This increases
-                  upon samples or frames exiting, having completed their time in the buffer (and
+                  the time it is received by the jitter buffer and to the time it exits the jitter buffer.
```
alvestrand, line 1560: Need to add "the time the first packet of the frame is received by the jitter buffer"?

If the assembly is within the jitter buffer. If it's before the jitter buffer, it's the last packet.

henbos, line 1560: Clarify that frames in jitter buffer refers to encoded frames.

vr000m:
@alvestrand @henbos Please take a look

webrtc-stats.html
```diff
@@ -1554,9 +1554,17 @@ <h2>
               </dt>
               <dd>
                 <p>
+                  The purpose of the jitter buffer is to recombine RTP packets into frames (in the case of video)
+                  and have smooth playout. The model described here assumes that the samples or frames are 
+                  still compressed and have not yet been decoded.
```
henbos, line 1559: Is it correct to add: "...not yet ben decoded, but a valid implementation could decode the packets prior to insertion into the jitter buffer." ?

webrtc-stats.html
```diff
@@ -1554,9 +1554,17 @@ <h2>
...
                   It is the sum of the time, in seconds, each <a>audio sample</a> or video frame takes from
-                  the time it is received and to the time it exits the jitter buffer. This increases
-                  upon samples or frames exiting, having completed their time in the buffer (and
+                  the time it the first packet received by the jitter buffer (ingest timestamp) and to the
```
henbos, line 1561: the time the first packet

webrtc-stats.html
```diff
@@ -1554,9 +1554,17 @@ <h2>
...
               <dd>
                 <p>
+                  The purpose of the jitter buffer is to recombine RTP packets into frames (in the case of video)
+                  and have smooth playout. The model described here assumes that the samples or frames are 
+                  still compressed and have not yet been decoded.
                   It is the sum of the time, in seconds, each <a>audio sample</a> or video frame takes from
-                  the time it is received and to the time it exits the jitter buffer. This increases
-                  upon samples or frames exiting, having completed their time in the buffer (and
+                  the time it the first packet received by the jitter buffer (ingest timestamp) and to the 
+                  time it exits the jitter buffer (emit timestamp).
```
henbos, line 1562: Do we need to clarify "to the time [the last packet of that frame] exits the jitter buffer"?

webrtc-stats.html
```diff
@@ -1554,9 +1554,17 @@ <h2>
...
                   It is the sum of the time, in seconds, each <a>audio sample</a> or video frame takes from
-                  the time it is received and to the time it exits the jitter buffer. This increases
-                  upon samples or frames exiting, having completed their time in the buffer (and
+                  the time it the first packet received by the jitter buffer (ingest timestamp) and to the 
+                  time it exits the jitter buffer (emit timestamp). 
+                  In the case of audio, several samples belong to the same RTP packet, hence they will have the same 
+                  ingest timestamp but different jitter buffer emit timestamps. 
+                  In the case of video, the frame is received over several RTP packets, hence the ingest timestamp
+                  is the earliest packet of the frame that entered the jitter buffer.
+                  This increases upon samples or frames exiting, having completed their time in the buffer (and
```
henbos, line 1567: Would it also be valid to count per-packet? So that even if a frame spans 3 different packets, this would yield 3 different emitted counts?

webrtc-stats.html
vr000m: Typically, Jitter buffer makes sure it hands the full frame for decoding and also approximates the time to playout. So major part of the buffering is done here and not downstream in the playout buffer, which is a small circular buffer (keeps a frame or few)

webrtc-stats.html
vr000m: the packets would have the same emit timestamp, it kinda depends on what we want to divide the number with? frames or packets?

I think frames.

webrtc-stats.html
vr000m: If there is down stream delays or buffer. we need a different buffer metric, for example a playout buffer metric.

PR #568, opened 2020-04-21
playout delay

### LINKED ISSUE #549: Define reception time for jitterBufferDelay stat

vr000m:
Fixes #549

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1539,17 +1539,22 @@ enum RTCStatsType {
                   It is the sum of the time, in seconds, each <a>audio sample</a> or video frame takes from
                   the time the first RTP packet is received (reception timestamp) and to the time 
                   the corresponding sample or frame is delivered to the MediaStreamTrack for playout (playout timestamp). 
+                  Typically ready for playout here means after the decoding audio sample or video frame.
+                </p>
+                <p>
                   Given the complexities involved, the time of arrival or the reception timestamp is measured 
                   as close to the network layer as possible and the playout timestamp is measured as close to
                   the rendering as possible.
+                </p>
+                <p>
                   In the case of audio, several samples are received in the same RTP packet, all samples
                   will share the same reception timestamp and different playout timestamps. 
                   In the case of video, the frame are received over several RTP packets, in this
                   case the earliest timestamp containing the frame is counted as the reception timestamp,
-                  and the reception timestamp corresponds to when the complete frame is played out.
+                  and the playout timestamp corresponds to when the complete frame is ready to be played out.
                   This metric is not incremented for frames that are not decoded, 
                   i.e., {{framesDropped}}, {{partialFramesLost}},
-                  The average playoutDelay delay can be calculated by dividing the {{playoutDelay}} with the
+                  The average playout delay can be calculated by dividing the {{playoutDelay}} with the
                   {{totalSamplesReceived}} for audio or {{framesDecoded}} for video.
                 </p>
                 
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1190,7 +1190,7 @@ enum RTCStatsType {
              unsigned long        firCount;
              unsigned long        pliCount;
              unsigned long        sliCount;
-             double               totalPlayoutDelay;
+             double               totalProcessingDelay;
              DOMHighResTimeStamp  estimatedPlayoutTimestamp;
              double               jitterBufferDelay;
              unsigned long long   jitterBufferEmittedCount;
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1190,6 +1190,7 @@ <h3>
              unsigned long        firCount;
              unsigned long        pliCount;
              unsigned long        sliCount;
+             double               playoutDelay;
```
alvestrand, line 1193: playoutDelaySum or totalPlayoutDelay

webrtc-stats.html
```diff
@@ -1529,6 +1530,26 @@ <h2>
...
                 </p>
               </dd>
+              <dt>
+                <dfn>playoutDelay</dfn> of type <span class=
+                "idlMemberType">double</span>
+              </dt>
+              <dd>
+                <p>
+                  It is the sum of the time, in seconds, each <a>audio sample</a> or video frame takes from
+                  the time it is received and to the time it is played out. Given the complexities involved,
```
alvestrand, line 1540: "it" = first packet or last packet?

webrtc-stats.html
```diff
@@ -1529,6 +1530,26 @@ <h2>
...
+                  the time of arrival or reception timestamp is measured as close to the network layer as 
+                  possible and the playout timestamp is measured as close to rendering as possible.
+                  In the case of audio, several samples are received in the same RTP packet, all samples
+                  will share the same reception timestamp and different playout timestamps. 
+                  Whereas, in the case of video the frame are received over several RTP packets, in this
```
alvestrand, line 1545: English: Delete "Whereas," - it confuses.

webrtc-stats.html
```diff
@@ -1529,6 +1530,26 @@ <h2>
...
+                  case the earliest timestamp containing the frame is counted as the reception timestamp,
+                  and the reception timestamp corresponds to when the complete frame is played out.
+                  The average playoutDelay delay can be calculated by dividing the {{playoutDelay}} with the
+                  sum of {{framesDropped}}, {{partialFramesLost}},{{framesReceived}}.
```
alvestrand, line 1549: This assumes that {{framesDropped}} and {{partialFramesLost}} are "played out" in some sense; you have to define a playout time for them (might be "same as following frame").

webrtc-stats.html
henbos, line 1540: The time the audio sample or video frame is delivered to the MediaStreamTrack.

webrtc-stats.html
alvestrand: Better to only counts framesReceived.

henbos, line 1549: framesReceived is only defined for video. Which metrics are the corresponding ones for audio? totalSamplesReceived?

vr000m: Used framesDecoded

vr000m: framesDecoded for video and totalSamplesReceived for audio

vr000m:
@alvestrand @henbos Please take a look and merge if acceptable

webrtc-stats.html
```diff
@@ -1529,6 +1530,30 @@ <h2>
...
+                  the time the first RTP packet is received (reception timestamp) and to the time 
+                  the corresponding sample or frame is delivered to the MediaStreamTrack for playout (playout timestamp). 
+                  Given the complexities involved, the time of arrival or the reception timestamp is measured 
+                  as close to the network layer as possible and the playout timestamp is measured as close to
+                  the rendering as possible.
+                  In the case of audio, several samples are received in the same RTP packet, all samples
+                  will share the same reception timestamp and different playout timestamps. 
+                  In the case of video, the frame are received over several RTP packets, in this
+                  case the earliest timestamp containing the frame is counted as the reception timestamp,
+                  and the reception timestamp corresponds to when the complete frame is played out.
```
henbos, line 1549: Change "complete frame is played out" to "complete frame is ready to be played out"

webrtc-stats.html
```diff
@@ -1529,6 +1530,30 @@ <h2>
...
+                  This metric is not incremented for frames that are not decoded, 
+                  i.e., {{framesDropped}}, {{partialFramesLost}},
+                  The average playoutDelay delay can be calculated by dividing the {{playoutDelay}} with the
+                  {{totalSamplesReceived}} for audio or {{framesDecoded}} for video.
```
henbos, line 1553: totalSamplesReceived are "The total number of samples that have been received on this RTP stream."
But on call it sounded like we have encoded samples and decoded samples being received... for playout, we need to event where the samples have finished decoding and are ready to be played out. Is totalSamplesReceived before or after ready to play out?

webrtc-stats.html
```diff
@@ -1529,6 +1530,30 @@ <h2>
...
+                  as close to the network layer as possible and the playout timestamp is measured as close to
+                  the rendering as possible.
+                  In the case of audio, several samples are received in the same RTP packet, all samples
+                  will share the same reception timestamp and different playout timestamps. 
+                  In the case of video, the frame are received over several RTP packets, in this
+                  case the earliest timestamp containing the frame is counted as the reception timestamp,
+                  and the reception timestamp corresponds to when the complete frame is played out.
+                  This metric is not incremented for frames that are not decoded, 
+                  i.e., {{framesDropped}}, {{partialFramesLost}},
+                  The average playoutDelay delay can be calculated by dividing the {{playoutDelay}} with the
```
henbos, line 1552: "average playoutDelay delay" -> "average playout delay"

webrtc-stats.html
henbos: reception timestamp should say playout timestamp here

webrtc-stats.html
```diff
@@ -1529,6 +1530,30 @@ <h2>
...
+                  {{totalSamplesReceived}} for audio or {{framesDecoded}} for video.
+                </p>
```
henbos, line 1554: It should be clear that totalPlayoutDelay is incremented at the same time as the samples counter, in this time framesDecoded. This implies that the "playout time" is the time that the frames have been decoded.

This part imples that the earlier "[playout] timestamp corresponds to when the complete frame is [ready to be played out]." can be rephrased as "playout timestamp corresponds to when the complete frame is decoded".

If this isn't the intent of the PR I think we need a new frames counter. totalPlayableFrames? Otherwise, all the references to "playout" should change to "decoded"

webrtc-stats.html
vr000m: totalSamplesReceived is before decoding

webrtc-stats.html
vr000m: decoded is the correct point of view.

vr000m:
@henbos PTAL, I feel this is closer to what we want.

webrtc-stats.html
```diff
@@ -1529,6 +1530,35 @@ <h2>
...
+                <p>
+                  In the case of audio, several samples are received in the same RTP packet, all samples
+                  will share the same reception timestamp and different playout timestamps. 
+                  In the case of video, the frame are received over several RTP packets, in this
+                  case the earliest timestamp containing the frame is counted as the reception timestamp,
+                  and the playout timestamp corresponds to when the complete frame is ready to be played out.
+                  This metric is not incremented for frames that are not decoded, 
+                  i.e., {{framesDropped}}, {{partialFramesLost}},
+                  The average playout delay can be calculated by dividing the {{playoutDelay}} with the
```
henbos, line 1557: totalPlayoutDelay

henbos:
We need to have a delay that goes up at exactly the same time as the samples counters. How about totalDecodingDelay and samplesDecoded and framesDecoded

vr000m:
1. refer to totalPlayoutDelay. or rename to totalPipelineDelay or totalProcessingDelay
2. add totalSamplesDecoded 
3. change to when packet is decoded and not as close to playout

webrtc-stats.html
vr000m: added totalSamplesDecoded

webrtc-stats.html
vr000m: added totalSamplesDecoded

vr000m: the new text should align with this

vr000m:
@henbos PTAL

henbos (approved):
Now it looks good to me. We can always bikeshed on the name, but I can't think of a better one! Approved.

PR #569, opened 2020-04-21
Add "required" specifiers within dictionaries

### LINKED ISSUE #304: Stats need to mark up which members are required
Some fields are needed for the stats to make any sense at all. Others may or may not be present, depending on the state of the connection or the quality of the implementation.

Those that are absolutely necessary should be marked up with "required" in the IDL.

alvestrand:
Fixes #304

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -159,11 +159,11 @@
         reproduced here for ease of reference.
       </p>
       <pre class="idl">
-        dictionary RTCStats {
-    DOMHighResTimeStamp timestamp;
-    RTCStatsType        type;
-    DOMString           id;
-        };
+dictionary RTCStats {
+    required DOMHighResTimeStamp timestamp;
+    required RTCStatsType        type;
+    required DOMString           id;
+};
       </pre>
       <p>
         Timestamps are expressed with {{DOMHighResTimeStamp}} [[HIGHRES-TIME]], and are defined as
@@ -750,7 +750,7 @@ enum RTCStatsType {
         </h3>
         <div>
           <pre class="idl">dictionary RTCRtpStreamStats : RTCStats {
-             unsigned long       ssrc;
+             required unsigned long       ssrc;
              DOMString           kind;
              DOMString           transportId;
              DOMString           codecId;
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -2898,8 +2905,8 @@ <h3>
         </p>
         <div>
           <pre class="idl">dictionary RTCRtpContributingSourceStats : RTCStats {
-             unsigned long contributorSsrc;
-             DOMString     inboundRtpStreamId;
+             required unsigned long contributorSsrc;
+             required DOMString     inboundRtpStreamId;
              unsigned long packetsContributedTo;
```
henbos, line 2910: Here's an example of a counter that might not be very important, but it always has a defined value if implemented (unlike framesDecoded which is only valid for video and thus has a sensible definition of undefined: the audio stream case).

Opinion: if the only meaning of "undefined" for a metric is "it hasn't been implemented" then I think is should be required, otherwise it is like saying "you can implemented this as undefined by not implementing it"... that sounds self-contradictory.

alvestrand: Seems like a reasonable argument. However, it applies to all counters.
The counterargument is for the case where an implementor sees value in knowing what ssrcs exist as CSRCs, but doesn't care about the packet value (or finds it too hard to implement correctly) - that implementation is either conformant or non-conformant to this spec; if we make the counter mandatory, it's non-conformant even in the absence of a conformance statement anywhere else.

I think I'm arguing myself into leaving it non-required.

webrtc-stats.html
```diff
@@ -750,7 +750,7 @@ <h3>
         </h3>
         <div>
           <pre class="idl">dictionary RTCRtpStreamStats : RTCStats {
-             unsigned long       ssrc;
+             required unsigned long       ssrc;
              DOMString           kind;
```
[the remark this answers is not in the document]
alvestrand: Good point. I was considering whether or not we needed an affordance for kinds that weren't supported yet, but I think that would not be an RTP stream. Let's make the kind mandatory.
(Adding a new kind is a huge undertaking in our current document structure anyway.)

webrtc-stats.html
henbos: OK. Am I correct in thinking that this implies that `required` means "the subset of mandatory to implement stats that can't be undefined if implemented"?

PR #576, opened 2020-07-22
clarify roundtriptimemeasurements

fippo:
fixes a missing word and removes another word which is in the wrong place


fippo:
@henbos as promised. Easy to review, the location of the first change is what one needs to understand why the second change is required

henbos:
@vr000m I suggest we merge this without a meeting, will you click the button?

fippo:
first we need to convince @dontcallmedom that this is non substantive for IPR - which should be easy ;-)

PR #583, opened 2020-09-08
Describe what we mean by ssrc for remote stats.

### LINKED ISSUE #454: Clarify that remote-[in/out]bound-rtp.ssrc refers to the localId's ssrc

vr000m:
Fixes #454

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -765,7 +765,6 @@ enum RTCStatsType {
                 <p>
                   The synchronization source (ssrc) identifier is an unsigned integer value per [[RFC3550]]
                   used to identify the stream of RTP packets that this stats object is describing. 
-                  This is not the SSRC in the RTCP packets.
                 </p>
                 <p>
                   For outbound and inbound local, ssrc describes the stats for the track that were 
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -763,8 +763,15 @@ <h2>
               </dt>
               <dd>
                 <p>
-                  The 32-bit unsigned integer value per [[RFC3550]] used to identify the source of
-                  the stream of RTP packets that this stats object concerns.
+                  The synchronization source (ssrc) identifier is an unsigned integer value per [[RFC3550]]
+                  used to identify the stream of RTP packets that this stats object is describing. 
+                  This is not the SSRC in the RTCP packets.
```
fippo, line 768: That isn't true?
For RR (inbound) it will be the SSRC_n, i.e. the one in the report block. It will *not* be the ssrc of the packet sender
Similarly for SR (outbound) it will be the SSRC of sender.

However those details are covered by RFC 3550 so I think it is best to remove this sentence.
```suggestion
```

vr000m: To desire is to decouple from the SSRC in the rtcp packet to the SSRC that's being measured.

PR #586, opened 2020-10-07
Make concept definitions normative

dontcallmedom:
Only the imported IDL fragment is meant to be information-only


vr000m (approved):
lgtm

Issue #621, opened 2022-04-07
Move unimplemented stats to webrtc-provisional-stats


henbos:
After several years of various metrics staying unimplemented, it's time to move them out of the main spec document in order to unblock webrtc-stats being able to reach Proposed Recommendation.

Sean-Der:
Hey @henbos 

I don't know where things like this are discussed, but I would be happy to help implement these! I have always been looking for/interested in getting involved. 

Do you think these getting removed because they really aren't useful or just can't find an owner?

alvestrand:
For instance, rtcpTransportStatsId is not getting implemented because nobody uses a separate transport for rtcp.
It's specified in the specs that it's possible to negotiate a WebRTC connection using a separate RTCP connection, but nobody does it. (I'm not sure if it works or not, in fact).

henbos:
> Do you think these getting removed because they really aren't useful or just can't find an owner?

A little bit of both. The fact that nobody has implemented them is a sign that they're not that important (people seem to survive without them), but there may be exceptions to this.

Patches are welcome, but considering the large number of unimplemented metrics, I'd make sure there is a solid use case for the metric you want to implement.

If we move a metric over to [the provisional spec](https://w3c.github.io/webrtc-provisional-stats/) and it later gains implementation support, we can always move it back into this spec when it has matured.

henbos:
I think we can close this issue now, but if there's anything else we missed we can file another issue.

PR #622, opened 2022-04-07
Remove most unimplemented stats

henbos:
One of the steps needed for #621.

> AFTER THIS PR HAS BEEN APPROVED BUT BEFORE LANDING IT:
> Create a corresponding webrtc-provisional-stats PR adding these metrics to that document instead.

The metrics removed in this PR are those not implemented according to:
- [wpt.fyi](https://wpt.fyi/results/webrtc-stats/supported-stats.html?label=experimental&label=master&aligned)
- https://webrtc-stats.callstats.io/verify/.
- [Dom's comment](https://github.com/w3c/webrtc-stats/issues/595#issuecomment-777557331)

Metrics are removed regardless if they are easy or hard to implement and regardless of their usefulness - this is entirely based on implementers' interest.

Everything is not covered in this PR, other stuff:
- #628
- #627
- #625
- #624
- #626

Stuff implemented but failing in the WPT:
- rid: This is implemented, but the WPT does not trigger simulcast.
- echoReturnLoss/echoReturnLossEnhancement: Manually verified that values show up in a real call.
- issuerCertificateId: No idea if it's possible to trigger this code path or how to do it, but there is code that maybe fills this in.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-stats/pull/622.html" title="Last updated on Jun 9, 2022, 11:15 AM UTC (7be489f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/622/c78394f...henbos:7be489f.html" title="Last updated on Jun 9, 2022, 11:15 AM UTC (7be489f)">Diff</a>

PR #623, opened 2022-04-07
Remove outbound-rtp.rid

henbos:

Chrome implemented simulcast stats, but it's not possible to say which layer an outbound-rtp stream belongs to other than sorting by size.

Should we remove it or fix Chrome?

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1990,7 +1990,6 @@ enum RTCStatsType {
              DOMString            mediaSourceId;
              DOMString            senderId;
              DOMString            remoteId;
-             DOMString            rid;
              DOMHighResTimeStamp  lastPacketSentTimestamp;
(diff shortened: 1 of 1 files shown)


solmaks:
@henbos , wasn't the idea of outbound-rtp.rid to let applications discover which simulcast stream specific stats refer to?

solmaks:
Let me move the conversation to the master PR - https://github.com/w3c/webrtc-stats/pull/622

henbos:
You're right, this is implemented ([code](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/pc/rtc_stats_collector.cc;l=644;drc=7fb345a0da63049b102e1c0bcdc8d7831110e324)), it's only missing because of the test not exercising simulcast! Abandoning this PR.

PR #624, opened 2022-04-07
Remove outbound-rtp.rtxSsrc.

henbos:

The retransmission counters are present regardless if the RTX is on the same or a separate SSRC, so rtxSsrc is just additional info not needed for the RTX counters to work.


henbos:
> If this information is not filled (it should) then there's no way to get the SSRC used by the RTX RTP stream, which seems like an issue.

Agree. If anyone cares enough about this issue to implement the metric we'll add it back to the main spec

PR #628, opened 2022-04-07
Remove sender, receiver and transceiver.

henbos:

This removes the sender, receiver and transceiver stats objects that were never implemented.

This PR also moves some attributes from transceiver to inbound/outbound-rtp to fill the missing holes.
- _trackIdentifier_ and _kind_ are added at inbound-rtp. This info is currently only implemented in obsolete track stats, so without receiver stats this information needs to be available in inbound-rtp in order to get rid of obsolete track stats.
- They're NOT added to outbound-rtp though, since this information is already available via the media-source in that case.
- _mid_ is added to inbound/outbound-rtp. Though technically this one is at risk, so perhaps we should just remove it?

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -396,9 +396,6 @@ enum RTCStatsType {
 "data-channel",
 "stream",
 "track",
-"transceiver",
-"sender",
-"receiver",
 "transport",
 "sctp-transport",
 "candidate-pair",
@@ -444,9 +441,8 @@ enum RTCStatsType {
             <p>
               When there are multiple <a>RTP stream</a>s connected to the same sender due to using
               simulcast, there will be one {{RTCOutboundRtpStreamStats}} per <a>RTP stream</a>,
-              with distinct values of the {{RTCRtpStreamStats/ssrc}} member, and all these senders will have a
-              reference to the same {{RTCStatsType/"sender"}} object (of type {{RTCAudioSenderStats}} or
-              {{RTCVideoSenderStats}}) and {{RTCStatsType/"track"}} object (of type
+              with distinct values of the {{RTCRtpStreamStats/ssrc}} member, and all these streams will have a
+              reference to the same {{RTCStatsType/"track"}} object (of type
               {{RTCSenderAudioTrackAttachmentStats}} or
               {{RTCSenderVideoTrackAttachmentStats}}). RTX streams do not show up as separate
               {{RTCOutboundRtpStreamStats}} objects but affect the {{RTCSentRtpStreamStats/packetsSent}},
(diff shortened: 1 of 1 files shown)


henbos:
If this PR lands, I would lobby towards merging the "handler" stats dictionaries since these are no longer used outside of the obsolete section, and the only remaining reference to these things is for the definition of the obsolete RTCMediaStreamTrackStats dictionary.

It would be a lot more clear to the reader if the obsolete section actually contained what is implemented and observed in JavaScript rather than a hierarchy in the spec that does not map to anything observable (in either code or JS)

henbos:
Now that sender and receiver stats are gone we'll be able to flatten the hierarchy... we can just have obsolete "track" stats and get rid of the entire confusing hierarchy that was there for the sake of not having to define "track", "sender" and "receiver" metrics in multiple places...

In the interest of making those editorial changes possible, I'll simply NOT add sender, receiver and transceiver to the provisional spec. Nobody every implemented any of it and if we want to add them back later it will be easier to add them back from scratch without the hierarchy

FYI @alvestrand @vr000m

webrtc-stats.html
```diff
@@ -1654,21 +1642,23 @@ <h2>
...
-                sender of this stream, an {{RTCMediaSourceStats}}.
+                <p>
+                  If the {{RTCRtpTransceiver}} owning this stream has a
+                  {{RTCRtpTransceiver/mid}} value that is not null, this is that
+                  value, otherwise this member is not present.
+                </p>
               </dd>
-              <dt>
-                <dfn>senderId</dfn> of type <span class=
+                <dfn>mediaSourceId</dfn> of type <span class=
```
henbos, line 1655: Missing dt ?

jan-ivar:
This merge includes normative changes lacking agreement. It purports to fix `Editorial` #621, a _"move to ... webrtc-provisional-stats"_), yet `"sender"`, `"receiver"` and `"transceiver"` stats are [not there](https://w3c.github.io/webrtc-provisional-stats/), but gone. That's not "editorial".

This WG has regular weekly Editor's Meetings on Thursdays where PRs are merged or labeled `Editors can integrate` for merging outside of those meetings. This didn't happen ([May 3rd](https://github.com/w3c/webrtc-stats/pull/628#event-6540671983) was a Tuesday). The PR also lacks review.

Instead, a new non-editorial issue should have been opened for visibility, and member participation sought. I've opened https://github.com/w3c/webrtc-stats/issues/643 to have this discussion now.

henbos:
Yes it was a Tuesday because that’s when me, Varun and Harald have the biweekly webrtc stats meeting. This is how the spec has been sheperded for several years. It was reviewed as part of that meeting (Varun was present, Harald was not but had given the thumbs up to do merges while he was gone.)

See https://github.com/w3c/webrtc-stats/pull/628#issuecomment-1151151232 documenting what happened there (which was discussed on Tuesday). But to correct this not being editorial I filed https://github.com/w3c/webrtc-provisional-stats/issues/32 to make sure they get added back to the provisional spec.

The lifetime of the objects is definitely worth following up on but the current status is that sender/receiver/transceiver objects add no new information except for a lifetime statement that has lacked implementer interest since the dictionaries’ conception.

Let’s add them to provisional and continue the lifetime discussion in a separate issue

henbos:
@alvestrand @vr000m

PR #637, opened 2022-06-13
Merge obsolete sender/receiver stats into obsolete track stats.

### LINKED ISSUE #636: Merge obsolete sender/receiver attachment stats into obsolete track stats.
While things are less confusing now thanks to #580, there is still an opportunity to clarify some things.

The whole RTC[Audio/Video][Sender/Receiver]Stats and equally many "[...]AttachmentStats" varieties is pointless these days.
With sender and receivers removed the only dictionary that actually gets created at the end of the day is the "track" stats, which happens to be implemented as a single RTCMediaStreamTrackStats dictionary rather than this hierarchy:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/api/stats/rtcstats_objects.h;l=320;drc=6fb8d1a2d7c158caa1ddb63c9fa9b6a5e2548a43

Let's update the spec to match the implementation by flatting this hierarchy. It's a completely editorial change and only affects an obsolete section but it makes the spec much more readable.

henbos:
Fixes #636.

On the implementation level this has always been called RTCMediaStreamTrackStats. The spec only changed to allow different flavors (sender and receiver stats) but since these were never implemented and have now been removed from the spec, let's restore the spec to match the implementation again.

This makes the spec MUCH more readable since now dictionaries now map closely to RTCStatsType and implementation

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -231,7 +231,7 @@ dictionary RTCStats {
           <li>When a feature is not implemented on the platform, omit the dictionary member that is
           tracking usage of the feature.
           </li>
-          <li>When a feature is not applicable to an instance of an object (for example {{RTCAudioHandlerStats/audioLevel}}
+          <li>When a feature is not applicable to an instance of an object (for example {{RTCInboundRtpStreamStats/audioLevel}}
           on a video stream), omit the dictionary member. Do NOT report a count of zero, -1 or
           "empty string".
           </li>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -513,22 +507,19 @@ <h3>
...
           <dd>
             <p>
-              Statistics related to a specific {{MediaStreamTrack}}'s attachment to an
-              {{RTCRtpSender}} and the corresponding media-level metrics. It is accessed by
-              {{RTCSenderVideoTrackAttachmentStats}},
-              {{RTCSenderAudioTrackAttachmentStats}},
-              {{RTCReceiverVideoTrackAttachmentStats}} or
-              {{RTCReceiverAudioTrackAttachmentStats}}, all
-              inherited from {{RTCMediaHandlerStats}}.
+              This is now obsolete. Statistics related to a specific
```
vr000m, line 510: Ok!

PR #649, opened 2022-07-22
outbound-rtp: add active flag

### LINKED ISSUE #597: outbound-rtp.active
In simulcast, we may turn on or off layers on the fly based on adaptation.
- Should we have a metric for whether or not a layer is currently sending?
- Should we have a metric for if the layer is is active in the API? E.g. if we are not sending that could either be because setParameters() turned it off or it could be because we are unable to send (adaptation turned a layer off or we have disconnected).

fippo:
exposing
  https://w3c.github.io/webrtc-pc/#dom-rtcrtpencodingparameters-active
in stats as well.

Fixes #597

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-stats/pull/649.html" title="Last updated on Jul 22, 2022, 3:24 PM UTC (1eb96f8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/649/dc3120a...fippo:1eb96f8.html" title="Last updated on Jul 22, 2022, 3:24 PM UTC (1eb96f8)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1659,6 +1659,7 @@ enum RTCStatsType {
              unsigned long        firCount;
              unsigned long        pliCount;
              DOMString            encoderImplementation;
+             boolean              active;
 };</pre>
           <section>
             <h2>
@@ -2003,6 +2004,17 @@ enum RTCStatsType {
                 <p class="fingerprint">
                   If too much information is given here, it increases the fingerprint surface.
                   Since it is only given for active tracks, the incremental exposure is small.
+                </p>
+              </dd>
+              <dt>
+                <dfn>active</dfn> of type <span class=
+                "idlMemberType">boolean</span>
+              </dt>
+              <dd>
+                <p>
+                  Indicates whether this <a>RTP stream</a> is configured to be sent or disabled.
+                  Note that an active stream can still not be sending, e.g. when being limited by
+                  network conditions.
                 </p>
               </dd>
             </dl>


webrtc-stats.html
```diff
@@ -2005,6 +2006,15 @@ <h2>
                   Since it is only given for active tracks, the incremental exposure is small.
                 </p>
               </dd>
+              <dt>
+                <dfn>active</dfn> of type <span class=
+                "idlMemberType">boolean</span>
+              </dt>
+              <dd>
+                <p>
+                  Indicates whether this <a>RTP stream</a> is configured to be sent or disabled.
```
henbos, line 2015: Great. Can you just add another sentence like this or similar?
"An RTP stream that is configured to be sent can in some situations still be prevented from sending, such as when being CPU or bandwidth limited."

henbos:
IPR check is failing, what is needed here @fippo @dontcallmedom ? Are you with Microsoft now?

dontcallmedom:
the IPR isn't aware yet of @fippo's new affiliation which is pending some admin handling; no objection from to merge now despite the flag if wanted.

fippo:
we're waiting for some buttons to be pushed on our end which will happen in due time, we can wait.

webrtc-stats.html
henbos: Yes I agree with that I just want to clarify that "active" does not flip flop to false if we are limited

henbos: Just like getParameters()'s active does not return false during BW limitation

webrtc-stats.html
```diff
@@ -2005,6 +2006,17 @@ <h2>
...
+                  Note that an active  stream can still not be sending, e.g. when being limited by
```
henbos, line 2016: Remove double space

henbos (approved):

webrtc-stats.html
fippo: Done, mostly since i'm not sure that `limited` is easy to implement

henbos:
@vr000m @alvestrand in case you have time to take a look. This was previously discussed in #597 and that issue was marked as ready for PR

henbos:
Looks good to you @aboba ?

dontcallmedom:
FWIW IPR bot now happy

vr000m:
Is this implemented or soon to be implemented? To figure out if this is the correct spec, given we moved unimplemented items out.

I will check the metric inline.

webrtc-stats.html
```diff
@@ -2005,6 +2006,17 @@ <h2>
...
                 </p>
               </dd>
+              <dt>
+                <dfn>active</dfn> of type <span class=
+                "idlMemberType">boolean</span>
+              </dt>
+              <dd>
+                <p>
+                  Indicates whether this <a>RTP stream</a> is configured to be sent or disabled.
+                  Note that an active stream can still not be sending, e.g. when being limited by
```
vr000m, line 2016: do you think we will need a counter for the number of times the status flips between `active` and `limited`.

fippo: This flag isn't supposed to flip on its own and only modifiable via setParameters which we don't count.
For limited see #650 but that is partially captured by qualityLimitationDurations.

fippo:
> Is this implemented or soon to be implemented

I have a local chrome build with the implementation already

webrtc-stats.html
henbos: Having a counter for number of times setParameters flips this value makes little sense and is not very useful in my opinion. This metric is useful on its own and in fact very much needed for making sense of simulcast stats. Today you can't really tell "what is the highest resolution layer currently being sent?" or similar questions. We desperately need it for making sense of simulcast :)

henbos: Agree that qualityLimitationDurations is enough, this is more about the effects of the current limitation

PR #669, opened 2022-09-07
Only expose RTCCodecStats objects currently in use

### LINKED ISSUE #662: Don't expose so many RTCCodecStats!
We create an awful lot of `codec` stats objects. Every PT is listed for every m= section. And they are listed for both inbound and outbound direction, so you get all of them twice usually.

In a 50 participant Google Meet call I observed 872 `codec` stats objects, several KB worth of data.

> **Edit:** Turns out most of them are due to a bug, after the fix we're down to 43 `codec` stats objects in this case.

## Performance / bloat issues

This is bad for `getStats()` performance due so many objects being created on the both C++ heap and the V8 heap and all the string copies this entails. Every `DOMString` in the dictionary is copied from an `RTCStatsMember<std::string>` which is copied from lower layers. Copies and object creation does not just happen initially when the report is created, but also every time you iterate the `RTCStatsReport` (which an app likely does a few times). Because of WebIDL dictionaries being passed-by-value, copies and object creation happens every time you iterate, even if you've seen that `codec` before.

The legacy getStats() is more performant than the modern getStats(), and this may help explain why.

This also bloats the report so you typically have to filter them out on manual inspection. `chrome://webrtc-internals/` chooses [not to list them](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/webrtc/resources/stats_table.js;l=35;drc=ccc4b863a4a94811a24a4bbdb150b9b9dddf14c9) because they occlude the page and the rendering is unable to keep up if we include them all.

## This is not even useful

While useful to know what codec you are currently using, it is NOT useful to know every possible combination of PT on every m= section.
- If you wanted to know about codec capabilities, you'd use `RTCRtpSender/RTCRtpReceiver.getCapabilties().codecs`, not `getStats()`.
- If you wanted to inspect the SDP, you'd use `pc.localDescription` or `pc.remoteDescription`, which contains the same information already.

## Proposal

Let's only create `RTCCodecStats` objects representing what is _currently_ in use by `RTCOutboundRtpStreamStats`/`RTCInboundRtpStreamStats`. If it's not referenced by a `codecId`, it doesn't exist.

henbos:
Fixes #662

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-stats/pull/669.html" title="Last updated on Sep 7, 2022, 12:56 PM UTC (0d6a871)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/669/b708dd3...henbos:0d6a871.html" title="Last updated on Sep 7, 2022, 12:56 PM UTC (0d6a871)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -734,10 +734,23 @@ enum RTCStatsType {
           <dfn>RTCCodecStats</dfn> dictionary
         </h3>
         <p>
-          This object is created when a codec type is registered for
-          an RTP transport. It may be referenced by multiple RTP streams
-          in media sections using that transport, but similar codecs in
-          different transports have different RTCCodecStats objects.
+          Codecs are created when registered for an RTP transport,
+          but only the subset of codecs that are in use (referenced by an RTP
+          stream) are exposed in <code>getStats()</code>.
+        </p>
+        <p>
+          The {{RTCCodecStats}} object is created when one or more
+          <code>RTCOutboundRtpStreamStats.codecId</code> or
+          <code>RTCInboundRtpStreamStats.codecId</code> references the codec
+          stats object. When there no longer exists any reference to the
+          {{RTCCodecStats}}, the stats object is deleted. If the same codec is
+          used again in the future, the {{RTCCodecStats}} object is revived with
+          the same {{RTCStats/id}} as before.
+        </p>
+        <p>
+          Codec objects may be referenced by multiple RTP streams in
+          media sections using the same transport, but similar codecs in
+          different transports have different {{RTCCodecStats}} objects.
         </p>
         <p class="note">
           User agents are expected to coalesce information into a single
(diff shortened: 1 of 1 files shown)


henbos:
Can we merge this?

vr000m (approved):
I like the property that the same {{RTCStats/id} is revived but want to confirm that reviving the {{RTCStats/id} is not a burden on the implementation? 

If it is not a burden, I approve the merge.

vr000m:
cc: @jan-ivar WDYT about keeping the same ID?

henbos:
Codecs are created during negotiation so it is trivial to tie codec stats IDs to the monitored codec object, whether or not to expose the codec can be considered a filtering step. But if there are any concerns let's do a follow up

PR #670, opened 2022-09-07
Add powerEfficient[En/De]coder (#666) and fingerprint mitigations (#675).

### LINKED ISSUE #666: powerEfficientEncoder/powerEfficientDecoder
getStats() does not explicitly contain a boolean for whether or not the current encoder/decoder is hardware accelerated.
The question of HW vs SW has raised some concerns in the past, but in many other specs we've usually gone the "isPowerEfficient" path instead, e.g. [powerEfficient](https://www.w3.org/TR/media-capabilities/#dom-mediacapabilitiesinfo-powerefficient) or [powerEfficientPixelFormat](https://w3c.github.io/mediacapture-extensions/#def-constraint-powerEfficientPixelFormat). This frames the question about efficiency instead of hardware, which is the thing you really want to know (but in most cases HW = efficient, SW = inefficient).

Can we add this to RTCInboundRtpStreamStats and RTCOutboundRtpStreamStats?

The lack of this metric has lead some implementations (Chromium) to heuristically include this information in the [encoderImplementation](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-encoderimplementation), so this is to some extent exposed already. @eshrubs

henbos:
Fixes #666, #675

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-stats/pull/670.html" title="Last updated on Sep 27, 2022, 11:05 AM UTC (e983a31)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/670/48530b8...henbos:e983a31.html" title="Last updated on Sep 27, 2022, 11:05 AM UTC (e983a31)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -356,6 +356,39 @@ dictionary RTCStats {
         </p>
       </section>
     </section>
+    <section>
+      <h2>
+        Privacy considerations
+      </h2>
+      <p class="fingerprint">
+        The WebRTC's Statistics API exposes information about the the system,
+        including hardware capabilities and network characteristics. To limit
+        the finger printing surface imposed by this API, some metrics are only
+        exposed if allowed by the algorithms in this section.
+      </p>
+      <section>
+        <h3>
+          Limiting exposure of hardware capabilities
+        </h3>
+        <p>
+          To <dfn data-lt="exposing hardware is allowed">check if hardware
+          exposure is allowed</dfn>, run the following steps:
+          <ol>
+            <li>
+              <p>
+                If the <a href="https://w3c.github.io/mediacapture-main/#context-capturing-state">
+                context capturing state</a> is true, return true.
+              </p>
+            </li>
+            <li>
+              <p>
+                Return false.
+              </p>
+            </li>
+          </ol>
+        </p>
+      </section>
+    </section>
     <section id="rtctatstype-*">
       <h2>
         {{RTCStatsType}}
@@ -1510,15 +1543,12 @@ enum RTCStatsType {
                 "idlMemberType">DOMString</span>
               </dt>
               <dd>
-                <p>
-                  Identifies the decoder implementation used. This is useful for diagnosing
-                  interoperability issues.
-                </p>
                 <p class="fingerprint">
-                  As this field is a fingerprinting vector, it MUST only be
-                  exposed to contexts that the user interacted with in a deep
-                  manner, for instance if <a href="https://w3c.github.io/mediacapture-main/#context-capturing-state">
-                  context capturing state</a> returns true.
+                  Only defined when [= exposing hardware is allowed =].
+                </p>
+                <p>
+                  Identifies the decoder implementation used. This is useful for
+                  diagnosing interoperability issues.
                 </p>
               </dd>
               <dt>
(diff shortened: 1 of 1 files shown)


henbos:
@vr000m PTAL, @eshrubs is ready to implement this right now so we can add it to webrtc-stats rather than provisional

henbos:
Please take another look @youennf, @vr000m and @alvestrand. I've added the sentence from #675 to both powerEfficientEncoder and encoderImplementation (and dec).

henbos:
FYI @eshrubs

webrtc-stats.html
```diff
@@ -1491,8 +1492,29 @@ <h2>
                   interoperability issues.
                 </p>
                 <p class="fingerprint">
-                  If too much information is given here, it increases the fingerprint surface.
-                  Since it is only given for active tracks, the incremental exposure is small.
+                  As this field is a fingerprinting vector, it MUST only be
+                  exposed to contexts that the user interacted with in a deep
+                  manner, for instance if <a href="https://w3c.github.io/mediacapture-main/#context-capturing-state">
+                  context capturing state</a> returns true.
```
henbos, line 1498: This sentence is repeated a few times, let me know if I should move this to a hardware fingerprinting section and point to it to avoid repeated information

vr000m (approved):
LGTM, modulo the language around permissions

henbos:
@youennf Are you happy with the current fingerprinting language?

henbos:
@pes10k what are your thoughts on this PR? By not exposing unless deep user interaction such as the context capturing (requires user to approve a capture prompt) this limits the ability for a web app to use this for fingerprinting to a context where more sensitive information is already exposed (VC use case).

henbos:
Please take a look at the this updated PR with a well defined algorithm for determining whether or not to expose the HW information. @pes10k @youennf

pes10k:
@henbos thank you for this. Just to check my understanding, this would effectively only reveal these endpoints to the page if the user had already given the page permission to access at least one protected device, is that correct?

If so, i think thats a fine mitigation and works well enough to protect users from most realistic passive fingerprinters.

If i've misunderstood though, could you please kindly correct me?

henbos:
> Just to check my understanding, this would effectively only reveal these endpoints to the page if the user had already given the page permission to access at least one protected device, is that correct?

Correct, you would also need to currently be capturing (e.g. microphone or camera). For a page to be granted permissions for capture, the user would need to have accepted a browser permissions prompt.

webrtc-stats.html
```diff
@@ -356,6 +356,39 @@ <h3>
         </p>
       </section>
     </section>
+    <section>
+      <h2>
+        Privacy considerations
+      </h2>
+      <p class="fingerprint">
+        The WebRTC's Statistics API exposes information about the the system,
```
youennf, line 364: s/the the/the/

webrtc-stats.html
```diff
@@ -1528,6 +1561,22 @@ <h2>
...
+              </dt>
+              <dd>
+                <p class="fingerprint">
+                  Only defined when [= exposing hardware is allowed =].
+                </p>
+                <p>
+                  Whether the decoder currently used is considered power
+                  efficient by the user agent. This SHOULD reflect if the
+                  configuration results in hardware acceleration, but the user
+                  agent MAY take other information into account when deciding if
```
youennf, line 1576: Do we have examples of such information? Is it to cover the case of small resolutions for instance where some OSes may decide to go with software implementations

youennf (approved):

webrtc-stats.html
henbos: It gives implementations some wiggle-room, like if a HW implementation isn't performant we could still report false without violating the spec. I don't know if this is a real problem in practise though.

I also think that it is OK for an implementation to say that SW is power efficient if the resolution is so small that SW would be more performant than HW, but I don't think finding that cutoff point is required of the implementation.

pes10k:
Thanks @henbos for the explanation, i think this is a great change

henbos:
👍

webrtc-stats.html
```diff
@@ -356,6 +356,39 @@ <h3>
         </p>
       </section>
     </section>
+    <section>
+      <h2>
+        Privacy considerations
```
alvestrand, line 361: This needs to have a reason to go here rather than in "security and privacy considerations" at the end of the document.

I suggest using the title "Procedures for mitigating privacy concerns" and adding a reference to this section in the "Security and privacy considerations" section.

alvestrand (commented):
I am happy with this doc. Suggesting a refinement.

webrtc-stats.html
henbos: Fixed in follow-up: https://github.com/w3c/webrtc-stats/pull/697

PR #671, opened 2022-09-07
Make RTP creation dependent on SSRC being known rather than packet being sent

### LINKED ISSUE #667: When are RTP stream stats created?
Originally part of issue #643. See also [When are RTP streams destroyed?](https://github.com/w3c/webrtc-stats/issues/668)

Spec currently says when the SSRC is known. Seems to match implementation.

Proposal: Keep it this way

henbos:
Fixes #667.

This matches Chromium, which for example exposes outbound-rtp after setLocalDescription, even if ICE is not connected and packetsSent is 0. I hope all browsers align here but I'm not sure.

**EDIT:** Maybe we should just abandon this and update the implementation, see discussion below

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-stats/pull/671.html" title="Last updated on Sep 7, 2022, 2:26 PM UTC (0669acd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/671/b708dd3...henbos:0669acd.html" title="Last updated on Sep 7, 2022, 2:26 PM UTC (0669acd)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -612,7 +612,8 @@ enum RTCStatsType {
           is known. For send streams, the SSRC is decided when setting the SDP
           offer. For receive streams, the SSRC is typically known after
           receiving the first packet, however if the SSRC is signalled via SDP
-          then the RTP stream is created when the remote SDP is set.
+          then the RTP stream is created when the remote SDP is set containing
+          the SSRC information.
         </p>
         <p>
           RTP monitored objects are not [= deleted =].
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -607,9 +607,12 @@ enum RTCStatsType {
           RTCP SR or RR block.
         </p>
         <p>
-          The lifetime of all RTP [= monitored object =]s starts when the <a>RTP stream</a> is first
-          used: When the first RTP packet is sent or received on the <a>SSRC</a> it represents, or when
-          the first RTCP packet is sent or received that refers to the <a>SSRC</a> of the <a>RTP stream</a>.
+          The lifetime of all RTP [= monitored object =]s starts when the <a>RTP
+          stream</a> is first configured and the required <a>SSRC</a> attribute
+          is known. For send streams, the SSRC is decided when setting the SDP
+          offer. For receive streams, the SSRC is typically known after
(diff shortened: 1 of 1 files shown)


henbos:
Update spec to match Chrome or update Chrome to match spec with regards to RTP object creation?

With the spec being updated, we can expose the SSRC as soon as it is known, which may happen earlier than when packets start flowing.

If we update Chrome instead, we'll have to wait until packets are flowing. Pro: same behavior whether or not SSRCs are signalled. Con: in the event that they do get signalled, waiting to expose until first packet is slightly more racy.

fippo:
You shouldn't create the stats just after creating an offer and setting the local description as we want to avoid the (rare) case that the remote end rejects the transceiver

webrtc-stats.html
```diff
@@ -607,9 +607,13 @@ <h3>
...
-          The lifetime of all RTP [= monitored object =]s starts when the <a>RTP stream</a> is first
-          used: When the first RTP packet is sent or received on the <a>SSRC</a> it represents, or when
-          the first RTCP packet is sent or received that refers to the <a>SSRC</a> of the <a>RTP stream</a>.
+          The lifetime of all RTP [= monitored object =]s starts when the <a>RTP
+          stream</a> is first configured and the required <a>SSRC</a> attribute
+          is known. For send streams, the SSRC is decided when setting the SDP
+          offer. For receive streams, the SSRC is typically known after
+          receiving the first packet, however if the SSRC is signalled via SDP
+          then the RTP stream is created when the remote SDP is set containing
+          the SSRC information.
```
henbos, line 616: Let's abandon this PR and just add an "if packets > 0" check to the implementation? This avoids the dependency on whether or not the SSRC was signalled and it also avoids complexity about "what if the offer was rolled back"

fippo: You create them when you configure them, you destroy them when you rollback?

henbos: @jan-ivar What does Firefox do today in the case that SSRC is signalled in SDP? Does it wait for packets > 0 or does it do what Chrome does and expose the outbound-rtp as soon as the SSRC is known?

henbos:
Based on TPAC discussions I'm abandoning this PR

PR #672, opened 2022-09-07
Delete RTP stream stats when ssrc changes or transceiver stops

### LINKED ISSUE #668: When are RTP streams destroyed?
Originally part of issue https://github.com/w3c/webrtc-stats/issues/643. See also [When are RTP stream stats created?](https://github.com/w3c/webrtc-stats/issues/667)

Spec says never to delete RTP stream stats.

Proposal: Delete RTP stream stats at... RTCP bye? SSRC changes? Receiver is reconfigured?

henbos:
Fixes #668

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-stats/pull/672.html" title="Last updated on Sep 27, 2022, 2:27 PM UTC (fa79e34)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/672/6377f0c...henbos:fa79e34.html" title="Last updated on Sep 27, 2022, 2:27 PM UTC (fa79e34)">Diff</a>


henbos:
PTAL, might be controversial but the intent is to "only expose current stream data" and not require an infinite RTP stream stats history

webrtc-stats.html
```diff
@@ -612,8 +612,48 @@ <h3>
...
+          SSRC=222. Two "outbound-rtp" stats objects are observed (with
+          ssrc=111 and ssrc=222) and both objects' packet counters are
+          increasing.
+
+          One layer is inactivated using RTCRtpSender.setParameters(). While
+          this pauses one of the layers, the RTP monitored objects are not
+          deleted and the packet counter for the inactivated RTP stream reports
+          a non-increasing value.
+
+          Now the RTCRtpTransceiver is negotiated as "inactive". The RTP streams
```
fippo, line 638: This is a hold case. Hold does not that you can forget about the SSRC, you must keep the objects.

henbos: In your view, inactive should be treated the same as active=false regarding RTP lifetime? I.e. as long as the SSRC isn't changing, we don't delete?

henbos:
Updated as suggested (not to delete on "inactive", only "stop"). PTAL

jan-ivar (commented):
Seems a reasonable direction to me. @na-g @pehrsons any concerns?

webrtc-stats.html
```diff
@@ -612,8 +612,52 @@ <h3>
...
+
+          One layer is inactivated using RTCRtpSender.setParameters(). While
+          this pauses one of the layers (its packet counter freezes), the RTP
+          monitored objects are not deleted. The RTCRtpTransceiver is negotiated
+          as "inactive" and the RTP monitored objects are still not deleted.
+          When the RTCRtpTransceiver becomes "sendonly" again, the same
+          "outbound-rtp" objects continue to be used.
+
+          Later, RTCRtpTransceiver.stop() is called and both "outbound-rtp"
+          objects are immediately deleted.
```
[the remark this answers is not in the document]
henbos: While we stop sending and receiving at stop(), it is true we have to negotiate to become stopped. I checked what Chrome does and we do keep the RTP stream stats alive until fully stopped.

I clarified this and updated the example to delay deletion until after renegotiation.

fippo (approved):

webrtc-stats.html
```diff
@@ -612,8 +612,54 @@ <h3>
...
+          such that the {{RTCRtpStreamStats/ssrc}} or
+          {{RTCRtpStreamStats/codecId}} changes or if the
+          {{RTCRtpTransceiver}}'s {{RTCRtpTransceiver/currentDirection}} becomes
+          <code>"stopped"</code>. The monitored object is <i>not</i> deleted if
+          the transceiver is made <code>"inactive"</code> or if the encoding
+          parameters' {{RTCRtpEncodingParameters/active}} attribute is set to
+          <code>false</code>. When an RTP stream is reconfigured such that the
+          monitored object is deleted but the RTP stream continues to be used, a
+          new RTP monitored object and {{RTCStats/id}} will be created for the
+          new SSRC or codec.
```
henbos, line 625: TODO: Reconfiguring to send fewer simulcast layers should be treated as if the codec changed. We have to reset the first layer if the second layer disappears otherwise it can look like the number of packets sent decreased.

fippo: I think I disagree but can explain a bit more please? Probably also explain whether you mean spatial or temporal layer ;-)

henbos: I'm not talking about inactivating a layer, I'm talking about a renegotiation to permanently delete the layer (once it is gone, you cannot revive it). In the spirit of not having to implement caches for old SSRCs, we should delete that outbound-rtp. And if we do that, the only way to make sense of the remaining outbound-rtp layer is to delete and replace it, otherwise "sum of outbound-rtp packets with same mid before negotiation > sum of outbound-rtp packets with same mid after negotiation" which will trick the caller to calculate a negative packet rate. It's easier to think of this weird edge case as a codec reconfiguration.

henbos: I'm talking about simulcast, not SVC. getStats does not support per-SVC layer stats yet, so it only has one outbound-rtp for everything and those counters would always go up as far as I'm concerned. To be revisited when we have per-SVC layer stats.

henbos:
Can we merge this?

webrtc-stats.html
```diff
@@ -612,8 +612,54 @@ <h3>
           the first RTCP packet is sent or received that refers to the <a>SSRC</a> of the <a>RTP stream</a>.
         </p>
         <p>
-          RTP monitored objects are not [= deleted =].
+          RTP monitored objects are deleted when the RTP stream is reconfigured
+          such that the {{RTCRtpStreamStats/ssrc}} or
+          {{RTCRtpStreamStats/codecId}} changes or if the
+          {{RTCRtpTransceiver}}'s {{RTCRtpTransceiver/currentDirection}} becomes
+          <code>"stopped"</code>. The monitored object is <i>not</i> deleted if
```
Pehrsons, line 619: nit: I find that the use of "this RTP stream" in the descriptions of RTCRtpStreamStats/codecId and RTCRtpStreamStats/transportId could be made clearer since "this" (the RTCRtpStreamStats object) and the associated RTP stream now have different lifetimes.

Otherwise this change seems sensible and LGTM.

webrtc-stats.html
```diff
@@ -612,8 +612,54 @@ <h3>
...
+          An RTCRtpReceiver is receiving VP8 on SSRC=111 with PT=96. We observe
+          one "inbound-rtp" with ssrc=111 and a codecId reference to an
+          RTCCodecStats with payloadType=96.
+
+          Without renegotiating, we start receiving VP9 with PT=100 instead, on
+          the same SSRC. The existing "inbound-rtp" is deleted and a new
+          "inbound-rtp" stats object is created. The new "inbound-rtp" has the
+          same ssrc value, but a different stats ID and its packet counter
+          starts from zero. The new object's codecId references a different
+          RTCCodecStats with payloadType=100.
```
jan-ivar, line 661: I know you got this question during the meeting as well, but I didn't understand the answer, and we didn't have a lot of time to go deeper. What's the use case rationale to reset counters on codec change?

I say "reset counters", because that's what apps will experience here. Counters will go backwards. That's because the stats model relies on cross-referencing associations, which in this case is done using the SSRC, not the stats-id, right?

This means apps may not even notice a change in the stats-id. Asking apps to differentiate objects by stats-id uniqueness seems new. These are stats for an ssrc I thought,  not an ssrc+codec pair. Am I wrong?

henbos: > These are stats for an ssrc I thought, not an ssrc+codec pair. Am I wrong?

My reasoning for resetting (i.e. delete + create a replacing stats object) on codec change mainly had to do with "what if a simulcast layer goes away".

The simple case is if a layer is dropped via renegotiation because in this case the layer deletion is explicit in the SDP: if layer A has 10 packets and layer B has 10 packets, that's 20. But if layer B is permanently deleted and only layer A remains, this is now only 10 packets.

I admit **you don't need to** reset layer A here, the app will see that layer B went away so it is already aware that a reconfiguration happened, and should not be surprised that "sum of packets = 10" instead of 20. I mainly considered the reset of stats ID a **good signal** that reconfiguration happened, but we can remove this part of the PR if you prefer? WDYT? No strong opinion here.

In either case, if someone ends up using that SSRC again after this fact, that should be a new stats object, because otherwise the implementation would need to save history indefinitely (memory leak). So I don't consider RTP stats to mean "one SSRC", rather it is "one instance of an SSRC being used".

The other reason for resetting on codec change, and not just layer being dropped, was "what if we switch from simulcast to SVC?". I'm not sure what happens here...
- Let's say you negotiate SSRC 1 and SSRC 2 and you use both. Two outbound-rtp appears.
- Now let's say you change to sending SVC, meaning you only make use of SSRC 1 and not SSRC 2. You're no longer configured to send SSRC 2, but AFAIK the SDP may still reflect simulcast. So does the RTP stream for SSRC 2 exist or not?

My thinking was that deleting on codec change seemed like a good "catch all" to ensure the implementation never has to save history of old SSRCs, but maybe this case is already covered by "reducing layers".

henbos: Should we only delete on SSRC changing, layer dropping or stop(), and not on codec change?

alvestrand:
My opinion:

The case of dropping a simulcast layer is exactly the same as the case of dropping one RTP stream on a transport: If the application sums over a set of RTP senders / receivers, and one of them is dropped, the sum will decrease.

This is no different in the simulcast case, and once we've accepted that totals can decrease at all due to RTP objects going away, we should be consistent and do the same thing with the multiple RTP objects inside a simulcast stream. No special treatment just because they have the same MID.

Add warnings in text, but don't do special things for simulcast.

(A completely different question is if it's possible to remove simulcast layers, or whether the layers not currently accepted by the receiver are just deactivated. Not this bug.)

webrtc-stats.html
henbos: Updated the PR not to delete on codec change, only on SSRC going "away"

webrtc-stats.html
henbos: Updated the language to talk about RTP senders and RTP receivers being the things getting reconfigured. Removed the part about codecId changing.

henbos:
> The case of dropping a simulcast layer is exactly the same as the case of dropping one RTP stream on a transport: If the application sums over a set of RTP senders / receivers, and one of them is dropped, the sum will decrease.
> 
> This is no different in the simulcast case, and once we've accepted that totals can decrease at all due to RTP objects going away, we should be consistent and do the same thing with the multiple RTP objects inside a simulcast stream. No special treatment just because they have the same MID.

I've updated the PTAL not to treat codec changes as special events. PR now says that we only delete when the RTP stream "goes away", i.e. layer drop, ssrc change or transceiver inactivating.

webrtc-stats.html
```diff
@@ -622,8 +622,43 @@ <h3>
...
         <p>
-          RTP monitored objects are not [= deleted =].
+          RTP monitored objects are deleted when the corresponding RTP sender or
+          RTP receiver is reconfigured to remove the corresponding RTP stream.
+          This happens for the old SSRC when the {{RTCRtpStreamStats/ssrc}}
+          changes, a simulcast layer is dropped or the {{RTCRtpTransceiver}}'s
+          {{RTCRtpTransceiver/currentDirection}} becomes <code>"stopped"</code>.
+          The monitored object is <i>not</i> deleted if the transceiver is made
+          <code>"inactive"</code> or if the encoding parameters'
+          {{RTCRtpEncodingParameters/active}} attribute is set to
```
jan-ivar, line 632: Not an attribute
```suggestion
          <code>"inactive"</code> or if the encoding's
          {{RTCRtpEncodingParameters/active}} parameter is set to
```

webrtc-stats.html
```diff
@@ -622,8 +622,43 @@ <h3>
...
+          <code>false</code>. If an SSRC is recycled after a deletion event has
```
jan-ivar, line 633: Good to avoid "event" I think
```suggestion
          <code>false</code>. If an SSRC is recycled after deletion has
```

webrtc-stats.html
```diff
@@ -622,8 +622,43 @@ <h3>
...
+          happened, this is considered a new RTP monitored object and the new 
+          RTP stream stats will have reset counters and a new ID.
         </p>
+        <p>
+          For a given RTP stats object its total counters must always increase,
```
jan-ivar, line 638:
```suggestion
          For a given RTP stats object, its total counters must always increase,
```

webrtc-stats.html
```diff
@@ -622,8 +622,43 @@ <h3>
...
+          but due to changes in SSRC, simulast layers dropping or transceivers
+          stopping, an RTP stats object can be deleted and/or replaced by a new
+          RTP stats object. The caller should be aware of this when aggregating
```
jan-ivar, line 641: Good to avoid "should"
```suggestion
          RTP stats object. The caller will need to be aware of this when aggregating
```

webrtc-stats.html
```diff
@@ -622,8 +622,43 @@ <h3>
...
+          packet counters accross multiple RTP stats objects (the aggregates may
+          decrease due to deletions).
+        </p>
+        <pre class="example highlight">
+          An RTCRtpSender is sending two layers of simulcast on SSRC=111 and
+          SSRC=222. Two "outbound-rtp" stats objects are observed with ssrc=111
+          and ssrc=222. Both objects' packet counters are increasing.
```
jan-ivar, line 648: Without knowing layers and ssrcs are 1-1, I worry this may read like four layers. How about:
```suggestion
          An RTCRtpSender is sending two layers of simulcast (on SSRC=111 and
          SSRC=222). Two "outbound-rtp" stats objects are observed, one with
          ssrc=111, and the other with ssrc=222. Both objects' packet counters are
          increasing.
```
?

jan-ivar (approved):
Looks great! just some commas. Thanks!

henbos:
Thanks @jan-ivar, all suggestions added. @vr000m please merge this PR.

PR #682, opened 2022-09-20
Add RTCAudioPlayoutStats, synthesizedSamplesDuration and totalSamplesDuration

### LINKED ISSUE #676: Need metrics for playout glitches
After audio has been produced (decoding, jitter buffer, concealments etc) it is pulled for playout but the playout client may not be able to provide samples to the device in time causing playout glitches. There is currently no way to measure this in getStats().

When this happens, samples are synthesized. Similar to concealedSamples, in practise in Chrome this would be silence, so silentConcealedSamples.

We need per-playout metrics. How about?
- RTCAudioPlayoutStats.synthesizedSamplesDuration: total duration of synthesized samples during playout.
- RTCAudioPlayoutStats.totalSamplesDuration: the total playout

henbos:
Fixes #676

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-stats/pull/682.html" title="Last updated on Sep 20, 2022, 2:54 PM UTC (85d77f1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/682/072f9dd...henbos:85d77f1.html" title="Last updated on Sep 20, 2022, 2:54 PM UTC (85d77f1)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -948,6 +948,7 @@ enum RTCStatsType {
              double               totalSamplesDuration;
              unsigned long        framesReceived;
              DOMString            decoderImplementation;
+             DOMString            playoutId;
             };</pre>
           <section>
             <h2>
@@ -1503,6 +1504,16 @@ enum RTCStatsType {
                 <p class="fingerprint">
                   If too much information is given here, it increases the fingerprint surface.
                   Since it is only given for active tracks, the incremental exposure is small.
+                </p>
+              </dd>
+              <dt>
+                <dfn>playoutId</dfn> of type <span class=
+                "idlMemberType">DOMString</span>
+              </dt>
+              <dd>
+                <p>
+                  If audio playout is happening, this is used to look up the
+                  corresponding {{RTCAudioPlayoutStats}}.
                 </p>
               </dd>
             </dl>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -2456,6 +2477,55 @@ <h2>
...
         </div>
       </section>
+
+      <section id="playoutstats-dict*">
+        <h3>
+          {{RTCAudioPlayoutStats}} dictionary
+        </h3>
+        <p>
+          Only applicable if the playout path represents an audio devcice.
+          Represents one playout path, which may or may not be after mixing.
```
vr000m, line 2487: ...may not be after mixing the inbound audio streams

(clarifying that this may represent the mixed audio stream and there is no way to know which incoming stream may have caused the issue)

vr000m (approved):
What happens to these metrics if the incoming media pipeline is unavailable and the audio device is trying to pull media from the pipeline?

vr000m:
What happens if there are two media devices that are trying to play a particular audio call on to different devices.

vr000m:
Please add units for the time metrics. I assume these are measured in seconds.

henbos:
> What happens to these metrics if the incoming media pipeline is unavailable and the audio device is trying to pull media from the pipeline?

Added a clarification that synthesization is when underperforming and does not cover inbound RTP jitter buffer metrics.

> What happens if there are two media devices that are trying to play a particular audio call on two different devices.

In practise this is not possible, you may have a virtual device, but it may be that these metrics are not applicable if you attempt to do this.

PR #684, opened 2022-09-20
Add droppedSamplesDuration to RTCAudioSourceStats.

### LINKED ISSUE #678: Need metrics for capture glitches
Similar to #676 but regarding capture rather than playout. We can add these to RTCAudioSourceStats.

If a media source or audio capture pipeline does not consume samples provided from a capture device on time they are dropped.
- RTCAudioSourceStats.totalSamplesDuration: already exists.
- RTCAudioSourceStats.droppedSamplesDuration: new metric. Incremented every time we drop samples in the capture pipeline.

henbos:
Fixes #678.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2272,6 +2272,7 @@ enum RTCStatsType {
               double              totalSamplesDuration;
               double              echoReturnLoss;
               double              echoReturnLossEnhancement;
+              double              droppedSamplesDuration;
 };</pre>
           <section>
             <h2>
@@ -2352,11 +2353,11 @@ enum RTCStatsType {
                 </p>
                 <p>
                   Represents the total duration in seconds of all samples that have been produced
-                  by this source for the lifetime of this stats object. Can be used with
+                  by this source (excluding {{droppedSamplesDuration}}) for the lifetime of this stats object. Can be used with
                   {{totalAudioEnergy}} to compute an average audio level over
                   different intervals.
                 </p>
-                
+
               </dd>
               <dt>
                 <dfn>echoReturnLoss</dfn> of type <span class=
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -2388,6 +2389,17 @@ <h2>
...
+              <dt>
+                <dfn>droppedSamplesDuration</dfn> of type <span class=
+                "idlMemberType">double</span>
+              </dt>
+              <dd>
+                <p>
+                  Only applicable if this media source is backed by an audio
+                  capture device. The total duration of samples produced by the
+                  device that got dropped before reaching the media source.
+                </p>
```
vr000m, line 2401: Please add units for the time metrics. I assume these are measured in seconds.

vr000m (approved):
We already have framesDropped for Video, it makes sense to have a similar metric for Audio

webrtc-stats.html
```diff
@@ -2352,11 +2353,11 @@ <h2>
                 </p>
                 <p>
                   Represents the total duration in seconds of all samples that have been produced
-                  by this source for the lifetime of this stats object. Can be used with
+                  by this source (excluding {{droppedSamplesDuration}}) for the lifetime of this stats object. Can be used with
```
vr000m, line 2356: We should clarify that this is on the capture, the droppedSamples while have a consequence on the audioEnergy that is transmitted, it is not measured here.

vr000m: cc: @jan-ivar to confirm Firefox implementation.

henbos: According to WPT, Firefox has not implemented media-source, so I think we just need to clarify that we should not measure energy of dropped samples

PR #685, opened 2022-09-20
Add totalCaptureDelay and totalSamplesCaptured to RTCAudioSourceStats.

### LINKED ISSUE #681: Metrics for capture delay
Similar to #680 but for capture.

We can measure the time between audio being captured by the device and it reaching the media source which then emits them to the next part of the pipeline (including WebRTC). Let's add...
- RTCAudioSourceStats.totalCaptureDelay: delay between capture and being emitted, incremented for each sample.
- RTCAudioSourceStats.totalSamplesCount: total samples emitted from the source.

henbos:
Fixes #681

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-stats/pull/685.html" title="Last updated on Sep 20, 2022, 3:11 PM UTC (145be7a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/685/a2b7577...henbos:145be7a.html" title="Last updated on Sep 20, 2022, 3:11 PM UTC (145be7a)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2447,23 +2447,25 @@ enum RTCStatsType {
               <dd>
                 <p>
                   Only applicable if the audio source represents an audio
-                  capture device. This is the total delay for each audio sample
-                  between the time the sample was emitted by the capture device
-                  and the sample reaching the source. This can be used together
-                  with {{totalSamplesCaptured}} to calculate the average capture
-                  delay per sample.
+                  capture device. This is the total delay, in seconds, for each
+                  audio sample between the time the sample was emitted by the
+                  capture device and the sample reaching the source. This can be
+                  used together with {{totalSamplesCaptured}} to calculate the
+                  average capture delay per sample.
                 </p>
               </dd>
               <dt>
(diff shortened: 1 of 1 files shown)


vr000m:
Please add units for the time metrics. I assume these are measured in seconds.

webrtc-stats.html
```diff
@@ -2388,6 +2390,32 @@ <h2>
...
+              <dt>
+                <dfn>totalSamplesCaptured</dfn> of type <span class=
+                "idlMemberType">unsigned long long</span>
+              </dt>
+              <dd>
+                <p>
+                  Only applicable if the audio source represents an audio
+                  capture device. This is the total number of captured samples
+                  reaching the audio source, i.e. that were not dropped by the
+                  capture pipeline.
```
henbos, line 2416: Frequency here does not need to be the same as encoder FYI

vr000m:
What happens if the source is stereo? We should be consistent with other metrics.

Issue #679, opened 2022-09-20
How many times did capture glitches occur? (Follow-up to #678)


henbos:
Follow-up to #678.

If we can measure duration of capture gliches, can we also measure number of times capture glitches occurred (as discrete events)?
- RTCAudioSourceStats.droppedSamplesEvents: consecutive drops of samples is just one event

henbos:
#688 has merged

PR #688, opened 2022-09-21
Add droppedSamplesEvents and synthesizedSamplesEvents

### LINKED ISSUE #677: How many times did playout glitches occur? (Follow-up to #676)
Follow-up to #676.

If we can measure synthesized duration during playout it may also be interesting to know about how many synthesization events happen. In addition to synthesizedSamplesDuration and totalSamplesDuration (#676), can we also add?
- RTCAudioPlayoutStats.synthesizationEvents: number of discrete synthesization events

henbos:
Fixes #677, #679.

Because sample drops and sample synthesization tend to happen in bursts it is useful to not only know the duration but also the number of "events" happened.

The definition for these are very similar to [concealmentEvents](https://w3c.github.io/webrtc-stats/#dom-rtcinboundrtpstreamstats-concealmentevents).

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-stats/pull/688.html" title="Last updated on Sep 21, 2022, 9:20 AM UTC (391b21d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/688/8ae8a4c...henbos:391b21d.html" title="Last updated on Sep 21, 2022, 9:20 AM UTC (391b21d)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2306,6 +2306,7 @@ enum RTCStatsType {
               double              echoReturnLoss;
               double              echoReturnLossEnhancement;
               double              droppedSamplesDuration;
+              unsigned long       droppedSamplesEvents;
               double              totalCaptureDelay;
               unsigned long long  totalSamplesCaptured;
 };</pre>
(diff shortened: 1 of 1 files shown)


vr000m:
Thanks for adding these metrics, they would be useful to calculate frequency of events!

PR #694, opened 2022-09-26
inbound-rtp: add frame assembly time

fippo:
which allows measuring the average time to receive an inbound video frame.

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-stats/pull/694.html" title="Last updated on Oct 4, 2022, 7:55 PM UTC (b9fbefd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/694/55d60dd...fippo:b9fbefd.html" title="Last updated on Oct 4, 2022, 7:55 PM UTC (b9fbefd)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1023,8 +1023,8 @@ enum RTCStatsType {
              DOMString            decoderImplementation;
              DOMString            playoutId;
              boolean              powerEfficientDecoder;
-             double               totalAssemblyTime;
              unsigned long        framesAssembledFromMultiplePackets;
+             double               totalAssemblyTime;
             };</pre>
           <section>
             <h2>
@@ -1665,6 +1665,17 @@ enum RTCStatsType {
                 </p>
               </dd>
               <dt>
+                <dfn>framesAssembledFromMultiplePackets</dfn> of type <span class=
+                "idlMemberType">unsigned long</span>
+              </dt>
+              <dd>
+                <p>
+                  Only [= map/exist =]s for video.  It represents the total number of frames correctly decoded
+                  for this RTP stream that consist of more than one RTP packet. For such frames the
+                  {{totalAssemblyTime}} is incremented.
+                </p>
+              </dd>
+              <dt>
                 <dfn>totalAssemblyTime</dfn> of type <span class=
                 "idlMemberType">double</span>
               </dt>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1577,6 +1579,33 @@ <h2>
...
               </dd>
+              <dt>
+                <dfn>totalAssemblyTime</dfn> of type <span class=
+                "idlMemberType">double</span>
+              </dt>
+              <dd>
+                <p>
+                  Only [= map/exist =]s for video. The sum of the time, in seconds, each video frame takes
+                  from the time the first RTP packet is received (reception timestamp) and to the time
+                  the last RTP packet of a frame is received.
```
vr000m, line 1590: Does the last RTP packet require a market bit to be set? or what happens if the packet with marker bit is lost? i.e., would it wait for the next RTP timestamp to arrive?

Maybe another way of asking, the timer starts when the first RTP packet with a new timestamp arrives. The timer stops when the RTP packet with the same timestamp and the marker bit set is received OR when an RTP packet with a different timestamp is received.

webrtc-stats.html
```diff
@@ -1577,6 +1579,33 @@ <h2>
...
+                </p>
+                <p>
+                  Given the complexities involved, the time of arrival or the reception timestamp is measured as close to the network layer as possible.
+                  This metric is not incremented for frames that are not decoded, i.e., framesDropped, partialFramesLost or frames that fail decoding
+                  for other reasons (if any). Only incremented for frames consisting of more than one RTP packet. The average frame assembly time can
+                  be calculated by dividing the totalAssemblyTime with framesAssembledFromMultiplePackets.
```
vr000m, line 1596: Do we need `framesAssembledFromMultiplePackets` wouldnt framesDecoded suffice?

alvestrand: If majority of frames fit into 1 packet, frameAssemblyTime will be zero, and those drown out the assembly-time signal.

henbos, line 1596: Can you make the lines shorter to avoid these line breaks?

vr000m: Got the answer -- not all the framesDecoded are counted in the assembly time

webrtc-stats.html
```diff
@@ -1577,6 +1579,33 @@ <h2>
...
+                </p>
+              </dd>
+              <dt>
+                <dfn>framesAssembledFromMultiplePackets</dfn> of type <span class=
+                "idlMemberType">unsigned long</span>
+              </dt>
+              <dd>
+                <p>
+                  Only [= map/exist =]s for video.  It represents the total number of frames correctly decoded for this RTP stream that consist of
+                  more than one RTP packet. For such frames the totalAssemblyTime is incremented.
```
henbos, line 1606: {{totalAssemblyTime}}

webrtc-stats.html
henbos, line 1590: Only applicable to frames which consist of multiple packets?

webrtc-stats.html
```diff
@@ -1577,6 +1579,33 @@ <h2>
...
+              </dt>
+              <dd>
+                <p>
+                  Only [= map/exist =]s for video. The sum of the time, in seconds, each video frame takes
+                  from the time the first RTP packet is received (reception timestamp) and to the time
+                  the last RTP packet of a frame is received.
+                </p>
+                <p>
+                  Given the complexities involved, the time of arrival or the reception timestamp is measured as close to the network layer as possible.
+                  This metric is not incremented for frames that are not decoded, i.e., framesDropped, partialFramesLost or frames that fail decoding
```
henbos, line 1594: {{framesDropped}}, {{partialFramesLost}}

webrtc-stats.html
vr000m: from discussion. Incomplete frames do not count towards this metric

webrtc-stats.html
henbos: {{totalAssemblyTime}}, {{framesAssembledFromMultiplePackets}}

henbos: No because framesDecoded is after decoding and also includes frames that were single packet

webrtc-stats.html
```diff
@@ -995,6 +995,8 @@ <h3>
              DOMString            decoderImplementation;
              DOMString            playoutId;
              boolean              powerEfficientDecoder;
+             double               totalAssemblyTime;
```
alvestrand, line 998: Should this be totalFrameAssembly Time?
Possibly this reads better if it follows the framesAssembledFromMultiplePackets definition.

vr000m:
Could you reorder the metrics, the caveats are in the counter definition. Which makes sense to come first

alvestrand:
Seems we agree on the goal, let's fiddle a little more and then merge it.

henbos:
Address comments and we can review again :) overall sounds good

vr000m (approved):
modulo comments

webrtc-stats.html
fippo: copied text from below after moving around which makes it a bit better

webrtc-stats.html
fippo: in practice this only happens with 320x180 thumbnails but it does happen sadly so we needed the separate counter.

fippo:
this reads better with the count coming first indeed (which raises the question if that is true for other counters as well)

Oddly preview doesn't seem to show the latest commit which moved the instructions how to calculate

PR #702, opened 2022-10-03
Move framesDropped to inbound-rtp.

### LINKED ISSUE #661: Move RTCRtpReceivedRtpStreamStats.framesDropped to RTCInboundRtpStreamStats.
It's only implemented in inbound-rtp, [not remote-inbound-rtp](https://wpt.fyi/results/webrtc-stats/supported-stats.html?label=experimental&label=master&aligned). Let's just move it?

henbos:
Fixes #661.

Only implemented for inbound-rtp, not remote-inbound-rtp.

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -921,7 +921,6 @@ enum RTCStatsType {
              unsigned long long   packetsReceived;
              long long            packetsLost;
              double               jitter;
-             unsigned long        framesDropped;
 };</pre>
           <section>
             <h2>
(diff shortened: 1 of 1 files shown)


alvestrand (approved):

vr000m:
@henbos could you make a corresponding change in the provisional stats? 
 
> I think in the extensions, it should reside in the `RTCRtpReceivedRtpStreamStats`, in case it is ever implemented as an XR block, it would be in the correct location.

vr000m (approved):
LGTM, could we add framesDropped to provisional stats

PR #709, opened 2022-11-29
Add RTCOutboundRtpStreamStats.scalabilityMode

### LINKED ISSUE #673: Adding SVC-related stats fields
From https://github.com/w3c/webrtc-svc/issues/72

It would be nice to have SVC-related fields in WebRTC stats.

For example, adding the DOMString scalabilityMode field of RTCOutboundRtpStreamStats to indicate which scalability mode is used by the current encoding would be helpful for debugging SVC-related issues.

henbos:
Fixes #673.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-stats/pull/709.html" title="Last updated on Nov 29, 2022, 10:32 AM UTC (33f66cb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/709/4da9296...henbos:33f66cb.html" title="Last updated on Nov 29, 2022, 10:32 AM UTC (33f66cb)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2252,8 +2252,8 @@ enum RTCStatsType {
                 <p>
                   The currently configured
                   <a href="https://w3c.github.io/webrtc-svc/#scalabilitymodes*">
-                  scalability mode</a> of the for this <a>RTP stream</a>. If one
-                  is not configured, this metric is left undefined.
+                  scalability mode</a> of this <a>RTP stream</a>. If one is not
+                  configured, this metric is left undefined.
                 </p>
               </dd>
             </dl>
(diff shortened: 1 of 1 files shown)


henbos:
Evan is able to provide an implementation

vr000m (approved):
LGTM

henbos:
> This is only on the send side, is there a similar property for the inbound/receive side? it could be useful if the SFU ends up dropping layers

Because of layer dropping the receiver only knows about a subset of the overall SVC layers and can't really tell what the "true" scalabilityMode was used on the send side. Could we still expose something useful? I don't know, but if so it would be a separate issue.

alvestrand:
The receiver doesn't know, and probably couldn't do anything with the info even if it knew - it will have to take the frames given to it and try to do something sensible. Since the mode isn't negotiated, there's no guarantee the sender is sticking to a specific mode - it might do a "custom" mode, even.

PR #717, opened 2022-12-13
Add framesRendered, update totalInterFrameDelay.

### LINKED ISSUE #716: Align inter-frame delay metrics with pause/freeze metrics by measuring at render time, not decode time
The inter-frame delay metrics are used to calculate the harmonic frame rate.

Historically, inter-frame delay had both an "at decode" and "at render" implementation in webrtc, but today's "at decode" are the only ones that exist because they are the one in the spec. But this proved to be an issue, because the harmonic fps at decode time is not closest to what the user experiences.

We already have "at render" metrics pauseCount/totalPausesDuration and freezeCount/totalFreezesDuration. We should make the inter-frame delay metrics align as was originally intended to also be "at render".

To resolve this issue:
1. Update totalInterFrameDelay and totalSquaredInterFrameDelay to talk be updated when a frame is rendered, not when it is decoded.
2. Add a framesRendered counter, since the fps variance is calculated using a frame counter, and frame rendering happens later than frame decoding.

henbos:
Fixes #716.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-stats/pull/717.html" title="Last updated on Dec 13, 2022, 10:51 AM UTC (be9918b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/717/d0dc3aa...henbos:be9918b.html" title="Last updated on Dec 13, 2022, 10:51 AM UTC (be9918b)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -988,6 +988,7 @@ enum RTCStatsType {
              DOMString            remoteId;
              unsigned long        framesDecoded;
              unsigned long        keyFramesDecoded;
+             unsigned long        framesRendered;
              unsigned long        framesDropped;
              unsigned long        frameWidth;
              unsigned long        frameHeight;
@@ -1099,6 +1100,15 @@ enum RTCStatsType {
                 </p>
               </dd>
               <dt>
+                <dfn>framesRendered</dfn>
+              </dt>
+              <dd>
+                <p>
+                  Only [= map/exist =]s for video. It represents the total number of frames that have been
+                  rendered. It is incremented just after a frame has been rendered.
+                </p>
+              </dd>
+              <dt>
                 <dfn>framesDropped</dfn> of type <span class="idlMemberType">unsigned
                 long</span>
               </dt>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1098,6 +1099,15 @@ <h2>
                   you the number of delta frames decoded.
                 </p>
               </dd>
+              <dt>
+                <dfn>framesRendered</dfn>
+              </dt>
+              <dd>
+                <p>
+                  Only [= map/exist =]s for video. It represents the total number of frames that have been
+                  rendered. It is incremented just after a frame has been rendered.
```
alvestrand, line 1108: I think it would be good to link to a definition of "rendered" - this may have to go into the terminology section.

alvestrand (approved):

webrtc-stats.html
henbos: I agree that this should be clarified, but because this language is already used in several places, I filed #718 as a follow-up to make sure we get all of them in one issue

alvestrand:
This part seems to be uncontroversial - landing so as not to leave this hanging.

Issue #724, opened 2023-01-11
Add RTCInboundRtpStreamStats:decoderFallback


xingri:
After [#712](https://github.com/w3c/webrtc-stats/issues/712) has been merged,  it is no longer possible to get the decoder fallback information in case of cloud gaming services which do not require hardware access permission in general use cases.

And the decoder fallback information is highly important to assess the session information whether the session has been fully accelerated by the hardware decoder or not. So I would like to propose adding decoderFallback in RTCInboundRtpStreamStats to get the information without hurting the security concern.
- https://github.com/w3c/webrtc-stats/commit/d5c96ad30313055f48f1600f5e2e840d5cf0e2ba

If the first proposal is not acceptable, as an alternative, I would also like to propose to remove fingerprint restriction on powerEfficientDecoder in RTCInboundRtpStreamStats.
- https://github.com/w3c/webrtc-stats/commit/01e32ce879c0cc4687771453c761efaf7674ca9f

My justification behind the second approach is that the powerEfficientDecoder can be modifiable when the decoder is getting fallback to the software decoder, so it is not a static variable to represent a user agent.

fippo:
@henbos ^
I've been pondering exposing the number of times this fallback happened which might be a bit more insightful

henbos:
I'm wondering if a boolean would suffice for "the selected codec does have HW but it's currently not used due to fallback to SW being active". So powerEfficient is false but softwareFallback is true.

I think a boolean is more helpful than a counter because a counter talks about the past rather than what is currently the case, and I'm not sure there is much value in counting the number of times an attempt is made, which seems like an implementation detail that could be different on different browsers. E.g. one browser may try to re-boot the HW after a while but another browser may stop using it for the rest of the session... the caller of getStats shouldn't care.

@eshrubs do you have any thoughts or opinions about adding SW fallback information to getStats? I think it would be a good addition to powerEfficient to have this.

> So I would like to propose adding decoderFallback in RTCInboundRtpStreamStats to get the information without hurting the security concern.

Sadly I disagree that this is not a privacy/security concern. This metric exposes that HW was available, so It also needs to be gated. I think what you want @xingri is more conditions to the "exposing hardware is allowed" check

eshrubs:
[...]

Issue #712, opened 2022-11-30
Fullscreen shouldn't be enough to reveal fingerprinting-surface


pes10k:
Currently the spec has two gates to exposing high-resolution / high-granularity fingerprinting surface [capturing or fullscreen](https://w3c.github.io/webrtc-stats/#limiting-exposure-of-hardware-capabilities).

Capturing seems like an appropriate guard since the user has already likely gone through some browser UI flow, and indicated a pretty high level of trust with the site.

Having a `fullscreenElement` though doesn't indicate any such level of trust, and is not a sufficient guard against fingerprinting.  This is especially true since the element that has fullscreen might be only loosely related to the page itself, and might not indicate any level of trust with any frame on the document (think, for example, a clickbate site hosting a youtube video, or any other `<video>`, etc).

Further, many (most?) fullscreen elements have nothing to do with WebRTC

I suggest the spec remove the `fullscreenElement` option in the guard (i.e. remove bullet 2 from [Section 6.1.](https://w3c.github.io/webrtc-stats/#limiting-exposure-of-hardware-capabilities))

henbos:
I don't mind removing this, unless someone objects. @aboba would you care if it was no longer possible to read HW information from getStats (e.g. powerEfficientDecoder, decoderImplementation) in the use case of being in fullscreen and not having GUM access?

henbos:
The gate was originally added as to not break Cloud Gaming use cases, but it is a pretty nieche use case to do fullscreenw without GUM and there is already MediaCapabilities that should give accurate enough information

pes10k:
Terrific, thank you all!

PR #725, opened 2023-01-12
Add RTCInboundRtpStreamStats:decoderFallback

xingri:
Fixes https://github.com/w3c/webrtc-stats/issues/724, https://github.com/w3c/webrtc-stats/issues/730

***
<a href="https://pr-preview.s3.amazonaws.com/xingri/webrtc-stats/pull/725.html" title="Last updated on Mar 10, 2023, 7:08 AM UTC (1efea03)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/725/3057aa7...xingri:1efea03.html" title="Last updated on Mar 10, 2023, 7:08 AM UTC (1efea03)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1032,6 +1032,7 @@ enum RTCStatsType {
              boolean              powerEfficientDecoder;
              unsigned long        framesAssembledFromMultiplePackets;
              double               totalAssemblyTime;
+             boolean              decoderFallback;
             };</pre>
           <section>
             <h2>
@@ -1710,6 +1711,16 @@ enum RTCStatsType {
                   as close to the network layer as possible. This metric is not incremented for frames that
                   are not decoded, i.e., {{framesDropped}} or frames that fail decoding for other reasons
                   (if any). Only incremented for frames consisting of more than one RTP packet.
+                </p>
+              </dd>
+              <dt>
+                <dfn>decoderFallback</dfn> of type <span class=
+                "idlMemberType">boolean</span>
+              </dt>
+              <dd>
+                <p>
+                  Only [= map/exist =]s for video. It indicates that the current WebRTC session is fallbacked
+		  to the Software Decoder. When the session has started with Hardware Decoder.
                 </p>
               </dd>
             </dl>


webrtc-stats.html
```diff
@@ -1709,6 +1710,16 @@ <h2>
...
                 </p>
               </dd>
+              <dt>
+                <dfn>decoderFallback</dfn> of type <span class=
+                "idlMemberType">boolean</span>
+              </dt>
+              <dd>
+                <p>
+                  Only [= map/exist =]s for video. It indicates that the current WebRTC session is fallbacked
+		  to the Software Decoder. When the session has started with Hardware Decoder.
```
henbos, line 1720: Since this also exposes HW information, I wonder if we need to use the same gate as for powerEfficientDecoder, or if this is limited enough in scope that it is okay not to...

```
                <p class="fingerprint">
                  Only defined when [= exposing hardware is allowed =].
                </p>
```

Is the idea that you would use MediaCapabilities to query HW support and then use decoderFallback to detect when HW is not achieved anymore?

xingri: @henbos  thank you for the review. That's right. We would like to detect decoderFallback when hardware decoder get fallback. I believe this does not need the fingerprint gate because it does not expose the hardware information explicitly like other metrics like powerEfficiency and decoderImplementation since the decoder fallback information might be dynamic information based on the runtime status of the decoder even the browser is capable of hardware decoding.

fippo: If you see a decoderFallback being true in the stats that tells you that there was HW at some point. You'll need to trigger this fallback somehow which seems tricky but possible e.g. with insertable streams.

We really need to answer the question how use-cases like gamestreaming which we are supposed to support can get access to this.

xingri: @fippo & @henbos  Thank you for reviewing the changes. I agree that new metric could expose the HW capability at some instance of the time. 

However I would like to note that what the fingerprint guidance concerns is not only the hardware exposer of the Browser(User Agent) but risk of the fingerprinting of the browser user agent as defined by <https://w3c.github.io/fingerprinting-guidance/> as follows:

_**..., browser fingerprinting is the capability of a site to identify or re-identify a visiting user, user agent or device via configuration settings or other observable characteristics.**_

To share the transitions of this metric, it may look like following:
- In case of no HW Decoder:  stays false
- In case of HW Decoder: starts with false and turns to true only when the decoder get fallback and changes to true again when the decoder gets recovered to HW.
 
So the current metric could not be used for identifying the user, user agent or device because of ephemeral characteristics of decoder fallback. 

And when we consider the cloud gaming's general use case, the camera and microphone permission are not required in general because it works as one-way streaming from the gaming server to the browser client, so it does not require user permission for the hardware access.

And we could detect any streaming abnormalities through the decoder fallback from decoder implementation metric and it is currently blocked by the fingerprint concern by [#712](https://github.com/w3c/webrtc-stats/issues/712).

That's why I proposed this new metric which might not expose the finger print of the user agent directly like decoder implementation and power efficiency metrics.

Diego-Perez-Botero: @fippo @henbos @xingri - Adding to this conversation from the perspective of Xbox Cloud Gaming. Right now, we care about decoderImplementation to be able to detect SW decode fallback cases, so that we can trigger resolution downgrade logic and/or identify regressions in devices that we know should support HW decode. For instance, we've hit instances where a not-as-important "splash screen" asset using the same codec as our game stream would sometimes prevent the actual game stream from being able to acquire the HW decoder, and the diagnosis/monitoring of that problem was only possible thanks to the decoderImplementation info from getStats.

As we all know, the decoderImplementation info from WebRTC getStats() will now require getUserMedia() permissions, which is a no-go for most cloud gaming scenarios, as @xingri has already pointed out. Xbox Cloud Gaming only asks for getUserMedia() in the Voice Chat scenario. We cannot ask for getUserMedia() permissions in all game streams because it's invasive (prompts users), overkill (gives us access to the device camera/mic hardware when we don't always need it), and has a high chance of failing (users have no reason to allow it unless they want to use Voice Chat). This proposed decoderFallback property satisfies our functional requirements of being able to reliably detect SW vs HW decode, while no longer exposing detailed device-specific information that could be categorized as fingerprintable.

We understand that MediaCapabilities APIs allow us to query the HW/SW capabilities of the client device, but there's no guarantee that those capabilities will take effect during the actual game stream. Xbox Cloud Gaming and similar scenarios care more about what actually happened during the stream (was HW decode actually used?) than what *may* happen during the stream (do we have a chance that HW decode will be used?), since we need that data point to make mid-stream decisions. This is why having decoderFallback be part of the RTCInboundRtpStreamStats makes sense to us. Without this, our best bet is to monitor decode times (via requestVideoFrameCallback) and have heuristics around what SW and HW decode should look like for a particular device, which is error-prone and is arguably even more of a fingerprinting behavior.

henbos: > So the current metric could not be used for identifying the user, user agent or device because of ephemeral characteristics of decoder fallback.

There might be a way to consistently trigger fallback, but agree this is a much smaller concern than the "always expose HW" problem.

> We cannot ask for getUserMedia() permissions

Agreed. As the spec is currently written I'd say we are not solving your use case.

> We understand that MediaCapabilities APIs allow us to query the HW/SW capabilities of the client device, but there's no guarantee that those capabilities will take effect during the actual game stream.

The privacy inconsistency where one spec exposes HW and another spec doesn't is frustrating. I don't have a good answer to that, which is why I think we should file a separate issue to discuss the HW implications and solve them in a standalone issue. But until we have solved those, I think the conservative thing to do is to add this metric with the same HW check as the other HW metrics.

henbos: Not trying to shut down the privacy discussion, just trying to move it to a standalone issue

henbos: I filed https://github.com/w3c/webrtc-stats/issues/730 to follow up on the discussion

xingri: @henbos. Thank you for the feedback and consideration of the issue. 
As the chrome 110 will be released tomorrow, the cloud game services will be officially impacted by the hardware exposure check on decoderImplementation metric updated by the recent commit of https://github.com/w3c/webrtc-stats/issues/712.

Also, if the new metric will have the same limitation as decoderImplementation, it will not  make any improvements on the current situation based on the difficulties of getUserMedia() permissions  as @Diego-Perez-Botero  and I described.

So I am wondering whether we have a timeline for the discussion of hardware exposure checking <https://github.com/w3c/webrtc-stats/issues/730>? 
If we could get the decision in a timely manner, that would be really appreciated.

henbos:
Let's follow up on this after #730 has been resolved

webrtc-stats.html
henbos: See comments in #730 and PR #732

xingri:
@henbos If you are agreeing with the current approach, could you approve the pull request?

henbos:
I think this sort of thing needs WG input (_especially_ after #746). From where I stand, I think exposing this bit of information is fine, but I don't think I can approve without WG consultation

xingri:
> I think this sort of thing needs WG input (_especially_ after #746). From where I stand, I think exposing this bit of information is fine, but I don't think I can approve without WG consultation

@henbos  Thank you for the update. Will wait for the WG's decision. If the decision will take time longer than next Tuesday monthly meeting, could we bring this item on the monthly meeting as you presented last month?

xingri:
> > could we bring this item on the monthly meeting as you presented last month?
> 
> Sorry I've not had time for this, very busy week and it's currently not on [the agenda](https://docs.google.com/presentation/d/1OfAN3sfVXhvOBCqgLMegrNcHz7-V4qTnuHjBP3gLdgg/edit#slide=id.g1f91c40f700_0_5).
> 
> @aboba is this something you'd be interested in presenting?

@henbos  I understand. Thanks you for the reply. 
@aboba As Henrik already mentioned, if there is a way to update the agenda by including this item, I would happy to prepare it.

henbos:
Please request edit access on the slides link

xingri:
> Please request edit access on the slides link

Thank you Henrik, I have requested.

xingri:
@henbos  As we getting positive feedback from today's call. Could you please advise me for the next steps?

xingri:
@henbos  I am still waiting for the video recording and meeting notes but as far ar I understood, the working group members were agreed upon the current approach. Could you please approve the current PR if that is correct understanding?

henbos:
I think we should take the discussion to https://github.com/w3c/webrtc-extensions/issues

xingri:
> @xingri My understanding was that the WG was not in support of exposing this in getStats(). I did sense support for wanting to solve your use case, but my take was that...
> 
> * This should be an event rather than a getStats() metric.
> * We should try to make sure the event is not useful for correlating HW usage with different websites, but for example if the encoder failed because you asked for an unsupported scalabilityMode, the event could say that without any cross-site issues.

@henbos thank you again for the feedback. May be I was so optimistic about the discussion on the last meeting. And I agree with your suggestion. Let me follow up on https://github.com/w3c/webrtc-extensions/issues/146.

PR #732, opened 2023-02-07
Expose powerEfficientEncoder/Decoder if MediaCapabilities allows it

### LINKED ISSUE #730: The HW exposure check does not solve Cloud Gaming use cases
After some back and forth, the [current HW exposure check](https://w3c.github.io/webrtc-stats/#limiting-exposure-of-hardware-capabilities) only considers contexts that have capture (getUserMedia). This [obviously does not work for Cloud Gaming](https://github.com/w3c/webrtc-stats/pull/725#discussion_r1093134014).

What's even more frustrating is that MediaCapabilities already expose the HW capability information, but it does not expose what is currently used (e.g. powerEfficientDecoder or decoderFallback). The specs are inconsistent in how much they care about exposure information. The [MC privacy consideration section](https://www.w3.org/TR/media-capabilities/#decoding-encoding-fingerprinting) says that there is very little information exposed so it is not a big problem, but that a user agent might want to throttle or give vague answers, which is very much up for interpretation. Why would getStats() be less restrictive than MC?

Possible paths forward:
1. Declare decoderFallback a weak fingerprinting vector and skip the HW exposure check for that particular metric (but not other metrics like powerEfficientDecoder).
2. Reference MediaCapablities.powerEfficient. Even if the steps are something vague like "let the UA decide", we would at least have something to reference, and we could say that "if the MC check passes, return true" in our own HW exposure algorithm.
3. Other suggestions welcome?

In reality though, because MC is currently up to the UA which in practise appears to be implemented as "always expose", doing 2) would basically turn off all finger printing protection from getStats(). But at least the two specs would be consistent and we would have a placeholder algorithm (in MC spec) to update when necessary.

Let's discuss

henbos:
Fixes #730.

This exposes powerEfficientEncoder and powerEfficientDecoder (but not encoderImplementation or decoderImplementation) in contexts that are not capturing but where MediaCapabilities' powerEfficient already allows exposure.

This should not significantly increase fingerprinting surface beyond what MediaCapabilities is already doing and if MC becomes more restrictive than so would getStats.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-stats/pull/732.html" title="Last updated on Feb 7, 2023, 3:15 PM UTC (cb5197c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/732/2ada0c9...henbos:cb5197c.html" title="Last updated on Feb 7, 2023, 3:15 PM UTC (cb5197c)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -373,17 +373,27 @@ dictionary RTCStats {
         </h3>
         <p>
           To avoid passive fingerprinting, hardware capabilities should only be
-          exposed in capturing contexts. This is tested using the algorithm
-          below.
+          exposed in capturing contexts or contexts where Media Capabilities
+          permit hardware exposure. This is tested using the algorithm below.
         </p>
         <p>
           To <dfn data-lt="exposing hardware is allowed">check if hardware
-          exposure is allowed</dfn>, run the following steps:
+          exposure is allowed</dfn>, optionally taking a
+          <var>codec configuration</var> as an argument, run the following
+          steps:
           <ol>
             <li>
               <p>
                 If the <a href="https://w3c.github.io/mediacapture-main/#context-capturing-state">
                 context capturing state</a> is true, return true.
+              </p>
+            </li>
+            <li>
+              <p>
+                If a <var>codec configuration</var> was passed to this
+                algorithm and <a href="https://w3c.github.io/media-capabilities/#dom-mediacapabilitiesinfo-powerefficient">
+                Media Capabilities' powerEfficient</a> is allowed to be exposed
+                for this configuration and context, return true.
               </p>
             </li>
             <li>
@@ -1670,7 +1680,9 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p class="fingerprint">
-                  Only defined when [= exposing hardware is allowed =].
+                  Only defined when [= exposing hardware is allowed =] with
+                  <var>codec configuration</var> being set to the current codec
+                  configuration.
                 </p>
                 <p>
                   Whether the decoder currently used is considered power
@@ -2226,7 +2238,9 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p class="fingerprint">
-                  Only defined when [= exposing hardware is allowed =].
+                  Only defined when [= exposing hardware is allowed =] with
+                  <var>codec configuration</var> being set to the current codec
+                  configuration.
                 </p>
                 <p>
                   Whether the encoder currently used is considered power


webrtc-stats.html
```diff
@@ -373,19 +373,29 @@ <h3>
...
                 If the <a href="https://w3c.github.io/mediacapture-main/#context-capturing-state">
                 context capturing state</a> is true, return true.
               </p>
             </li>
+            <li>
+              <p>
+                If a <var>codec configuration</var> was passed to this
+                algorithm and <a href="https://w3c.github.io/media-capabilities/#dom-mediacapabilitiesinfo-powerefficient">
+                Media Capabilities' powerEfficient</a> is allowed to be exposed
+                for this configuration and context, return true.
```
henbos, line 396: Is it clear that we're piggybacking on the exposure steps specifically, and not necessarily implying that the codecs must be the same for any given config? E.g. powerEfficient could return true in MC but due to SW fallback powerEfficient is false in getStats()

henbos:
Going for [option 2](https://github.com/w3c/webrtc-stats/issues/730#issue-1567732462). @xingri does this PR make sense to you? And if you do have powerEfficientDecoder as a boolean in the Cloud Gaming use case, is the fallback metric still of interest?

xingri:
> Going for [option 2](https://github.com/w3c/webrtc-stats/issues/730#issue-1567732462). @xingri does this PR make sense to you? And if you do have powerEfficientDecoder as a boolean in the Cloud Gaming use case, is the fallback metric still of interest?

@henbos It make sense and we do not need the fallback metric as long as we gets the same information through the powerEfficientDecoder.

vr000m:
this looks good to me.

vr000m (approved):

webrtc-stats.html
alvestrand: I think it is clear that we're asking for "allowed to be exposed", and not the actual value.

alvestrand:
I'd like to check with @pes10k that it's reasonable to use this criterion for both the power efficient boolean and the implementation string. The first one is obvious to me, the second less so.

aboba:
FYI, I'm thinking of creating a PR to expose the same info in WebCodecs. 

Related: https://github.com/w3c/webcodecs/issues/604

henbos:
Interesting that it keeps coming up. I think when multiple specs expose ~the same information for ~the same reason it would be great if we could all point to a single exposure step, in this case referring to MC, to avoid inconsistencies.

pes10k:
@alvestrand thank you for the ping.  I apologize though, im finding this discussions very difficult to follow so bare with me. 

My understanding is:
1. Currently, a site can only access a device's `RTCInboundRtpStreamStats.powerEfficientDecoder` and `RTCInboundRtpStreamStats.powerEfficientEncoder` properties if _any_ `getUserMedia`-gated device is [currently playing](https://w3c.github.io/mediacapture-main/#context-capturing-state)
2. The change being discussed here would change things, so that the above two properties are also available anytime _any devices_ `MediaCapabilitiesInfo.powerEfficient` is available

Is the above correct? If so, I have two questions:

1. I was surprised by the "any" bits in the above, but i don't see anything in the [exposing hardware check](https://w3c.github.io/webrtc-stats/#dfn-exposing-hardware-is-allowed) that ties the properties of the device being described with the device playing). Is this intentional?
2. I'm not able to find any algorithm or description of when `MediaCapabilitiesInfo.powerEfficient` is made available. Am I missing it? If so, can you kindly point me to it?  If it doesn't exist, then the change in this PR is worrying, since it would change a useful, well defined privacy protection, into a privacy protection that was loosely or (un-)defined

Apologies if I'm missing obvious things, just, a bit difficult to make sure i have the entire picture right, since it spans 2-3 specs/repos

xingri:
> Is the above correct? If so, I have two questions:
> 
> 1. I was surprised by the "any" bits in the above, but i don't see anything in the [exposing hardware check](https://w3c.github.io/webrtc-stats/#dfn-exposing-hardware-is-allowed) that ties the properties of the device being described with the device playing). Is this intentional?
> 2. I'm not able to find any algorithm or description of when `MediaCapabilitiesInfo.powerEfficient` is made available. Am I missing it? If so, can you kindly point me to it?  If it doesn't exist, then the change in this PR is worrying, since it would change a useful, well defined privacy protection, into a privacy protection that was loosely or (un-)defined

@pes10k, thank you for the feedback.

As far as I understood, your understanding is correct.  And as the following sentences as @henbos mentioned at the begining of this issue. We thought this couldn't worsen the security concern beyond the level of MediaCapabilities exposure.

_This exposes powerEfficientEncoder and powerEfficientDecoder (but not encoderImplementation or decoderImplementation) in contexts .... This should not significantly increase fingerprinting surface beyond what MediaCapabilities is already doing and if MC becomes more restrictive than so would getStats._

However, if you are still concerning on this level of exposure. Could you please share your opinion about exposing the decoderFallback as an alternative as I proposed on https://github.com/w3c/webrtc-stats/issues/724?

My opinion is decoderFallback metric obscures the hardware exposure because it is only getting enabled when there is a hardware decoder fallback gets intiated. And then it will be gone when it is recovered.

henbos:
> My understanding is: ...

Your understanding is correct but one clarification is needed in your use of the word "any".

It's not "if MC's powerEfficient is exposed for any device, period" rather it is "if MC's powerEfficient is exposed for the same codec, resolution, frame rate and scalability mode that WebRTC is using right now". 99% of the time the device queried in MC is the same device that is in use by WebRTC and getStats's powerEfficientDecoder exposes the _same bit of information_. But 1% of the time HW fails resulting in WebRTC falling back to a SW implementation for the same codec configuration. Basically:
- MC's powerEfficient already allows querying for HW device support before even using the HW.
- getStats' powerEfficient is initially expected to be the same as MC's powerEfficient, but if HW error happens during the use of the device, SW fallback can happen, changing this bit.

Simply put, if you want to query device support you use MC with any configuration you like and you can make repeated queries for a range of different configurations. If you want to query current operational status of a WebRTC session, you use getStats for the current configuration.

Using getStats to query support in a scenario where MC is already available is technically possible by reconfiguring the encoder/decoder, but is backwards and jumping through unnecessary hoops when MC already exposes this information for you without the additional overhead and API complexity.

>  If [well-defined MC gating steps] doesn't exist, then the change in this PR is worrying, since it would change a useful, well defined privacy protection, into a privacy protection that was loosely or (un-)defined.

Correct. MC's [fingerprinting note](https://w3c.github.io/media-capabilities/#decoding-encoding-fingerprinting) is not a well-defined algorithm. We may want to file an issue on the MC spec to clarify this.

But there are problems too with having getStats be more restrictive than MC (breaks Cloud Gaming) and it is also a problem if multiple specs try to solve how to expose the same bit of information differently. We should be consistent amongst specs and to me it makes more sense if getStats references the HW exposure spec than the other way around.

henbos:
Ping @pes10k

xingri:
@henbos, @alvestrand  Are you planning to attend WG February 2023 call schedule on 21 February 2023, 16:00 - 18:00 UTC?
If we couldn't reach the decision, could we discuss this issue over the call?

henbos:
> @henbos, @alvestrand Are you planning to attend WG February 2023 call schedule on 21 February 2023, 16:00 - 18:00 UTC?
> If we couldn't reach the decision, could we discuss this issue over the call?

I'll be there. I created [a couple of slides](https://docs.google.com/presentation/d/1t9Ijbfjt8c1QAC7G5MrTGjwLzo92VS00ZlwBROAcufc/edit#slide=id.g20f28c9366c_15_1) for this issue for the interim.

xingri:
> > @henbos, @alvestrand Are you planning to attend WG February 2023 call schedule on 21 February 2023, 16:00 - 18:00 UTC?
> > If we couldn't reach the decision, could we discuss this issue over the call?
> 
> I'll be there. I created [a couple of slides](https://docs.google.com/presentation/d/1t9Ijbfjt8c1QAC7G5MrTGjwLzo92VS00ZlwBROAcufc/edit#slide=id.g20f28c9366c_15_1) for this issue for the interim.

Thank you so much @henbos. See you in the meeting.

xingri:
> > > @henbos, @alvestrand Are you planning to attend WG February 2023 call schedule on 21 February 2023, 16:00 - 18:00 UTC?
> > > If we couldn't reach the decision, could we discuss this issue over the call?
> > 
> > 
> > I'll be there. I created [a couple of slides](https://docs.google.com/presentation/d/1t9Ijbfjt8c1QAC7G5MrTGjwLzo92VS00ZlwBROAcufc/edit#slide=id.g20f28c9366c_15_1) for this issue for the interim.
> 
> Thank you so much @henbos. See you in the meeting.

@henbos  thank you so much for presenting this issue on the meeting.
Please let me know if you need any follow ups from my end for supporting the PR.

henbos:
My feeling was "yes, we want a go-to place to refer to about finding out whether or not we can expose stuff like this" but that there was no consensus on piggybacking on MC's powerEfficient for now, in particular there was a concern that this is exposing more information since current usage may influence availability.

Other browser editors did say that they would find it convincing if we could create a demo for heuristically detecting SW fallback based on other metrics. I imagine we could add an experimental flag to force SW fallback after X number of frames and do a demo where we plot performance metrics. If we see a sharp decline, we might come up with a formula "if such and such goes down by such and such".

But since this is being discussed in https://github.com/w3c/webcodecs/issues/604 already, I'd wait for @aboba for the next steps

alvestrand:
Add an interface under chrome.test (nonstandard stuff) that forces the source of a track into software rendering mode?
Demo pages, unlike WPTs, can use non-standard functionality.

My optimistic read of the meeting was that the other browser vendors wanted to wait until the Media WG had had the chance to review the issue (a week from now?) before supporting the "let MC own it" approach.

xingri:
> Add an interface under chrome.test (nonstandard stuff) that forces the source of a track into software rendering mode? Demo pages, unlike WPTs, can use non-standard functionality.
> 
> My optimistic read of the meeting was that the other browser vendors wanted to wait until the Media WG had had the chance to review the issue (a week from now?) before supporting the "let MC own it" approach.

@henbos, @alvestrand  Thank you for the updated feedback. Would be the correct understanding if I could prove that software fallback could be identifiable by only using the MC interface, that would be more convincing to other browser vendors?

henbos:
MC does not detect SW fallback since it gives an optimistic view of capabilities rather than what is currently used, but perhaps other getStats() metrics could such as a sudden change in delta totalEncodeTime / delta framesEncoded despite resolution and frame rate being the same?

pes10k:
@henbos i take your point and agree, it would be good to have a similar, common, shared algorithm different-but-similar specs to hook into, to make sure there is a common gate for exposing FP-related information.

I think, though, that it'd be a serious privacy loss/harm to remove the mostly well defined (*) algoritm thats in place, and instead point to a currently undefined algorithm, and hope that the pointed-to-algorithm becomes stronger.

Could we invert things, and keep the getStats algo as is, and work with the folks defining `MediaCapabilitiesInfo.powerEfficient` to have it point to the `getStats` algo, and unify things that way?

Or, if better, spin up a third/distinct doc that defines the "when and what can the page learn about user hardware at different states of user interaction" algorithm(s) and have getStats, MediaCapabilitiesInfo, and anyone else point to that?

---

* I just said "mostly" bc I re-read the text and it wasn't 100% unambiguous that the exist getStat's algo would require the clarification you provided in https://github.com/w3c/webrtc-stats/pull/732#issuecomment-1426657623 (though I appreciate that could easily be ignorance on my end)

henbos:
@pes10k The current algoritm is too aggressive because it breaks Cloud Gaming. Updating all the specs to point to it would certaily make things more well-defined, but that breaks that use case in the other specs too. We want to support the use case of streaming video and knowing about HW usage. So the way I see it, we have to modify the current algoritm, the question is how?

What change would you propose that satisfies the use case?

pes10k:
> What change would you propose that satisfies the use case if not to piggyback on MC already exposing it?

Im not trying to be flip, but im not sure i have an answer at the moment. My role is just to identify things that would be privacy harming to the platform.  I appreciate that seems like a small issue on your end, but if even (handwave) 1/4 of reviewed specs / changes added this much fingerprinting surface it would be catastrophic privacy wise, so its really important to not add FP surface.

Just to throw a straw proposal on the table though, why can't the cloud gaming folks just prompt the users? Would a permission-approach be possible? Cloud gaming uses may already be prompting users for WebUSB or WebBluetooth or other permissions

xingri:
> Just to throw a straw proposal on the table though, why can't the cloud gaming folks just prompt the users? Would a permission-approach be possible? Cloud gaming uses may already be prompting users for WebUSB or WebBluetooth or other permissions

@pes10k  thank you for the comments. To share the current status of the Nvidia Cloud Gaming, as I shared through https://github.com/w3c/webrtc-stats/pull/725, we do not request any additional permissions other than what is necesasray for the streaming in general usecase. Some use cases require a microphone but most of the users do not use the microphone.
I believe the same scenario may apply for the Xbox Cloud Gaming as Diego shared on the same issue.

Also, I understand your concern and I am willing to participate in the discussion about the approaches which could protect both of the user privacy and the usability of the Cloud Gaming as a next step.

xingri:
> > its really important to not add FP surface
> 
> @pes10k In your opinion, is exposing powerEfficientDecoder adding FP surface _in a context that already has MC's powerEfficient_? The difference between the two is so subtle that I would argue that hiding powerEfficientDecoder _in this context_ does not really improve the FP situation compared to exposing it. And if it does not, this pull request is approximately FP neutral. In which case we're hurting a real use case without improving the FP situtation.
> 
> > Just to throw a straw proposal on the table though, why can't the cloud gaming folks just prompt the users?
> 
> I'm open to the idea that other APIs prompt (if it makes sense for those APIs) and that getStats() piggybacks on the permissions of the current context, but getStats() should not prompt since this is a passive API being polled in the background to get quality metrics and not relating to a user action. There are 200+ metrics so the app could poll this API for any reason, so asking for HW permission as a result of someone wanting to know the bitrate of their data channel does not make sense.
> 
> > we do not request any additional permissions other than what is necesasray for the streaming in general usecase
> 
> @xingri Is it the case that the normal use case of no microphone has 0 prompts? Or is there any other prompt that we could piggyback on?

@henbos There are two usecases for the "normal use" and "first use", when the "first use" use case, the Nvidia Cloud Game requests the clipboard and the microphone prompts but the decisions on the prompts do not impact the "normal use".  And once the decision has chosen, there are no additional prompts for "the normal" use case. So in case of "normal" use case, it has 0 prompt.
firstuse-prompt-1-clipboard
firstuse-prompt-2-microphone

aboba:
FYI, am proposing to expose an event in WebCodecs providing similar information: https://github.com/w3c/webcodecs/pull/645

pes10k:
@henbos 

> In your opinion, is exposing powerEfficientDecoder adding FP surface in a context that already has MC's powerEfficient

No, i don't think so, at least as far as I understand the two properties. But we have a lot of legacy privacy harm in the platform, and so im extremely uncomfortable with new specs reducing the privacy protections they have because they exist in legacy / older features.

In other words, i think the right outcome to this mismatch is to bring the protections around MC's `powerEfficient` up to par with `getStats()`, and to leave the terrific and well defined algoritm around `getStats()` as a model for other specs to align on, instead of removing or weakening a well defined protection to align it with a undefined one.

> There are 200+ metrics so the app could poll this API for any reason, so asking for HW permission as a result of someone wanting to know the bitrate of their data channel does not make sense.

Totally understood and agreed.  There is a lot behind `getStats()` that has little or not affect on user privacy, and bugging users w/ a prompt for those details would be all harm and no benefit. You could imagine a `getStats(sensitive=false)` call that returns the non-sensitive values, and a `getStats(sensitive=true)` one that prompts (again just a straw suggestion).

xingri:
> No, i don't think so, at least as far as I understand the two properties. But we have a lot of legacy privacy harm in the platform, and so im extremely uncomfortable with new specs reducing the privacy protections they have because they exist in legacy / older features.

@pes10k  Could you also share your opinion on the [option1 ](https://github.com/w3c/webrtc-stats/issues/730#issue-1567732462) which brings a new metric as an alternative approach?

pes10k:
@xingri im not sure i understand [option1](https://github.com/w3c/webrtc-stats/issues/730#issue-1567732462) fully.

1. I can't find a definition or anything called `decoderFallback` in the [spec](https://w3c.github.io/webrtc-stats/). Where can i learn more?
2. Under option1, what happens with `powerEfficientDecoder` and other non-weak fingerprinting vectors?

xingri:
> @xingri im not sure i understand [option1](https://github.com/w3c/webrtc-stats/issues/730#issue-1567732462) fully.
> 
> 1. I can't find a definition or anything called `decoderFallback` in the [spec](https://w3c.github.io/webrtc-stats/). Where can i learn more?
> 2. Under option1, what happens with `powerEfficientDecoder` and other non-weak fingerprinting vectors?

FYI, added follow up information on https://github.com/w3c/webrtc-stats/issues/730.

henbos:
Let's close this, events are being discussed elsewhere

PR #739, opened 2023-03-03
Clarify "exists" and "only defined for"

### LINKED ISSUE #737: Confusing use of the terms defined and undefined
The terms _"defined"_, _"undefined"_, the JS value `undefined`, and member present/omitted/[exists](https://infra.spec.whatwg.org/#map-exists) are different concepts.

This spec appears to be using these terms interchangeably in some places. For example, [encoderImplementation](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-encoderimplementation) says:
- _"Only defined when [exposing hardware is allowed](https://w3c.github.io/webrtc-stats/#dfn-exposing-hardware-is-allowed). ... Only [exist](https://infra.spec.whatwg.org/#map-exists)s for video."_

Here the term [exists](https://infra.spec.whatwg.org/#map-exists) is well-defined, but "defined" is not (no pun intended). There's no way to tell what it means, but I suspect what was meant was:
- "Only [exists](https://infra.spec.whatwg.org/#map-exists) for video, and only when [exposing hardware is allowed](https://w3c.github.io/webrtc-stats/#dfn-exposing-hardware-is-allowed)."

See https://github.com/w3c/webrtc-stats/issues/736#issuecomment-1452556494 for how the concepts differ.

Action items:
1. I count 4 hits for _"Only defined"_, which should probably be _"Only [exists](https://infra.spec.whatwg.org/#map-exists)"_ (plural seems better)
2. Most other hits for "defined" seem to be correct (references to where something is _specified_) but I could have missed some.
3. The 14 hits for "undefined" are more suspect. I see phrases like _"Calculated as defined in section 6.4.1. of [[RFC3550](https://w3c.github.io/webrtc-stats/#bib-rfc3550)]. If no [RTCP Receiver Report](https://w3c.github.io/webrtc-stats/#dfn-receiver-report) is received with a DLSR value other than 0, the round trip time is left undefined"_

In the last one, _"defined"_ seems to mean _specified_, whereas _"left undefined"_ I suspect means omitted.

fippo:
fixes #737 (there is more)

Please review individual commits, pay particular attention to the one about DLSR things

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-stats/pull/739.html" title="Last updated on Mar 7, 2023, 2:13 PM UTC (23768a4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/739/e148089...fippo:23768a4.html" title="Last updated on Mar 7, 2023, 2:13 PM UTC (23768a4)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1083,7 +1083,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Only defined for video. It represents the total number of frames correctly decoded
+                  Does not [= map/exist =] for audio. It represents the total number of frames correctly decoded
                   for this <a>RTP stream</a>, i.e., frames that would be displayed if no frames are dropped.
                 </p>
               </dd>
@@ -1093,7 +1093,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Only defined for video. It represents the total number of key frames, such as key
+                  Does not [= map/exist =] for audio. It represents the total number of key frames, such as key
                   frames in VP8 [[RFC6386]] or IDR-frames in H.264 [[RFC6184]], successfully
                   decoded for this RTP media stream. This is a subset of
                   {{framesDecoded}}. <code>framesDecoded - keyFramesDecoded</code> gives
@@ -1105,7 +1105,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Only defined for video. It represents the total number of frames that have been
+                  Does not [= map/exist =] for audio. It represents the total number of frames that have been
                   rendered. It is incremented just after a frame has been rendered.
                 </p>
               </dd>
@@ -1115,7 +1115,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Only defined for video. The total number of frames dropped prior to decode or dropped
+                  Does not [= map/exist =] for audio. The total number of frames dropped prior to decode or dropped
                   because the frame missed its display deadline for this receiver's track. The measurement
                   begins when the receiver is created and is a cumulative metric as defined in
                   Appendix A (g) of [[!RFC7004]].
@@ -1127,7 +1127,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Only defined for video. Represents the width of the last decoded frame. Before the
+                  Does not [= map/exist =] for audio. Represents the width of the last decoded frame. Before the
                   first frame is decoded this member does not [= map/exist =].
                 </p>
               </dd>
@@ -1137,7 +1137,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Only defined for video. Represents the height of the last decoded frame. Before
+                  Does not [= map/exist =] for audio. Represents the height of the last decoded frame. Before
                   the first frame is decoded this member does not [= map/exist =].
                 </p>
               </dd>
@@ -1147,7 +1147,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Only defined for video. The number of decoded frames in the last second.
+                  Does not [= map/exist =] for audio. The number of decoded frames in the last second.
                 </p>
               </dd>
               <dt>
@@ -1156,7 +1156,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Only defined for video. The sum of the QP values of frames decoded by this
+                  Does not [= map/exist =] for audio. The sum of the QP values of frames decoded by this
                   receiver. The count of frames is in {{framesDecoded}}.
 
                 </p>
@@ -1332,7 +1332,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Only defined for video. Count the total number of Full Intra Request (FIR) packets,
+                  Does not [= map/exist =] for audio. Count the total number of Full Intra Request (FIR) packets,
                   as defined in [[!RFC5104]] section 4.3.1, sent by this receiver. Does not count the RTCP FIR
                   indicated in [[?RFC2032]] which was deprecated by [[?RFC4587]].
                 </p>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1083,7 +1083,7 @@ <h2>
               </dt>
               <dd>
                 <p>
-                  Only [= map/exist =]s for video. It represents the total number of frames correctly decoded
+                  Only defined for video. It represents the total number of frames correctly decoded
```
henbos, line 1086: Doesn't not being defined for audio imply that it is undefined?
What was wrong [= map/exists =]?

fippo: We don't have a definition of framesDecoded for audio which by virtue of "we can only have things which are defined" means this can not exist on audio.

`only defined` is a bit stronger than `only exists`. For example, encoderImplementation only exists on video but only after a frame was encoded.

One can also say the opposite, "Never exists for audio" but... that is [so negative](https://www.youtube.com/watch?v=dQw4w9WgXcQ)

webrtc-stats.html
```diff
@@ -1791,7 +1791,8 @@ <h2>
               <dd>
                 <p>
                   Represents the total number of <a>RTCP RR</a> blocks received for this <a>SSRC</a> that contain
-                  a valid round trip time. This counter will not increment if the {{roundTripTime}} is undefined.
+                  a valid round trip time. This counter will not increment if the {{roundTripTime}} is can not
+                  be calculated because no <a>RTCP Receiver Report</a> with a DLSR value other than has been received.
```
vr000m, line 1795: Other than 0 has been received?

webrtc-stats.html
```diff
@@ -3622,14 +3624,16 @@ <h2>
                   measurement does not count the size of the IP or other transport layers like TCP
                   or UDP. It is similar to the TIAS defined in [[!RFC3890]], i.e., it is measured
                   in bits per second and the bitrate is calculated over a 1 second window.
+
+                  For candidate pairs in use, the estimate is normally no lower than the
```
vr000m, line 3628: Is this paragraph requiring that the 1s interval be calculated as a sliding window of 1s that ends at the lastPacketTimestamp?

We should further clarify this aspect

vr000m (commented):
Will re-review, left some quick comments.

webrtc-stats.html
fuzzy-coats, line 1795: Would it be clearer to say something along the lines of: "This counter will remain 0 until a <a>RTCP Receiver Report</a> with a non-zero DSLR has been received."?

fippo: Agree that this whole things needs a bit of clarification, it tries to explain how RTCP RTT calculation works but does so in each definition instead of summarzing the RFC 3550 appendix in a single note.

(but that is for a separate change)

webrtc-stats.html
henbos: I don't want us to use "only defined" language because of the above confusing. I think the metric is "defined" for audio - the definition is: "if audio, this attribute does not exist".

If you really want to avoid the word exist you can say "only applicable to", but I don't see any downsides with being clear about it not existing, since this is in fact different from existing and having the value undefined

henbos: Much better to say "only exists for video" or "does not exist for audio". Let's say what we mean and not something that is up for interpretation

fippo: I think this reduces the searchability but with the latest commit one can still search for "Does not exist for"

webrtc-stats.html
```diff
@@ -1890,9 +1891,9 @@ <h2>
               </dt>
               <dd>
                 <p>
-                  If the {{RTCRtpTransceiver}} owning this stream has a
-                  {{RTCRtpTransceiver/mid}} value that is not null, this is that
-                  value, otherwise this member is not present.
+                  Only [= map/exist =]s if the {{RTCRtpTransceiver}} owning this stream has a
+                  {{RTCRtpTransceiver/mid}} value that is not null., this is that
```
henbos, line 1895: nit: .,

webrtc-stats.html
```diff
@@ -1920,10 +1921,11 @@ <h2>
               </dt>
               <dd>
                 <p>
-                  Exposes the {{RTCRtpCodingParameters/rid}}
-                  encoding parameter of this <a>RTP stream</a> if it has been set, otherwise it is
-                  undefined. If set this value will be present regardless if the RID RTP header
-                  extension has been negotiated.
+                  Does not [= map/exist =] for audio.
```
henbos, line 1924: nit: Can you remove the newline here and adjust this paragraph to not exceed the per-line char limit?

webrtc-stats.html
```diff
@@ -2213,10 +2215,10 @@ <h2>
               </dt>
               <dd>
                 <p class="fingerprint">
-                  Only defined when [= exposing hardware is allowed =].
+                  Omitted unless [= exposing hardware is allowed =].
```
henbos, line 2218: Here and other places that use "Omitted": I prefer "Does not [= map/exist =] unless [= exposing hardware is allowed =]." for the same reasons as I preferred "Does not [= map/exist =] for audio."

henbos (approved):
LGTM % comments

webrtc-stats.html
```diff
@@ -2253,10 +2255,9 @@ <h2>
               </dt>
               <dd>
                 <p>
-                  The currently configured
+                  Does not [= map/exist =] for audio. Only [= map/exist =]s when a
                   <a href="https://w3c.github.io/webrtc-svc/#scalabilitymodes*">
-                  scalability mode</a> of this <a>RTP stream</a>. If one is not
-                  configured or not applicable, this metric is left undefined.
+                  scalability mode</a> is currently configured of this <a>RTP stream</a>.
```
vr000m, line 2260:
```suggestion
                  scalability mode</a> is currently configured for this <a>RTP stream</a>.
```

vr000m (approved):
lgtm

Issue #746, opened 2023-03-09
What is the relationship between WebRTC stats and the Working Group?


henbos:
It was my understanding that 6-7 years ago, the WebRTC stats editors was asked to lead the stats work on behalf of the WG.
I've been committed to this work since at least February, 2017 and we've come a long way in our goal to achieve [browser-compatible stats](https://wpt.fyi/results/webrtc-stats/supported-stats.https.html?label=experimental&label=master&aligned), finally being able to [deprecate legacy getStats](https://groups.google.com/a/chromium.org/g/blink-dev/c/PxQQtEM7za0).

This was primarily achieved through bi-weekly meetings between me, @vr000m and @alvestrand. The WG has been involved regarding the [mandatory to implement](https://w3c.github.io/webrtc-pc/#mandatory-to-implement-stats) list, WebRTC stats has frequently been presented at TPAC ([example](https://docs.google.com/presentation/d/1Xprz0ElNTUWRDdX6kyp11cZ4ItTvUFUmzoXsKZdXxuU/edit#slide=id.g41c8f53d0b_1_286)) and occasionally WebRTC stats has been featured at Virtual Interims to widen the audience.

But the bulk of the work has happened in the bi-weekly stats meetings and metrics defined there, while not mandatory to implement unless the WG says so, have been considered ready to implement and ship, and the libwebrtc implementation has slowly evolved over many years based on this process. So I was quite surprised about [this comment](https://github.com/w3c/webrtc-stats/issues/742#issuecomment-1460380778) by @jan-ivar. The full comment is worth reading, but here is a quote from it:

> My understanding of the role of editors includes discretion to craft and merge editorial PRs, maybe PRs for normative bug fixes where WG opinion can be inferred from earlier discussion, and PRs for issues that have been discussed in the WG that the WG agrees to proceed with.

According to this view, we should not have done much of anything without asking the WG. While the WG has been involved numerous times, stats work being editorial has not been practised since the inception of getStats(), though not all stats have been considered mandatory to implement either.

What should the process be going forward? @jan-ivar @aboba @youennf @alvestrand @vr000m

henbos:
@dontcallmedom do you remember when auto publishing was turned on or how that decision was made?

dontcallmedom:
[...]

Issue #742, opened 2023-03-07
Assorted comments on `RTCAudioPlayoutStats`


padenot:
`playoutId` is frequently not implementable, there can be more than one playout path, or none, because a `MediaStreamTrack` can be sent to multiple `AudioContext` and/or `HTMLMediaElement` for rendering.

There can also be thing that look like playout paths that are not, e.g., piping a inbound `MediaStreamTrack` to an `AudioContext` and just sending the data to an `AnalyzerNode` or an `AudioWorkletNode` to compute some metrics or simply record the output (e.g. using a Web Worker and Web Codecs, or whatever else).

An `AudioContext` can also be connected to another `AudioContext`, that are on different devices. The actual output can be on the first, second or both. This is another case where `playoutId` cannot be implemented.

All the other metrics are reimplementations of  Web Audio API's [Audio Render Capacity](https://webaudio.github.io/web-audio-api/#audiorendercapacity) and latency metrics.

Only application authors know the shape of their audio output path, and therefore they only can determine a quality metric, based on number provided by the object they actually use to render audio.

jan-ivar:
It appears this was merged fairly recently in https://github.com/w3c/webrtc-stats/pull/682. Was it presented to the WG? I don't recall any WG discussion on it, nor could I locate any mention of it in minutes. Perhaps it shouldn't have been merged?

henbos:
> It appears this was merged fairly recently in https://github.com/w3c/webrtc-stats/pull/682. Was it presented to the WG? I don't recall any WG discussion on it, nor could I locate any mention of it in minutes. Perhaps it shouldn't have been merged?

This was not presented to the WG, it was presented on the bi-weekly stats meeting. In your opinion, what changes do or do not require WG interaction? We've had bi-weekly stats editors meetings for years (since the inception of the getStats API), but as of recently there have been a couple of questions around WG involvement. I don't want to give off the impression that we're going behind the WG's back.

Should WG presentation be required any time we add a new stats type? Any time it is "large"? Or do you want to get involved in the bi-weeklies as well?

henbos:
We should clarify what to do if there are multiple playout paths. Having multiple IDs is one option, but if this is as niche use case as I think it is, I would prefer if the playout can be represented as a single object.

henbos:
[...]

Issue #741, opened 2023-03-07
Feedback on audio capture stats


padenot:
Being able to know the input latency is a welcome change, but there are a number of problems with the prose:

`totalSamplesCaptured`: this should be named `totalFramesCaptured`, to handle multichannel and to be consistent with the rest of the Web Platform. A frame can contain multiple samples (one per audio channel).

`droppedSamplesDuration`: it's unclear what this means. When does a capture device drop audio? It is because of a catastrophic real-time load problem? A bug in the implementation?

> The frequency of the media source is necessarily the same as the frequency of encoders later in pipeline.

This is using incorrect terminology, I think this is referring to the sampling-rate or sampling-frequency, not the frequency.

henbos:
Samples rather than frames were chosen for consistency with other metrics (we consistently use samples rather than frames when it comes to audio, e.g. [totalSamplesReceived](totalSamplesReceived)). This might not be ideal in hindsight, but it is addressed in the [Terminology section](https://w3c.github.io/webrtc-stats/#terminology):

> An _**audio sample**_ refers to having a sample in any channel of an audio track - if multiple audio channels are used, metrics based on samples do not increment at a higher rate, simultaneously having samples in multiple channels counts as a single sample.

> When does a capture device drop audio?

If the browser has a polling based model it may not be able to keep up with all the samples in all circumstances, though I am not the audio expert here. I can try to figure out more information about this if you like

henbos:
Asked the audio experts, they said "The capture device is a realtime device and if the browser does not process them in a timely manner it has to drop them"

padenot:
Right, so this is when a "catastrophic real-time load problem" occurs, which makes sense.

We need to revert this and have something on the `MediaStream` that represents an audio input device, or something like this.

It's not desirable to have coupling between the audio input system and WebRTC, and it's best to use the terms used in the industry and the rest of the Web Platform. `gUM` is frequently used without anything related to WebRTC, and knowing the input latency is capital in those scenarios (e.g., anything related to synchronizing recordings).

henbos:
Having media source metrics is not new and having measurements of glitches seem to fit well together with existing sample counters and audio energy metrics that we already get from the track.

[...]

PR #750, opened 2023-03-30
Document recently added audio metrics as features at risk

henbos:
The metrics relate to #741 and #742. The at risk discussion is also part of a wider discussion around WG consensus, see #746.


jan-ivar (approved):

vr000m:
We should remove them from webrtc-stats, https://github.com/w3c/webrtc-stats/issues/741

PR #756, opened 2023-05-03
Clarify inbound-rtp RTX handling

### LINKED ISSUE #755: Add clarification that inbound-rtp.bytesReceived includes retransmissions.
For bytesSent we say:

> Total number of bytes sent for this [SSRC](https://w3c.github.io/webrtc-stats/#dfn-ssrc). This includes retransmissions. Calculated as defined in [[RFC3550](https://w3c.github.io/webrtc-stats/#bib-rfc3550)] section 6.4.1.

For bytesReceived we say:

> Total number of bytes received for this [SSRC](https://w3c.github.io/webrtc-stats/#dfn-ssrc). Calculated as defined in [[RFC3550](https://w3c.github.io/webrtc-stats/#bib-rfc3550)] section 6.4.1.

They both seemingly refer to the same thing, and having symmetry here is good (bytesSent should be roughly equal to bytesReceived). And yes, we should include retransmissions because header bytes + bytes should equal to the total bitrate of that SSRC.

So it seems like a small discrepancy that we forgot to say "This includes retransmissions" for the bytesReceived case as well. Right? @alvestrand @vr000m @fippo

fippo:
fixes #755

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-stats/pull/756.html" title="Last updated on May 3, 2023, 6:30 AM UTC (e8ad4b9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/756/2b6f736...fippo:e8ad4b9.html" title="Last updated on May 3, 2023, 6:30 AM UTC (e8ad4b9)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -946,8 +946,8 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Total number of RTP packets received for this <a>SSRC</a>. At the receiving endpoint,
-                  this is calculated as defined in [[!RFC3550]] section 6.4.1. At the sending
+                  Total number of RTP packets received for this <a>SSRC</a>. This includes retransmissions.
+                  At the receiving endpoint, this is calculated as defined in [[!RFC3550]] section 6.4.1. At the sending
                   endpoint the {{packetsReceived}} is estimated by subtracting the Cumulative Number of Packets Lost
                   from the Extended Highest Sequence Number Received, both reported in the <a>RTCP Receiver
                   Report</a>, and then subtracting the initial Extended Sequence Number that was sent to this SSRC in a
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -946,8 +953,8 @@ <h2>
               </dt>
               <dd>
                 <p>
-                  Total number of RTP packets received for this <a>SSRC</a>. At the receiving endpoint,
-                  this is calculated as defined in [[!RFC3550]] section 6.4.1. At the sending
+                  Total number of RTP packets received for this <a>SSRC</a>. This includes retransmissions.
+                  At the receiving endpoint, this is calculated as defined in [[!RFC3550]] section 6.4.1. At the sending
```
fippo, line 957: Somewhat related, a lot of this language is pretty confusing since, due to inheritance, sometimes it relates to what is sent/received via RTCP. However, there are local counters too (you can count what you send/receive and that should actually be the basis for what you put into RTCP SR/RR)

vr000m: Typically the stats reported in RTCP are snapshots of the running counters.

For example jitter is calculated when each RTP packet is received. The algorithm in 3550 should apply irrespective. However depending on the timestamp in JSON along with the type of report explains if the data is from the most recent packet received or most recent RTCP.

What I think may be makes sense to do is that we explain this somewhere and then refer to the explanation section to avoid confusion.

fippo: Yeah. But for example [packetsReceived](https://w3c.github.io/webrtc-stats/#dom-rtcreceivedrtpstreamstats-packetsreceived) should be based on a counter when it is on inbound-rtp but based on RTCP on remote-inbound-rtp. The inheritance makes it weird, describing it separately might help.

henbos (approved):

henbos (approved):

PR #762, opened 2023-06-02
Specify fecBytesReceived

### LINKED ISSUE #751: RTX and FEC stats are incomplete.
While we have statistics about retransmission [packetsSent](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-retransmittedpacketssent) and [bytesSent](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-retransmittedbytessent) on [outbound-rtp](https://w3c.github.io/webrtc-stats/#dom-rtcstatstype-outbound-rtp) stats, we lack corresponding `retransmittedPacketsReceived` and `retransmittedBytesReceived` for [inbound-rtp](https://w3c.github.io/webrtc-stats/#dom-rtcstatstype-inbound-rtp).

I *believe* the reason is that without RTX it is not possible to identify retransmissions. However, RTX is available in all browsers these days (for video). When RTX is not used, these should not be present. #735 defines those.

Similarly, for FEC we have `inbound-rtp` stats for [fecPacketsReceived](https://w3c.github.io/webrtc-stats/#dom-rtcinboundrtpstreamstats-fecpacketsreceived) and [fecPacketsDiscarded](https://w3c.github.io/webrtc-stats/#dom-rtcinboundrtpstreamstats-fecpacketsdiscarded) but no corresponding `outbound-rtp` `fecPacketsSent` and no `fecBytesSent`. We also lack `fecBytesReceived`.

I *believe* this is due to FEC focusing on opus inband audio fec where the amount of bytes is somewhat hard to calculate *because opus*. For video FEC we have both `ulpfec` and `flexfec` implemented in browsers but don't have good statistics on them.

fippo:
similar to how retransmittedBytesReceived is defined.

Fixes #751

For audio "FEC" usually means opus inband FEC (also known as low-bitrate redundacy, LBRR). This is typically sent together with an actual opus primary frame as secondary payload. Yet we count those as fecPacketsReceived but you can't add or subtract this from the RTP packets received. Since this is part of the same RTP payload counting separate fecHeaderBytesReceived does not make sense (we do not count RTP headers separately for RTX either btw)

For video we have two FEC mechanisms, ULPFEC and FLEXFEC.
* FlexFEC uses a separate SSRC so is similar to RTX when it comes to counting things. I do not think that implementations currently properly account for flexfec bytes but the rationale that it should be a subset of bytesReceived holds.
* ULPFEC uses the same SSRC but a different payload type. It should be accounted for if one counts all packets with a particular SSRC already

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-stats/pull/762.html" title="Last updated on Jun 14, 2023, 1:00 PM UTC (b42b284)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/762/6e9db1c...fippo:b42b284.html" title="Last updated on Jun 14, 2023, 1:00 PM UTC (b42b284)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -466,6 +466,13 @@ enum RTCStatsType {
               and {{RTCInboundRtpStreamStats/retransmittedBytesReceived}} counters of the relevant
               {{RTCInboundRtpStreamStats}} objects.
             </p>
+            <p>
+              FEC streams do not show up as separate
+              {{RTCInboundRtpStreamStats}} objects but affect the {{RTCReceivedRtpStreamStats/packetsReceived}},
+              {{RTCInboundRtpStreamStats/bytesReceived}}, {{RTCInboundRtpStreamStats/fecPacketsReceived}}
+              and {{RTCInboundRtpStreamStats/fecBytesReceived}} counters of the relevant
+              {{RTCInboundRtpStreamStats}} objects.
+            </p>
           </dd>
           <dt>
             <dfn>outbound-rtp</dfn>
@@ -1283,7 +1290,9 @@ enum RTCStatsType {
               <dd>
                 <p>
                   Total number of RTP FEC bytes received for this <a>SSRC</a>, only including payload bytes.
-                  This is a subset of {{RTCInboundRtpStreamStats/bytesReceived}}.
+                  This is a subset of {{RTCInboundRtpStreamStats/bytesReceived}}. If a FEC mechanism that
+                  uses a different {{RTCRtpStreamStats/ssrc}} was negotiated, FEC packets are sent over a
+                  separate ssrc but is still accounted for here.
                 </p>
               </dd>
               <dt>
@@ -1292,9 +1301,11 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  Total number of RTP FEC packets received for this <a>SSRC</a>. This counter can also be
-                  incremented when receiving FEC packets in-band with media packets (e.g., with
-                  Opus).
+                  Total number of RTP FEC packets received for this <a>SSRC</a>. If a FEC mechanism that
+                  uses a different {{RTCRtpStreamStats/ssrc}} was negotiated, FEC packets are sent over a
+                  separate ssrc but is still accounted for here.
+                  This counter can also be incremented when receiving FEC packets in-band with media packets
+                  (e.g., with Opus).
                 </p>
               </dd>
               <dt>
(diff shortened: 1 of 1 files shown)


bdrtc:
The stats only exist for video ?  audio fec is inband payload for opus codec.

fippo:
For audio it would be possible to account for the FEC LBRR frames

webrtc-stats.html
```diff
@@ -1275,6 +1276,16 @@ <h2>
...
                 </p>
               </dd>
+              <dt>
+                <dfn>fecBytesReceived</dfn> of type <span class=
+                "idlMemberType">unsigned long long</span>
+              </dt>
+              <dd>
+                <p>
+                  Total number of RTP FEC bytes received for this <a>SSRC</a>, only including payload bytes.
+                  This is a subset of {{RTCInboundRtpStreamStats/bytesReceived}}.
```
henbos, line 1286: This looks good to me. What about fecHeaderBytesReceived as a subset of headerBytesReceived for completeness or is there something about that that doesn't make sense?

fippo:
This is somewhat complicated so 🐻 with me.
* For audio "FEC" usually means opus inband FEC (also known as low-bitrate redundacy, LBRR. This is typically sent together with an actual opus primary as secondary payload. Yet we count those as fecPacketsReceived but you can't add or subtract this from the RTP packets received. Since this is part of the same RTP payload counting separate fecHeaderBytesReceived does not make sense (we do not count RTP headers separately for RTX either btw)

* For video we have two FEC mechanisms, ULPFEC and FLEXFEC.
  * FlexFEC uses a separate SSRC so is similar to RTX when it comes to counting things. I do not think that implementations currently properly account for flexfec bytes but the rationale that it should be a subset of bytesReceived holds.
  * ULPFEC uses the same SSRC but a different payload type. It should be accounted for if one counts all packets with a particular SSRC already

webrtc-stats.html
henbos: Ack

henbos (approved):
Good explanation. Let's be consistent with FEC and not add more counters than we need. LGTM

webrtc-stats.html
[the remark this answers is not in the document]
fippo: True, tweaked accordingly!

henbos (approved):

henbos:
Fippo says yes! Merge

PR #774, opened 2023-12-27
[editorial] add "only exists" clause to various audio/video stats

fippo:
and add linebreaks after those clauses.
Also sync definition of "mid" stat between inbound-rtp and outbound-rtp

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1137,6 +1137,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
+                  Does not [= map/exist =] for audio.
                   Total number of seconds that have been spent decoding the {{framesDecoded}}
                   frames of this stream. The average decode time can be calculated by dividing this
                   value with {{framesDecoded}}. The time it takes to decode one frame is the
@@ -1177,6 +1178,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
+                  Does not [= map/exist =] for audio.
                   Count the total number of video pauses experienced by this receiver.
                   Video is considered to be paused if time passed since last rendered
                   frame exceeds 5 seconds. {{pauseCount}} is incremented when a frame
@@ -1189,6 +1191,7 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
+                  Does not [= map/exist =] for audio.
                   Total duration of pauses (for definition of pause see {{pauseCount}}),
                   in seconds. This value is updated when a frame is rendered.
                 </p>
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -1913,9 +1922,9 @@ <h2>
               </dt>
               <dd>
                 <p>
-                  Only [= map/exist =]s if the {{RTCRtpTransceiver}} owning this stream has a
-                  {{RTCRtpTransceiver/mid}} value that is not null. this is that
-                  value.
+                  If the {{RTCRtpTransceiver}} owning this stream has a
```
fippo, line 1925: Taken from https://w3c.github.io/webrtc-stats/#dom-rtcinboundrtpstreamstats-mid
I wonder if inbound-rtp and outbound-rtp need to take care of the edge case where mid was offered but not negotiated but that seems orthogonal

henbos:
I think this is PR is great thing!

But nit: while "no implementation change intended", this is not strictly "editorial" since we change from "applicable to both kinds" to "applicable to only video".

fippo:
But only for things which we already knew were video-only but this was not documented?

fippo:
Adjusted the title, this is above "fixes whitespace" but technically a normative change.

henbos (approved):
This PR accurately describes the kinds implemented/not-implemented in libwebrtc.

@jan-ivar, do we have the permission to merge this "almost editorial" PR?

webrtc-stats.html
henbos: I'd update both of them to say "otherwise this member does not [= map/exist =]." so that they're consistent with each other but we also make use of the "map/exist" reference used everywhere else

henbos: But the edge case you're describing seems orthogonal and not sure it's needed since this is just piggybacking on whatever the transceiver is doing

fippo: I think we need to distinguish between "exist" and "present"

henbos: But the [implementation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/peerconnection/rtc_stats_report.cc;l=137;drc=e5f8ff312fe542f39c7d34b4440f18d1500b15f9) does not make such distinction. It either is there and has a string value or it is not there. We don't support null or even undefined values. If a value on a dictionary is not set, it does not [= map/exist =].

fippo: Yeah. But I think we should only use exist for things that will never exist, not things that will not be set on some condition

henbos: But it will never exist for that object. Every getStats call returns a new object, so we only have objects where it always existed and objects where it never existed. The continuation over time is an illusion.

fippo: Not a hill to die on, i'm happy using the previous text as long as it is done in both locations

webrtc-stats.html
```diff
@@ -1137,6 +1137,7 @@ <h2>
               </dt>
               <dd>
                 <p>
+                  Does not [= map/exist =] for audio.
```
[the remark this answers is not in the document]
jan-ivar: To be followed up in #775

PR #780, opened 2024-01-25
clarify that sdpFmtpLine may not be there

### LINKED ISSUE #779: Is codec.sdpFmtpLine present when there is no fmtp line in the SDP?
Which is typically the case for VP8 which has no parameters. In Chrome this leads to
image
while sdpFmtpLine is defined as 
```
DOMString     sdpFmtpLine;
```
[here](https://w3c.github.io/webrtc-stats/#codec-dict*).

It seems like this should only exist when the codec has a fmtp line? No code changes required but we might need some WPT updates.

fippo:
fixes #779

oddly this does not render as `as defined by [RFC8829] (section 5.8.)` like the corresponding part in [webrtc-pc](https://w3c.github.io/webrtc-pc/#rtcrtpcodec) -- and as this shows getting the section numbers wrong is easy ;-)
It seems we need [linkToJsep magic](https://github.com/w3c/webrtc-pc/blob/main/webrtc.js#L105)

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-stats/pull/780.html" title="Last updated on Jan 25, 2024, 3:24 PM UTC (df1c1ff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/780/12c9dd3...fippo:df1c1ff.html" title="Last updated on Jan 25, 2024, 3:24 PM UTC (df1c1ff)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -879,7 +879,7 @@ enum RTCStatsType {
                   <code class="sdp">a=fmtp</code> line in the SDP
                   corresponding to the codec, if one [= map/exist =]s,
                   as defined by
-                  <span data-jsep="parsing-a-desc">[[!RFC8829]]</span>.
+                  [[!RFC8829]] (<a href="https://www.rfc-editor.org/rfc/rfc8829#section-5.8">section 5.8</a>).
                 </p>
               </dd>
             </dl>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -875,8 +875,11 @@ enum RTCStatsType {
               </dt>
               <dd>
                 <p>
-                  The a=fmtp line in the SDP corresponding to the codec, i.e., after the colon
-                  following the PT. This defined by [[!JSEP]] in Section 5.7.
+                  The "format specific parameters" field from the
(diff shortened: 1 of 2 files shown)


webrtc-stats.html
```diff
@@ -875,8 +875,11 @@ <h2>
               </dt>
               <dd>
                 <p>
-                  The a=fmtp line in the SDP corresponding to the codec, i.e., after the colon
-                  following the PT. This defined by [[!JSEP]] in Section 5.7.
+                  The "format specific parameters" field from the
+                  <code class="sdp">a=fmtp</code> line in the SDP
+                  corresponding to the codec, if one [= map/exist =]s,
+                  as defined by
+                  <span data-jsep="parsing-a-desc">[[!RFC8829]]</span>.
```
fippo, line 882:
```suggestion
                  [[!RFC8829]] (<a href="https://www.rfc-editor.org/rfc/rfc8829#section-5.8">section 5.8</a>).
```

fippo: (this would resolve it by hand since it is the only reference)

alvestrand (approved):

PR #788, opened 2024-09-17
Add definitions of corruption detection measurements

### LINKED ISSUE #787: Add definitions of corruption detection measurement statistics
We'd like to add a new feature that enables capturing video corruptions, and that requires a new stats value.

Some background:  We're try to catch bugs that cause the input to the encoder to diverge from the output of the decoder in ways that are unexpected (not artifacts that are expected from lossy compression).

These can be cause by bugs in a number of components:
* Encoder implementations
* Incorrect metadata
* RTP Packetization
* SFU/backend logic
* Jitter buffer/frame assembly
* Decoder implementations

Common among these bugs is that they are very hard to catch as they are not visible in any existing statistics, apart from user complaints. That makes them very hard and time consuming to triage and root cause.

In http://www.webrtc.org/experiments/rtp-hdrext/corruption-detection we propose a new header extension that adds metadata to the encoded frames that a receiver can use to validate that the decoder is in a consistent state. In order to surface this signal to the application layer, we need a new value in RTCInboundRtpStreamStats.

Note that we don't wish to tie the definition of the statistics too hard to our proposed implementation. Rather, we'd like to use a "probability of corruption" likelihood measure that may be produced using that header extension, or it may be produced some other way (e.g. using image recognition techniques on the receive side alone).

sprangerik:
Fixes #787

This PR adds new statistics for a corruption probability measure of a received video stream.

We include both the sum and sum of squares for this measure, so that a variance can be calculated.

***
<a href="https://pr-preview.s3.amazonaws.com/sprangerik/webrtc-stats/pull/788.html" title="Last updated on Oct 29, 2024, 1:35 PM UTC (55c3100)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/788/d6c2dbf...sprangerik:55c3100.html" title="Last updated on Oct 29, 2024, 1:35 PM UTC (55c3100)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1793,6 +1793,11 @@ enum RTCStatsType {
                   measurements SHOULD be [= map/exist | present =].
                 </p>
                 <p class="note">
+                  Note that the corruption-detection header extension documented at
+                  http://www.webrtc.org/experiments/rtp-hdrext/corruption-detection is experimental. The
+                  identifier and format may change once an IETF standard has been established.
+                </p>
+                <p class="note">
                   The user agent MAY produce corruption likelihood measurements using any method available.
                   That could be facilitated by e.g. the mentioned corruption detection header extension,
                   reference-less image analysis such as natural image statistics, ML-based classification,
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1793,26 +1793,25 @@ enum RTCStatsType {
                   measurements SHOULD be [= map/exist | present =].
                 </p>
                 <p class="note">
-                  Note that the corruption-detection header extension documented at
-                  http://www.webrtc.org/experiments/rtp-hdrext/corruption-detection is experimental. The
-                  identifier and format may change once an IETF standard has been established.
-                </p>
-                <p class="note">
                   The user agent MAY produce corruption likelihood measurements using any method available.
                   That could be facilitated by e.g. the mentioned corruption detection header extension,
                   reference-less image analysis such as natural image statistics, ML-based classification,
-                  or anything else capable of producing an estimate.
-                </p>
-                <p class="note">
-                  Note that these values are estimates - not guarantess. Even if the estimate is 0.0 there
-                  could be corruptions present (i.e. it's a false negative) for instance if random sampling
-                  is used and only a very small area of the frame is affected. Similarly, even if the estimate
-                  is 1.0 there might not be corruptions present (i.e. it's a false positive) for instance if
-                  the distribution of per-macroblock quantization parameters is extremely wide.
-
+                  or anything else capable of producing an estimate.<br>
+                  <br>
+                  The corruption likeliehood values are estimates - not guarantess. Even if the estimate is
+                  0.0 there could be corruptions present (i.e. it's a false negative) for instance if random
+                  sampling is used and only a very small area of the frame is affected. Similarly, even if the
+                  estimate is 1.0 there might not be corruptions present (i.e. it's a false positive) for
+                  instance if an image recognition technique is used while presenting a screenshot of a
+                  video encoding corruption.<br>
                   Just like there are edge cases for e.g. PSNR measrumenets, these metrics should primarily
                   be used as a basis for statistical analysis rather than be used as an absolute truth on a
-                  per-frame basis.
+                  per-frame basis.<br>
+                  <br>
+                  The corruption-detection header extension documented at
+                  <a href="http://www.webrtc.org/experiments/rtp-hdrext/corruption-detection">
+                  http://www.webrtc.org/experiments/rtp-hdrext/corruption-detection</a> is experimental. The
+                  identifier and format may change once an IETF standard has been established.
                 </p>
               </dd>
             </dl>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1004,6 +1004,9 @@ enum RTCStatsType {
              unsigned long long   retransmittedBytesReceived;
              unsigned long        rtxSsrc;
              unsigned long        fecSsrc;
+             double               totalCorruptionProbability;
+             double               totalSquaredCorruptionProbability;
+             unsigned long long   corruptionMeasurements;
             };</pre>
           <section>
             <h2>
@@ -1743,6 +1746,48 @@ enum RTCStatsType {
                   of the FEC stream that is associated with this stream's {{RTCRtpStreamStats/ssrc}}.
                   If FEC is not negotiated or uses the same <a>RTP stream</a>, this value MUST NOT be
                   [= map/exist | present =].
+                </p>
+              </dd>
+              <dt>
+                <dfn>totalCorruptionProbability</dfn> of type <span class=
+                "idlMemberType">double</span>
+              </dt>
+              <dd>
+                <p>
+                  MUST NOT [= map/exist =] for audio. Represents the cumulative sum of all corruption
+                  probability measurements that have been made for this SSRC, see {{corruptionMeasurements}}
+                  regarding when this attribute SHOULD be [= map/exist | present =].
+                </p>
+              </dd>
+              <dt>
+                <dfn>totalSquaredCorruptionProbability</dfn> of type <span class=
+                "idlMemberType">double</span>
+              </dt>
+              <dd>
+                <p>
+                  MUST NOT [= map/exist =] for audio. Represents the cumulative sum of all corruption
+                  probability measurements squared that have been made for this SSRC, see
+                  {{corruptionMeasurements}} regarding when this attribute SHOULD be [= map/exist | present =].
+                </p>
+              </dd>
+              <dt>
+                <dfn>corruptionMeasurements</dfn> of type <span class=
+                "idlMemberType">unsigned long long</span>
+              </dt>
+              <dd>
+                <p>
+                  MUST NOT [= map/exist =] for audio. When the user agent is able to make a corruption
+                  probability measurement, this counter is incremented for each such measurement and
+                  {{totalCorruptionProbability}} and {{totalSquaredCorruptionProbability}} are aggregated
+                  with this measurement and measurement squared respectively.
+                </p>
+                <p>
+                  If the
+                  <a href="http://www.webrtc.org/experiments/rtp-hdrext/corruption-detection">Corruption Detection</a>
+                  header extension has been negotatied, the user agent SHOULD produce measurements
+                  according to the algorithm in the header extension-explainer, but the user agent MAY produce
+                  measurements by other means, for example by reference-less image analysis compeletely on the
+                  receiver-side.
                 </p>
               </dd>
             </dl>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1766,6 +1766,16 @@ enum RTCStatsType {
                   likely some corruption visible, but it could for instance have a low magnitude or be present
                   only in a small portion of the frame.
                 </p>
+                <p class="note">
+                  The corruption likelihood values are estimates - not guarantees. Even if the estimate is
+                  0.0 there could be corruptions present (i.e. it's a false negative) for instance if only a
+                  very small area of the frame is affected. Similarly, even if the estimate is 1.0 there might
+                  not be a corruption present (i.e. it's a false positive) for instance if there are
+                  macroblocks with a QP far higher than the frame average.
+                  Just like there are edge cases for e.g. PSNR measurements, these metrics should primarily
+                  be used as a basis for statistical analysis rather than be used as an absolute truth on a
+                  per-frame basis.
+                </p>
               </dd>
               <dt>
                 <dfn>totalSquaredCorruptionProbability</dfn> of type <span class=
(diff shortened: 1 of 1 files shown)


alvestrand:
Awaiting the TPAC presentation.

jan-ivar:
Thanks! Sounds like a good idea to me. Marking as draft pending upcoming discussion.

Marking `CR blocking` due to the [google source reference](https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/corruption-detection). Any volunteers to pushing the RTP header extension through IETF?

webrtc-stats.html
```diff
@@ -1745,6 +1748,48 @@ <h2>
...
+              </dt>
+              <dd>
+                <p>
+                  MUST NOT [= map/exist =] for audio. When the user agent is able to make a corruption
+                  probability measurement, this counter is incremented for each such measurement and
+                  {{totalCorruptionProbability}} and {{totalSquaredCorruptionProbability}} are aggregated
+                  with this measurement and measurement squared respectively.
+                </p>
+                <p>
+                  If the
```
Orphis, line 1785: It would be better to have just a note mentioning the experimental header extension rather than a normative step.
As the header extension is not an established publication, it would block the document from advancing to Candidate Recommendation.
It may leave open the corruption detection algorithm for each implementation to allow alternative techniques.

sprangerik: I didn't intend this to be normative. That header extension does not need to be either negotiated or implemented. How would you phrase it differently? Something like this?

'The user agent may produce corruption likelihood measurements using any method available. That could be facilitated by for instance: a side-channel for validation metadata such as the header extension described at <a href="http://www.webrtc.org/experiments/rtp-hdrext/corruption-detection">http://www.webrtc.org/experiments/rtp-hdrext/corruption-detection</a>, reference-less image analysis such as natural image statistics, ML-based classification, or anything else capable of producing an estimate.'

handellm:
Depending on the math behind `totalCorruptionProbability` and `totalSquaredCorruptionProbability` and the fact that these are `double`s and if these are ever-increasing quantities, could there be numerical issues mounting by incrementing over longer times?

sprangerik:
> Thanks! Sounds like a good idea to me. Marking as draft pending upcoming discussion.
> 
> Marking `CR blocking` due to the [google source reference](https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/corruption-detection). Any volunteers to pushing the RTP header extension through IETF?

I have update the text to make it more clear that method is not normative. ptal!

Also, I volunteer to push that header extension through IETF :)

sprangerik:
> Depending on the math behind `totalCorruptionProbability` and `totalSquaredCorruptionProbability` and the fact that these are `double`s and if these are ever-increasing quantities, could there be numerical issues mounting by incrementing over longer times?

Doubtful. Even in the case where a call is run for a very very long time, remember that the individual samples are capped in the [0.0, 1.0] range and are likely only (individually) useful to a few decimal places. With a base of 52 bits, that's still an enormous amounts of samples that can be accumulated without noticeable loss. Even further reducing the scope of this problem is that we expect the "normal" values to be essentially 0.0. If you run extremely long sessions with constant corruption, might you have other problems to prioritize first.

webrtc-stats.html
```diff
@@ -1745,6 +1748,49 @@ <h2>
...
+                "idlMemberType">unsigned long long</span>
+              </dt>
+              <dd>
+                <p>
+                  MUST NOT [= map/exist =] for audio. When the user agent is able to make a corruption
+                  probability measurement, this counter is incremented for each such measurement and
+                  {{totalCorruptionProbability}} and {{totalSquaredCorruptionProbability}} are aggregated
+                  with this measurement and measurement squared respectively.
+                </p>
+                <p class="note">
```
henbos, line 1784: From editor's meeting:
- @youennf and @jan-ivar prefers to convert from a note to more normative language. Can we say MUST and point to this header extension, rather than saying "anything"?
- If we can't say exactly what the algorithm is, can we at least define what a acceptable algorithm behaves?
- An adopted IETF draft is needed to reference a link that isn't /experiments/, this would be a blocker for candidate recommendation but not a blocker for merging this PR, can we add a note that says something like this?

sprangerik:
@jan-ivar would you mind reviewing this?

sprangerik:
@youennf would you mind reviewing this?:

henbos:
Ping @youennf, can we merge this as is or is there more work needed?

webrtc-stats.html
```diff
@@ -1745,6 +1748,72 @@ <h2>
...
+                  MUST NOT [= map/exist =] for audio. When the user agent is able to make a corruption
+                  probability measurement, this counter is incremented for each such measurement and
+                  {{totalCorruptionProbability}} and {{totalSquaredCorruptionProbability}} are aggregated
+                  with this measurement and measurement squared respectively.
+                  If the <a href="http://www.webrtc.org/experiments/rtp-hdrext/corruption-detection">
+                  corruption-detection</a> header extension has been negotiated, corruptioned probability
+                  measurements SHOULD be [= map/exist | present =].
+                </p>
+                <p class="note">
+                  The user agent MAY produce corruption likelihood measurements using any method available.
```
youennf, line 1796: Can we have a MAY inside a note?
Also, is Chrome planning to use any other method than the header?
If not, I suggest not saying this, and we can remove this note.
We can always extend the allowed methods later on when there are more concrete proposals.

webrtc-stats.html
```diff
@@ -1745,6 +1748,72 @@ <h2>
...
+                  reference-less image analysis such as natural image statistics, ML-based classification,
+                  or anything else capable of producing an estimate.<br>
+                  <br>
+                  The corruption likeliehood values are estimates - not guarantess. Even if the estimate is
+                  0.0 there could be corruptions present (i.e. it's a false negative) for instance if random
+                  sampling is used and only a very small area of the frame is affected. Similarly, even if the
+                  estimate is 1.0 there might not be corruptions present (i.e. it's a false positive) for
+                  instance if an image recognition technique is used while presenting a screenshot of a
+                  video encoding corruption.<br>
+                  Just like there are edge cases for e.g. PSNR measrumenets, these metrics should primarily
```
youennf, line 1807:
```suggestion
                  Just like there are edge cases for e.g. PSNR measurements, these metrics should primarily
```

webrtc-stats.html
```diff
@@ -1745,6 +1748,72 @@ <h2>
...
+              </dt>
+              <dd>
+                <p>
+                  MUST NOT [= map/exist =] for audio. When the user agent is able to make a corruption
+                  probability measurement, this counter is incremented for each such measurement and
+                  {{totalCorruptionProbability}} and {{totalSquaredCorruptionProbability}} are aggregated
+                  with this measurement and measurement squared respectively.
+                  If the <a href="http://www.webrtc.org/experiments/rtp-hdrext/corruption-detection">
+                  corruption-detection</a> header extension has been negotiated, corruptioned probability
+                  measurements SHOULD be [= map/exist | present =].
```
youennf, line 1793: It is negotiation or the presence of the header extension in RTP packets of the stream?

webrtc-stats.html
```diff
@@ -1745,6 +1748,72 @@ <h2>
...
+                <p class="note">
+                  The user agent MAY produce corruption likelihood measurements using any method available.
+                  That could be facilitated by e.g. the mentioned corruption detection header extension,
+                  reference-less image analysis such as natural image statistics, ML-based classification,
+                  or anything else capable of producing an estimate.<br>
+                  <br>
+                  The corruption likeliehood values are estimates - not guarantess. Even if the estimate is
+                  0.0 there could be corruptions present (i.e. it's a false negative) for instance if random
+                  sampling is used and only a very small area of the frame is affected. Similarly, even if the
+                  estimate is 1.0 there might not be corruptions present (i.e. it's a false positive) for
```
youennf, line 1804:
```suggestion
                  estimate is 1.0 there might not be corruption present (i.e. it's a false positive) for
```

webrtc-stats.html
```diff
@@ -1745,6 +1748,72 @@ <h2>
...
+                  corruption-detection</a> header extension has been negotiated, corruptioned probability
+                  measurements SHOULD be [= map/exist | present =].
+                </p>
+                <p class="note">
+                  The user agent MAY produce corruption likelihood measurements using any method available.
+                  That could be facilitated by e.g. the mentioned corruption detection header extension,
+                  reference-less image analysis such as natural image statistics, ML-based classification,
+                  or anything else capable of producing an estimate.<br>
+                  <br>
+                  The corruption likeliehood values are estimates - not guarantess. Even if the estimate is
```
youennf, line 1801:
```suggestion
                  The corruption likelihood values are estimates - not guarantess. Even if the estimate is
```

webrtc-stats.html
```diff
@@ -1745,6 +1748,72 @@ <h2>
...
+                "idlMemberType">unsigned long long</span>
+              </dt>
+              <dd>
+                <p>
+                  MUST NOT [= map/exist =] for audio. When the user agent is able to make a corruption
+                  probability measurement, this counter is incremented for each such measurement and
+                  {{totalCorruptionProbability}} and {{totalSquaredCorruptionProbability}} are aggregated
+                  with this measurement and measurement squared respectively.
+                  If the <a href="http://www.webrtc.org/experiments/rtp-hdrext/corruption-detection">
+                  corruption-detection</a> header extension has been negotiated, corruptioned probability
```
youennf, line 1792:
```suggestion
                  corruption-detection</a> header extension has been negotiated, corruption probability
```
?

webrtc-stats.html
youennf: Also s/guarantess/guarantees/

webrtc-stats.html
sprangerik: That is indeed a better distinction, updated. I'll change the SHOULD to a MUST with that change too.

webrtc-stats.html
sprangerik: Good question, I'll change it to a lower-case may then.

I have plans to add other methods yes, but it's further down the line for sure. If it causes less confusion I can remove this part for now.

The part about expected values is still valid however, but I'll move that up to the part where the measurement values are described.

sprangerik (commented):
Thanks the comments, ptal at the update!

sprangerik:
ping @youennf, any further comments?

webrtc-stats.html
```diff
@@ -1745,6 +1748,67 @@ <h2>
...
+                  Each measurement added to {{totalCorruptionProbability}} MUST be in the range [0.0, 1.0],
+                  where a value of 0.0 indicates the system has estimated there is no or negligible corruption
+                  present in the processed frame. Similarly a value of 1.0 indicates there is almost certainly
+                  a corruption visible in the processed frame. A value in between those two indicates there is
+                  likely some corruption visible, but it could for instance have a low magnitude or be present
+                  only in a small portion of the frame.
+                </p>
+                <p class="note">
+                  The corruption likelihood values are estimates - not guarantees. Even if the estimate is
+                  0.0 there could be corruptions present (i.e. it's a false negative) for instance if only a
```
youennf, line 1771:
```suggestion
                  0.0, there could be corruptions present (i.e. it's a false negative) for instance if only a
```

youennf (approved):
LGTM.
I would file an issue once this lands to track the work to replace the webrtc.org link with an IETF link.

sprangerik:
> LGTM. I would file an issue once this lands to track the work to replace the webrtc.org link with an IETF link.

Thanks! Yes, that makes sense.

sprangerik:
@jan-ivar anything further from your side?

henbos (approved):

webrtc-stats.html
jan-ivar: Please use "can" or "might" or otherwise rephrase to avoid "may" or "MAY" in notes

jan-ivar (approved):
LGTM. As a newb on this, it might be good to give the reader an example formula they can use when reading these stats. Like we do in https://w3c.github.io/webrtc-stats/#dom-rtcinboundrtpstreamstats-totalinterframedelay

But not a blocker.

sprangerik:
> LGTM. As a newb on this, it might be good to give the reader an example formula they can use when reading these stats. Like we do in https://w3c.github.io/webrtc-stats/#dom-rtcinboundrtpstreamstats-totalinterframedelay
> 
> But not a blocker.

Good point, but let med do that as a follow-up.

alvestrand:
@jan-ivar why did you add the "CR Blocker" tag on this one?

jan-ivar:
@alvestrand see https://github.com/w3c/webrtc-stats/pull/788#issuecomment-2361162211. To my understanding there's still a normative reference to a webrtc.org doc [here](https://pr-preview.s3.amazonaws.com/sprangerik/webrtc-stats/pull/788.html#dom-rtcinboundrtpstreamstats-corruptionmeasurements). @sprangerik volunteered in https://github.com/w3c/webrtc-stats/pull/788#issuecomment-2394326319 to push it through IETF.

So merging this would block this spec from recycling at CR. We may be OK with that, but we should track it somehow.

Erik, can you open a follow-up issue to come back and fix this link whenever your IETF efforts have concluded? Then we can label that issue and merge this, if that's how we'd like to proceed.

jan-ivar:
Reading through the thread again I see there was an effort to make it non-normative, but this part still sounds normative to me:
- _"If the [corruption-detection](http://www.webrtc.org/experiments/rtp-hdrext/corruption-detection) header extension is present in the RTP packets, corruption probability measurements MUST be [present](https://infra.spec.whatwg.org/#map-exists)."_

sprangerik:
> @alvestrand see [#788 (comment)](https://github.com/w3c/webrtc-stats/pull/788#issuecomment-2361162211). To my understanding there's still a normative reference to a webrtc.org doc [here](https://pr-preview.s3.amazonaws.com/sprangerik/webrtc-stats/pull/788.html#dom-rtcinboundrtpstreamstats-corruptionmeasurements). @sprangerik volunteered in [#788 (comment)](https://github.com/w3c/webrtc-stats/pull/788#issuecomment-2394326319) to push it through IETF.
> 
> So merging this would block this spec from recycling at CR. We may be OK with that, but we should track it somehow.
> 
> Erik, can you open a follow-up issue to come back and fix this link whenever your IETF efforts have concluded? Then we can label that issue and merge this, if that's how we'd like to proceed.

Yes, that was my intention! https://github.com/w3c/webrtc-stats/pull/788#issuecomment-2444255504 might not have been clear on that point.

sprangerik:
> Reading through the thread again I see there was an effort to make it non-normative, but this part still sounds normative to me:
> 
> * _"If the [corruption-detection](http://www.webrtc.org/experiments/rtp-hdrext/corruption-detection) header extension is present in the RTP packets, corruption probability measurements MUST be [present](https://infra.spec.whatwg.org/#map-exists)."_

In the sense that _if_ you have decided to implement support for that extensions and have negotiated it for a particular RTP stream, then it makes sense to me to say that we should surface the stats value for it. Do you have a different formulation in mind you think is more suitable?

sprangerik:
I filed https://github.com/w3c/webrtc-stats/issues/790 to track updating the definitions. I don't seem to have access to change labels though. @jan-ivar can you assist with that?

henbos:
I added the CR blocker label on the issue filed and removed it from this PR, sounds good?

Orphis (approved):

PR #792, opened 2025-01-16
Add counters that support deployment of L4S

### LINKED ISSUE #793: Need support for statistics for L4S
If WebRTC connections are to be deployed on L4S congestion controlled networks, we need stats to measure success (or not).

alvestrand:
These counters also depend on RFC8888 being deployed.

An explainer for the design is included.

Fixes #793

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/pull/792.html" title="Last updated on Mar 24, 2025, 7:31 PM UTC (e820135)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/792/02094bc...e820135.html" title="Last updated on Mar 24, 2025, 7:31 PM UTC (e820135)">Diff</a>

### DIFF
--- a/l4s-explainer.md
+++ b/l4s-explainer.md
@@ -31,7 +31,7 @@ There are a number of scenarios we want to diagnose:
 *   Network drops: Packets marked with ECT(1) get dropped, packets without it get through
 *   CE-less chokepoints: We observe packet loss, but no CE markings.
 
-We may also want to look at packet loss vs packet reordering - when packets are lost, the reports will contain loss markings; when packets are reordered, later reports will overlap the previous reports and add info on packets previously lost.
+We may also want to look at packet loss vs packet reordering - when packets are lost, the reports will contain loss markings; when packets are reordered, later reports will overlap the previous reports and add info on packets previously reported as missing.
 
 
 ## Suggested counters and where to attach them
@@ -70,9 +70,9 @@ Diagnosing is performed at the sending endpoint.
 
 
 
-*   CE-aware chokepoints: ECT(1) is sent, ECT(1) and CE are in reports, reported ECT(1) + CE + lost add up to number of sent  packets (modulo in-flight). Bleaching stays at zero.
+*   CE-aware chokepoints: ECT(1) is sent, ECT(1) and CE are in reports, reported ECT(1) + CE + lost add up to the number of sent packets (modulo in-flight). Bleaching stays at zero.
 *   Bleaching: ECT(1) is sent, but “no marking” is reported. Number of sent packets and number of received packets + lost are roughly equal.
 *   Network drops: Packets sent with normal and ECT(1) are both above zero, but packets received are only normal, not ECT(1), and correspond to the number of normal packets. (NOTE: This is a failure scenario for deploying L4S)
-*   CE-less chokepoints: ECT(1) is sent and received, but CE counter remains at zero.
+*   CE-less chokepoints: ECT(1) is sent and received, but CE counter remains at zero. Number of packets reported as lost in RFC8888 reports is above zero.
 
 If excessive reordering occurs, the “reported later” counters will go up; the precise interpretation of that number depends on the strategy used for scheduling RFC8888 reports (longer intervals will allow more reordered packets to be recovered without this being visible in the reports).
--- a/l4s-explainer.md
+++ b/l4s-explainer.md
@@ -63,6 +63,13 @@ Attached to RTCRemoteInboundRtpStreamStats (which is a subclass of RTCReceivedRt
 
 Data in RTCRemoteInboundRtpStreamStats are computed based on a sender’s knowledge of the outgoing packets + data from remote reports (RR or RFC8888 reports). Computing “bleached” packets requires having both info on the sent packet and info on the RFC8888 report available.
 
+Attached to RTCTransportStats:
+
+* Number of RFC 8888 format feedback messages sent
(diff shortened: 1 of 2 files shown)


henbos:
Interesting stuff

webrtc-stats.html
```diff
@@ -917,6 +921,41 @@ <h2>
...
+              <dt>
+                <dfn>packetsReportedAsLost</dfn> of type <span class="idlMemberType">unsigned long long</span>
+              </dt>
+              <dd>
+                <p>
+                  Total number of RTP packets for which an [[RFC8888]] section 3.1 report has been sent with a zero R bit.
+                  Only reported if support for the "ccfb" feedback mechanism has been negotiated.
+                </p>
+              </dd>
+                <dfn>packetsReportedAsLostButRecovered</dfn> of type <span class="idlMemberType">unsigned long long</span>
```
dontcallmedom, line 950:
```suggestion
                <dt><dfn>packetsReportedAsLostButRecovered</dfn> of type <span class="idlMemberType">unsigned long long</span></dt>
```

jan-ivar:
@jesup any concerns with this?

l4s-explainer.md
```diff
@@ -0,0 +1,78 @@
...
+
+Attached to RTCReceivedRtpStreamStats:
+
+
+
+*   Number of packets received with normal markings
+*   Number of packets received with ECT(1) marking
+*   Number of packets received with CE marking
+*   Number of packets reported as lost in RFC8888 reports
+*   Number of packets reported as lost in one RFC8888 report but later reported as arrived
```
shrim27, line 56: Another counter useful would be RFC 8888 feedback send interval. Cumulative RFC 8888 feedback sent and Cumulative RFC feedback interval. Avg interval could be computed using those 2 stats.

vr000m:
I dont think there is any overlap with the other FEC stats, since we did not have loss recovery stats before. 

Re: RTCP interval, is the implementation of RFC8888 going to be compound packets or non-compound? former means we can reuse RTCP intervals for calculating the reporting interval.

alvestrand:
I think just counting the number of 8888 reports sent would be a better stat than the interval; the average interval can be computed from the increase in the count between two measurements.
We've discussed tweaks like sending immediate reports on CE feedback rather than waiting for the regular cadence; this might be required to satisfy L4S feedback loop speed requirements. If we do that, averages would be misleading.

webrtc-stats.html
```diff
@@ -917,6 +921,41 @@ <h2>
...
+              </dd>
+                <dt><dfn>packetsReportedAsLostButRecovered</dfn> of type <span class="idlMemberType">unsigned long long</span></dt>
+              </dt>
+              <dd>
+                <p>
+                  Total number of RTP packets for which an [[RFC8888]] section 3.1 report has been sent with a zero R bit,
+                  but a later report for the same packet has the R bit set to 1.
+                  Only reported if support for the "ccfb" feedback mechanism has been negotiated.
+                </p>
+              </dd>
```
vr000m, line 958: should we change the name from *ButRecovered to *ButReceivedLater because these packets were perhaps received late not recovered by retx or fec.

webrtc-stats.html
```diff
@@ -917,6 +921,41 @@ <h2>
...
+              <dt>
+              <dt>
+                <dfn>packetsReportedAsLost</dfn> of type <span class="idlMemberType">unsigned long long</span>
+              </dt>
+              <dd>
+                <p>
+                  Total number of RTP packets for which an [[RFC8888]] section 3.1 report has been sent with a zero R bit.
+                  Only reported if support for the "ccfb" feedback mechanism has been negotiated.
+                </p>
```
vr000m, line 948: Should we soften the metric from *AsLost to *NotReceived because R={0,1} can occur due to re-ordering. fractionLost and other packetLost metrics already capture the RTP/RTCP metrics.

webrtc-stats.html
```diff
@@ -3114,6 +3170,8 @@ <h3>
              RTCDtlsRole           dtlsRole;
              DOMString             srtpCipher;
              unsigned long         selectedCandidatePairChanges;
+             unsigned long         ccfbMessagesSent;
+	     unsigned long         ccfbMessagesReceived;
```
jan-ivar, line 3174: Bike-shedding: unsure about "messages" (RFC 8888 calls them packets)

The first one counts feedback sent from the receiver to the sender, is that right?

Might it be clearer to call them something like:
```suggestion
             unsigned long         rtcpCcFeedbackSent;
             unsigned long         rtcpCcFeedbackReceived;
```
to clarify RTCP goes in the other direction (and leaning on the word "feedback")?

alvestrand: The RFC uses "messages" in section 3.1
The RFC uses "ccfb" as the identifier in SDP, I don't want to deviate from that.

jan-ivar (approved):

fippo:
This is currently defined in `RTCSentRtpStreamStats` which means `RTCRemoteOutboundRtpStreamStats` inherits from it.
Since this feedback can't be carried in SRs should `packetsSentWithEct1` move to `RTCOutboundRtpStreamStats`?

alvestrand:
@fippo please raise an issue since this is post-merge (if I didn't fix it at the last minute). I assume that this is about packetsSentWitEct1 only.

Issue #619, opened 2022-01-26
WPT tests are wrong about when "outbound-rtp" and "inbound-rtp" stats appear


jan-ivar:
The [spec](https://w3c.github.io/webrtc-stats/#the-rtp-statistics-hierarchy) says: _"The lifetime of all RTP [monitored objects](https://w3c.github.io/webrtc-stats/#dfn-monitored-object) starts when the [RTP stream](https://w3c.github.io/webrtc-stats/#dfn-rtp-stream) is first used: When the first RTP packet is sent or received on the [SSRC](https://w3c.github.io/webrtc-stats/#dfn-ssrc) it represents, or when the first RTCP packet is sent or received that refers to the SSRC of the RTP stream."_

TL;DR: no packets no stats.

Yet these WPT tests assume stats appear immediately after negotiation (ahead of `"connected"` even):
* [RTCPeerConnection-track-stats.https.html](https://github.com/web-platform-tests/wpt/blob/7b0ebaccc62b566a1965396e5be7bb2bc06f841f/webrtc/RTCPeerConnection-track-stats.https.html#L106)
* [RTCRtpSender-getStats.https.html](https://github.com/web-platform-tests/wpt/blob/7b0ebaccc62b566a1965396e5be7bb2bc06f841f/webrtc/RTCRtpSender-getStats.https.html#L56)
* [RTCPeerConnection-track-stats.https.html](https://github.com/web-platform-tests/wpt/blob/7b0ebaccc62b566a1965396e5be7bb2bc06f841f/webrtc/RTCPeerConnection-track-stats.https.html#L141)

jan-ivar:
I've filed vendor bugs [crbug 1291019](https://crbug.com/1291019) and [webkit 235619](https://bugs.webkit.org/show_bug.cgi?id=235619).

Firefox has it right for `outbound-rtp` but not `inbound-rtp`, which is [bugzilla 1751532](https://bugzil.la/1751532).

fippo:
I do not think the spec definition is very useful. It would have prevented detection of the infamous ["macos audio bug"](https://bugs.chromium.org/p/webrtc/issues/detail?id=4799) which showed up as "no bytes sent".

I do agree that adding stats before negotiation is probably not useful though as in the (rare) case of a m-line getting rejected you would still retain a stats entry for it. At implementation level this can be determined by whether a payload type was negotiated or not.

(and yes, this is testable e.g. by your fiddles :-))

[...]

PR #794, opened 2025-02-05
Add PSNR (Y/U/V) for outbound-rtp

fippo:
This is similar to qpSum but codec-independent.

Since PSNR requires additional computation it is defined with an
accompanying psnrMeasurements counter to allow the computation of
an average PSNR.

Defined as a record with components for the Y, U and V planes respectively.

See also
  https://datatracker.ietf.org/doc/html/rfc8761#section-5

***
<a href="https://pr-preview.s3.amazonaws.com/fippo/webrtc-stats/pull/794.html" title="Last updated on Jul 10, 2025, 2:41 PM UTC (18f34e5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/794/02094bc...fippo:18f34e5.html" title="Last updated on Jul 10, 2025, 2:41 PM UTC (18f34e5)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2247,11 +2247,11 @@ enum RTCStatsType {
                 </p>
                 <p>
                   The PSNR is defined in [[ISO-29170-1:2017]].
+                  The frequency of PSNR measurements is [=implementation-defined=].
                 </p>
                 <p class="note">
 		  PSNR metrics should primarily be used as a basis for statistical analysis rather
 		  than be used as an absolute truth on a per-frame basis.
-                  The frequency of PSNR measurements is [=implementation-defined=].
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2232,7 +2232,7 @@ enum RTCStatsType {
                   The count of measurements is in {{psnrMeasurements}}.
                 </p>
                 <p>
-                  PSNR is defined in [[!iso-29170-1:2017]].
+                  PSNR is defined in [[ISO-29170-1:2017]].
                 </p>
               </dd>
               <dt>
@@ -2246,7 +2246,7 @@ enum RTCStatsType {
                   aggregated with this measurement.
                 </p>
                 <p>
-                  The PSNR is defined in [[!iso-29170-1:2017]].
+                  The PSNR is defined in [[ISO-29170-1:2017]].
                 </p>
                 <p class="note">
                   The frequency of PSNR measurements is outside the scope of this specification.
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2259,6 +2259,11 @@ enum RTCStatsType {
                   The PSNR is defined in [[ISO-29170-1:2017]].
                   The frequency of PSNR measurements is [=implementation-defined=].
                 </p>
+                <p class="note">
+                  User agents are encouraged to measure PSNR at a fixed frequency
+                  for each encoder implementation, and as frequently as possible as
+                  long as it does not introduce a noticeable performance penalty.
+                </p>
               </dd>
               <dt>
                 <dfn>totalEncodeTime</dfn> of type <span class=
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2225,6 +2225,9 @@ enum RTCStatsType {
                 record&lt;DOMString, double&gt;</span>
               </dt>
               <dd>
+                <p class="fingerprint">
+                  MUST NOT [= map/exist =] unless [= exposing hardware is allowed =].
+                </p>
                 <p>
                   MUST NOT [= map/exist =] for audio.
                   The cumulative sum of the PSNR values of frames encoded by this sender.
@@ -2244,6 +2247,9 @@ enum RTCStatsType {
                 </span>
               </dt>
               <dd>
+                <p class="fingerprint">
+                  MUST NOT [= map/exist =] unless [= exposing hardware is allowed =].
+                </p>
                 <p>
                   MUST NOT [= map/exist =] for audio.
                   The number of times PSNR was measured. The components of {{psnrSum}} are
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2249,7 +2249,9 @@ enum RTCStatsType {
                   The PSNR is defined in [[ISO-29170-1:2017]].
                 </p>
                 <p class="note">
-                  The frequency of PSNR measurements is outside the scope of this specification.
+		  PSNR metrics should primarily be used as a basis for statistical analysis rather
+		  than be used as an absolute truth on a per-frame basis.
+                  The frequency of PSNR measurements is [=implementation-defined=].
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1966,6 +1966,10 @@ enum RTCStatsType {
              unsigned long        framesEncoded;
              unsigned long        keyFramesEncoded;
              unsigned long long   qpSum;
+             double               psnrSumY;
+             double               psnrSumU;
+             double               psnrSumV;
+             unsigned long long   psnrMeasurements;
              double               totalEncodeTime;
              double               totalPacketSendDelay;
              RTCQualityLimitationReason                 qualityLimitationReason;
@@ -2216,6 +2220,37 @@ enum RTCStatsType {
                 <p>
                   Note that the QP value is only an indication of quantizer values used; many
                   formats have ways to vary the quantizer value within the frame.
+                </p>
+              </dd>
+              <dt>
+                <dfn>psnrSumY</dfn>, <dfn>psnrSumU</dfn>, <dfn>psnrSumV</dfn> of type
+                <span class="idlMemberType">double</span>
+              </dt>
+              <dd>
+                <p>
+                  MUST NOT [= map/exist =] for audio.
+                  The cumulative sum of the PSNR values of frames encoded by this sender, for Y/U/V components.
+                  The count of measurements is in {{psnrMeasurements}}.
+                </p>
+                <p>
+                  PSNR is defined in [[!iso-29170-1:2017]].
+                </p>
+              </dd>
+              <dt>
+                <dfn>psnrMeasurements</dfn> of type <span class="idlMemberType">unsigned long long
+                </span>
+              </dt>
+              <dd>
+                <p>
+                  MUST NOT [= map/exist =] for audio.
+                  The number of times PSNR was measured. {{psnrSumY}}, {{psnrSumU}} and {{psnrSumV}} are
+                  aggregated with this measurement.
+                </p>
+                <p>
+                  The PSNR is defined in [[!iso-29170-1:2017]].
+                </p>
+                <p class="note">
+                  The frequency of PSNR measurements is outside the scope of this specification.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2235,8 +2235,8 @@ enum RTCStatsType {
                   PSNR is defined in [[ISO-29170-1:2017]].
                 </p>
                 <p class="note">
-		  PSNR metrics should primarily be used as a basis for statistical analysis rather
-		  than be used as an absolute truth on a per-frame basis.
+		  PSNR metrics are intended for statistical analysis rather
+		  than representing absolute truth on a per-frame basis.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1966,9 +1966,7 @@ enum RTCStatsType {
              unsigned long        framesEncoded;
              unsigned long        keyFramesEncoded;
              unsigned long long   qpSum;
-             double               psnrSumY;
-             double               psnrSumU;
-             double               psnrSumV;
+             record&lt;DOMString, double&gt; psnrSum;
              unsigned long long   psnrMeasurements;
              double               totalEncodeTime;
              double               totalPacketSendDelay;
@@ -2223,13 +2221,14 @@ enum RTCStatsType {
                 </p>
               </dd>
               <dt>
-                <dfn>psnrSumY</dfn>, <dfn>psnrSumU</dfn>, <dfn>psnrSumV</dfn> of type
-                <span class="idlMemberType">double</span>
+                <dfn>psnrSum</dfn> of type <span class="idlMemberType">
+                record&lt;DOMString, double&gt;</span>
               </dt>
               <dd>
                 <p>
                   MUST NOT [= map/exist =] for audio.
-                  The cumulative sum of the PSNR values of frames encoded by this sender, for Y/U/V components.
+                  The cumulative sum of the PSNR values of frames encoded by this sender.
+                  The record includes values for the "y", "u" and "v" components.
                   The count of measurements is in {{psnrMeasurements}}.
                 </p>
                 <p>
@@ -2243,7 +2242,7 @@ enum RTCStatsType {
               <dd>
                 <p>
                   MUST NOT [= map/exist =] for audio.
-                  The number of times PSNR was measured. {{psnrSumY}}, {{psnrSumU}} and {{psnrSumV}} are
+                  The number of times PSNR was measured. The components of {{psnrSum}} are
                   aggregated with this measurement.
                 </p>
                 <p>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2234,6 +2234,10 @@ enum RTCStatsType {
                 <p>
                   PSNR is defined in [[ISO-29170-1:2017]].
                 </p>
+                <p class="note">
+		  PSNR metrics should primarily be used as a basis for statistical analysis rather
+		  than be used as an absolute truth on a per-frame basis.
+                </p>
               </dd>
               <dt>
                 <dfn>psnrMeasurements</dfn> of type <span class="idlMemberType">unsigned long long
@@ -2248,10 +2252,6 @@ enum RTCStatsType {
                 <p>
                   The PSNR is defined in [[ISO-29170-1:2017]].
                   The frequency of PSNR measurements is [=implementation-defined=].
-                </p>
-                <p class="note">
-		  PSNR metrics should primarily be used as a basis for statistical analysis rather
-		  than be used as an absolute truth on a per-frame basis.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.js
+++ b/webrtc-stats.js
@@ -99,6 +99,11 @@ var respecConfig = {
             status:   "Internet Draft",
             publisher:  "IETF"
         },
+	"iso-29170-1:2017": {
+	  title: "Information technology — Advanced image coding and evaluation",
+	  href: "https://www.iso.org/standard/63637.html",
+	  publisher: "ISO",
+	},
       },
   postProcess: [
     function generateStatsHierarchy(config, doc) {
--- a/webrtc-stats.js
+++ b/webrtc-stats.js
@@ -99,7 +99,7 @@ var respecConfig = {
             status:   "Internet Draft",
             publisher:  "IETF"
         },
-	"iso-29170-1:2017": {
+	"ISO-29170-1:2017": {
 	  title: "Information technology — Advanced image coding and evaluation",
 	  href: "https://www.iso.org/standard/63637.html",
 	  publisher: "ISO",


fippo:
https://www.researchgate.net/publication/383545049_Low-Complexity_Video_PSNR_Measurement_in_Real-Time_Communication_Products has a whole paper about this.

tl;dr is "gp is codec dependent", PSNR is not (but comes at a cost hence this can not be a simple sum)

@youennf the folks who implemented https://developer.apple.com/documentation/videotoolbox/kvtcompressionpropertykey_calculatemeansquarederror?changes=l_4_8 might be able to tell you more too.

cc @taste1981

fippo:
This one is encoder PSNR, not scaling PSNR. Scaling PSNR would end up living on media-source in stats I would have.
I'll have a copy of the paper sent to you

webrtc-stats.html
```diff
@@ -2218,6 +2220,38 @@ <h2>
...
+                <p>
+                  MUST NOT [= map/exist =] for audio.
+                  The number of times PSNR was measured. The components of {{psnrSum}} are
+                  aggregated with this measurement.
+                </p>
+                <p>
+                  The PSNR is defined in [[ISO-29170-1:2017]].
+                </p>
+                <p class="note">
+                  The frequency of PSNR measurements is outside the scope of this specification.
```
vr000m, line 2252: We should recommend that perceptually, Y-PSNR is used. And PSNR values can vary based on motion and encoder settings.

Typically, Average Y-PSNR is calculated by dividing with the number of frames.

vr000m:
The other thing that I am curious about is if the PSNR requires decoding the encoded video or is this calculated as part of the encoder operation. Mainly the impact on CPU if it requires some kind of decode step, I wonder if this is only calculated applied to I-frames or huge-frames.

webrtc-stats.html
alvestrand: that division is done by the application.

alvestrand:
@sprangerik have you looked at this?

jan-ivar:
@jesup thoughts?

sprangerik:
I am supportive of this.

For context, see also https://webrtc-review.googlesource.com/c/src/+/368960

sprangerik:
> The other thing that I am curious about is if the PSNR requires decoding the encoded video or is this calculated as part of the encoder operation. Mainly the impact on CPU if it requires some kind of decode step, I wonder if this is only calculated applied to I-frames or huge-frames.

The idea is to have it as part of the encoder process. The encoder is by definition also a decoder, so it can directly use both the raw input and reconstructed state without penalty. The actual PSNR calculation will of course often incur an extra CPU hit, unless it is already a part of e.g. a rate-distortion aware rate controller - but that's not often the case for real-time encoders. That's why it's proposed to limit the frequency of PSNR calculations.

This of course means the user cannot count on PSNR metrics being populated. Even for a given stream, the PSNR values might suddenly disappear if e.g. there is a software/hardware switching event and only one implementation supports PSNR output.

vr000m:
since webrtc-stats anyway does aggregate values, we could do a sumPsnr and countFrames, i.e., each time a psnr is calculated it is added and corresponding frame count counter goes up. If it is done for all frames, we would not need a frame counter

Drekabi:
This seems like a nice feature that could have a few uses. I do wonder if it could be a separate API instead of part of the outbound RTP stats. My initial concerns are calculating this data regardless if the application is even interested in the data and no specification or recommendations on frequency of measurements.

Some pros and cons that come to mind if this were implemented as a separate API instead.
Pros:
- This would allow for those interested in the information to request a PSNR directly. 
- Eliminates the need to constantly be collecting PSNR measurements that may not be used.
- It would allow for applications to have some control over the frequency of PSNR measurements.

Cons:
- Applications would not be getting PSNR sum for free and need to calculate and store this on their own.

If this should remain in the stats could we consider adding some sort of getStats object to enable logging for this kind of data?

alvestrand:
WebRTC users routinely log getStats data, so adding this would not be any big overhead. If the stats are collected on a timescale of seconds, the overhead is usually negligible. (polling stats for every frame is not a good idea).
I presume that the stats would not be populated if not implemented or enabled; I don't know if it's worth it to expose a control to turn computation on and off on a track.

jan-ivar:
If I understand correctly, the concern of overhead is in the browser doing an expensive calculation most websites would never request (though per-frame is not an issue due to caching, per-second might be; is _never_ an acceptable frequency?)

How expensive is this computation? Our webstats model is like a boat we keep loading with new stuff. Eventually, it becomes problematic. 

At some point (maybe now?) might we wish we had something like this?
```js
await sender.getStats({verbosity: "high"}) // low | medium (default) | high
```

henbos:
Btw this is unrelated to the polling frequency since the metrics only update when a measurement is made and a measurement happens in the background whether or not the app is polling getStats. (Polling getStats several times per second is bad because of the overhead of that call, not because of counters incrementing in the background)

jan-ivar:
One concern is this stat seems to require making two getStats call over some interval.

E.g. is the use case here to try one encoder setting, get stats, then wait 1 second and call getStats again expecting two different measurements? If so this might cause divide by zero error in one browser but not another.

henbos:
qpSum / framesDecoded might be a better example of a foot gun since that could fail when network glitches but not in stable environment.

In practice web developers will make helper functions that does lookup of deltas and rates taking care of foot guns.

Also you have to be prepared for a metric not being present all of a sudden

jan-ivar:
Yes I didn't mean to suggest the divide by zero hazard was limited to this API. The difference is something like `framesDecoded` is tied to real-world reception of RTP which won't vary by user agent.

I think the concern in this case is:
1. lack of cap on frequency might lead two browsers to implement vastly different intervals, e.g. 1 vs 15 seconds.
2. whether an anticipated use case is A/B codec testing where a website might tune its interval to the fastest browser

fippo:
PSNR is similar to qp so having it in getStats makes sense.

As the paper says we have done this at a frequency higher than one per second on devices where battery consumption is a concern and it works there. Hardware encoder support makes this "cheaper" even. Note that the calculation is done by the encoder so can not be triggered by calling getStats with some magic option. I considered whether it was possible to gate it on the corruption detection RTP header extension but that would have been quite awkward since it is not closely related (not without precedence, quite a few statistics depend on header extensions)

When I say "A/B testing" consider a project like [Jitsi moving to AV1](https://jitsi.org/blog/av1-and-more-how-does-jitsi-meet-pick-video-codecs/), in particular the "Metrics Captured" which, unsurprisingly, relies on getStats. See [here](https://youtu.be/zWvteeEkjJg?si=5wYeyGbNCXjjb7mF&t=465) for one uses PSNR to evaluate when it is available.

Such experiments are designed not to compare :apple:  to :banana: (different browsers, different operating systems) so letting a UA decide on sampling frequency is not a concern as long as it does so consistently.

henbos:
> I think the concern in this case is:
>
> 1. lack of cap on frequency might lead two browsers to implement vastly different intervals, e.g. 1 vs 15 seconds.
> 2. whether an anticipated use case is A/B codec testing where a website might tune its interval to the fastest browser

Polling is just asking "do you have any new measurements for me?"
Whether I poll once per 15 seconds and get that measurement on the first try or if I poll getStats 15 times and get 14 "no's" and 1 "yes", I think I end up in the same situation, so what exactly was the problem there (and why is this different from the other hundred+ metrics we have)?

It doesn't matter if app polling interval and browser polling interval aligns or not and it's clear from the [guidelines](https://w3c.github.io/webrtc-stats/#guidelines-for-design-of-stats-objects) that there is no control of sampling period. So I would argue that the only thing that matters is if the measurements are arriving at a granular enough level to be useful.

If the concern is that a browser implementer doesn't know what a useful measurement interval is, maybe we can provide some guidance there, but I fail to see the interop issue with different polling intervals that are all within a "useful" range. FTR I think 15 second is too large of an interval since a lot can happen in that period of time.

fippo:
I would not even poll getStats for A/B testing purposes. One would typically poll periodically and use the last result or call getStats explicitly **before** closing the peerconnection and then calculate the average PSNR as psnrSum_{y,u,v}/psnrMeasurements. Only calls with enough psnrMeasurements should be taken into account which one needs to irrespective of sampling frequency to exclude "short calls".

(while we are rambling: it seems Firefox throws when calling getStats on a closed peerconnection which is not my understanding of #3 arguably with all the transceivers gone all the interesting stats disappear *nowadays*)

jan-ivar:
> > I think the concern in this case is:
> >  ...
> > 2. whether an anticipated use case is A/B codec testing where a website might tune its interval to the fastest browser
> 
> Polling is just asking "do you have any new measurements for me?" Whether I poll once per 15 seconds and get that measurement on the first try or if I poll getStats 15 times and get 14 "no's" and 1 "yes", I think I end up in the same situation, so what exactly was the problem there

That's fine for telemetry. I think our concern was more someone making *runtime* decisions off stats, e.g.:
```js
// probe and switch to best codec for media being sent right now:
let bestCodec, bestY = 0;
for (const codec of sender.getParameters().codecs) {
  const params = sender.getParameters();
  params.encodings[0].codec = codec;
  await sender.setParameters(params);
  await wait(1000);
  const ortp1 = [...(await sender.getStats()).values()].find(({type}) => type == "outbound-rtp");
  await wait(1000);
  const ortp2 = [...(await sender.getStats()).values()].find(({type}) => type == "outbound-rtp");
  const y = (ortp2.psnrSum.y - ortp1.psnrSum.y) / (ortp2.psnrMeasurements - ortp1.psnrMeasurements);
  if (bestY < y) { bestY = y; bestCodec = codec;
}
const params = sender.getParameters();
params.encodings[0].codec = bestCodec; }
await sender.setParameters(params);
```

 > (and why is this different from the other hundred+ metrics we have)?

No other stat has this note:
image

> FTR I think 15 second is too large of an interval since a lot can happen in that period of time.

This might be good for an implementer to know.

fippo:
I can replace that with "is implementation-defined" linking to https://infra.spec.whatwg.org/#implementation-defined

henbos:
We could say that PSNR measurement frequency is implementation-specific but that "the user agent SHOULD make make PSNR measurements no less frequently than every X seconds". I still think the app needs to handle the case that a PSNR measurement is not available for a given codec or browser though, but this would give the app an upper bound on how long to maximally wait for.

webrtc-stats.html
```diff
@@ -2218,6 +2220,40 @@ <h2>
...
+              </dt>
+              <dd>
+                <p>
+                  MUST NOT [= map/exist =] for audio.
+                  The number of times PSNR was measured. The components of {{psnrSum}} are
+                  aggregated with this measurement.
+                </p>
+                <p>
+                  The PSNR is defined in [[ISO-29170-1:2017]].
+                </p>
```
jan-ivar, line 2250: This seems redundant with the same text under `psnrSum` (line 2235) which is already link to here.

```suggestion
```

webrtc-stats.html
```diff
@@ -2218,6 +2220,40 @@ <h2>
...
+                <p class="note">
+		  PSNR metrics should primarily be used as a basis for statistical analysis rather
+		  than be used as an absolute truth on a per-frame basis.
+                  The frequency of PSNR measurements is [=implementation-defined=].
+                </p>
```
jan-ivar, line 2255: We want to avoid "should" and normative statements in non-normative notes. If we want to be normative we should lift it out.

Based on https://github.com/w3c/webrtc-stats/pull/794#issuecomment-2693542664 are we comfortable picking a min frequency of every 5 seconds?

```suggestion
                <p>
                  If the current encoder supports taking PSNR measurements, their
                  frequency SHOULD be no less than every 5 seconds.
                </p>
                <p class="note">
                  This allows for testing. PSNR measurements are intended for
                  statistical analysis, and aren't expected to be accurate down
                  to a frame.
                </p>
```

jan-ivar (changes requested):

webrtc-stats.html
fippo: Can you explain why 5 seconds? `keyFramesEncoded` and `keyFramesDecoded` are examples very-low-frequency events happening already where you can not make an assumption about the minimum interval between two getStats calls that will give you an increase. Same for packetsLost.

jan-ivar: What value would you like? The frequency of `keyFramesEncoded` and `keyFramesDecoded` are determined by external factors unlikely to vary by user agent.

jan-ivar: 5 is arbitrary. In this thread I heard 15 seconds was too high, and that 1 second was not too expensive, but also that we'd rather not overconstrain implementations too much. 10 seconds?

This value, whatever we pick would go into WPT and also give web developers that can't wait for some reason a minimum time to wait to be interoperable.

jan-ivar: Hmm, maybe we need to also qualify that the encoder is actually encoding something? E.g. if the track is muted or a canvas track, then the frame rate may be less than what value we pick. Maybe we add _"... or the encoding frame rate whichever is lower"_?

fippo: tl;dr: this is a parameter of the video encoder configuration. It becomes *observable* in stats just like others.
What the spec should say is "implementation defined" - the implied warning is "do not compare apples to oranges"

The frequency of keyFramesEncoded is a good example, it is determined by the encoder setting such as gop size. One wants this as large as possible for good performance but this is something where the encoder can be tuned without "interoperability" constraints since the decoder behavior is required to be flexible.

If one wants to get fancy about PSNR one may need to take into account things like "is this a screen sharing track" (higher resolution, lower frame rate and wholly different content) or frame rate (which may depend on BWE). Worth the effort for statistical analysis? Unlikely.

15 seconds was arbitrary too I think with the purpose of "i dont think this makes sense anymore". One second picked in the code is 3x the value from the paper (which has a parameter study) and hence 1/3rd as expensive in terms of power impact.

jan-ivar: A SHOULD implies implementation-defined. Is there a value that would be satisfactory?

youennf:
If we have the psnr measurement frequency be implementation-defined, the spec should help the UA developer select a good frequency value. Spec should add some guidelines, for instance doing measurements every second or every 30 frames or so.

fippo:
> From editors meeting: Main reason for a number seems to be feature detection - people want to know that if they have waited this long between calls to GetStats, and the frame counter has increased by X, and the stat doesn't show up, the feature is off.

Huh? This is a counter, it can not disappear, it canonly stop increasing.

jan-ivar:
> I think the gist of it is that we want the frequency to be a high as possible as long as the performance impact can be kept negligible.

That's great, because I don't think anyone is proposing a ceiling on frequency. A floor would be nice though.

I disagree giving guidance is bad, since what you wrote sounds like good guidance already. Let's write it down.

fippo:
As we say in German, "Guter Rat ist teuer". The guidance should "talk to your video encoding guys". That is what "implementation-defined" implies.

youennf:
`talk to your video encoding guys` seems to induce that the rate selection may be encoder specific.

This naturally leads to privacy questions if the selection of the rate depends on the encoder, the system load, the CPU...
I only see benefits in the spec defining the implementor playground.

fippo:
> if the selection of the rate depends on the encoder, the system load, the CPU...

Who is proposing this?

> This naturally leads to privacy questions 

How is this a new concern given video codecs in WebRTC already are adapting to CPU, thermal state, etc? Which is not described by any specification either. Note that PSNR should be possible to compute with JS and existing APIs already.

And you are most welcome to invite privacy guys into your meetings obviously.

webrtc-stats.html
fippo: Moved the normative statement out of the note (and moved the note to the actual values)

webrtc-stats.html
```diff
@@ -2218,6 +2220,40 @@ <h2>
...
+                  MUST NOT [= map/exist =] for audio.
+                  The cumulative sum of the PSNR values of frames encoded by this sender.
+                  The record includes values for the "y", "u" and "v" components.
+                  The count of measurements is in {{psnrMeasurements}}.
+                </p>
+                <p>
+                  PSNR is defined in [[ISO-29170-1:2017]].
+                </p>
+                <p class="note">
+		  PSNR metrics should primarily be used as a basis for statistical analysis rather
```
jan-ivar, line 2238: Avoid "should" in non-normative notes. How about
```suggestion
		  Authors are expected to use PSNR metrics primarily as a basis for statistical analysis rather
```

webrtc-stats.html
```diff
@@ -2218,6 +2220,40 @@ <h2>
...
+              </dt>
+              <dd>
+                <p>
+                  MUST NOT [= map/exist =] for audio.
+                  The number of times PSNR was measured. The components of {{psnrSum}} are
+                  aggregated with this measurement.
+                </p>
+                <p>
+                  The PSNR is defined in [[ISO-29170-1:2017]].
+                  The frequency of PSNR measurements is [=implementation-defined=].
```
jan-ivar, line 2254: OK with me if for any reasonable value of X (I'm not married to any particular value, as long as we have one that lets us WPT test) 
```suggestion
                  The frequency of PSNR measurements is [=implementation-defined=],
                  but SHOULD be no less than every X seconds.
```

webrtc-stats.html
fippo: https://w3c.github.io/webrtc-stats/#dom-rtcinboundrtpstreamstats-totalcorruptionprobability -- literally copied from this note, can you explain why it is ok there but not here?

webrtc-stats.html
fippo: A single value is sufficient for existence and can be used to write a WPT.

webrtc-stats.html
jan-ivar: I missed it in review of https://github.com/w3c/webrtc-stats/pull/788. It's not a [hard rule](https://w3c.github.io/webrtc-stats/#conformance), but most WG editors I've spoken with agree avoiding lowercase requirement-laden words avoids confusion and improves readability.

Both notes appear to be speaking to [authors](https://w3ctag.github.io/design-principles/#priority-of-constituencies) rather than implementers, which is fine. But the primary audience of specs are implementers, so it might help to clarify when speaking to someone else. Specs have no authority saying authors should or shouldn't do anything. So it seems more accurate to describe the usage the design anticipates, which is how I interpret these notes.

webrtc-stats.html
jan-ivar: The argument here is a WPT testing an interoperable floor on frequency, not just existence.

> I only see benefits in the spec defining the implementor playground.

I believe @youennf also sought some implementer guidance here so we don't have to reverse engineer interoperable behavior.

If we can't resolve this in the PR let's aim to discuss this again next meeting.

webrtc-stats.html
fippo: No, this note documents that implementers (well, the one) do not think authors should be abusing this for per-frame analysis. 810d67d6b4894aba56c8404380baa497dbfb2110 avoids lower-case should.

webrtc-stats.html
fippo: I still have not seen an argument why the minimum frequency needs to be interoperable.
Also I assume you mean "intraoperable"?

jan-ivar: I mean [interoperable](https://www.w3.org/policies/process/#implementation-experience), as in websites written and tested in one browser won't break in another.

I gave an example above in https://github.com/w3c/webrtc-stats/pull/794#issuecomment-2691600502 showing this is JS-observable implementation-defined behavior. A note doesn't change that.

[§ 11.2. Specify completely and avoid ambiguity](https://w3ctag.github.io/design-principles/#avoid-ambiguity) emphasizes that every feature should be sufficiently specified so there is no room for unpredictable or diverging per-implementation decisions.

This suggests we need good reasons to _tolerate_ implementation-defined behavior, not the other way around.

I haven't heard any argument why implementations need this much flexibility to lower their frequency of measurement that we can't specify a floor at all. This strikes me as unreasonable.

Technically speaking, we'd need a MUST for WPT to test this. So I think asking for a SHOULD, which amounts to little more than minimal implementation guidance, is being quite generous.

Maybe we need two frequencies: one that it SHOULD be on average, and a floor that it MUST be higher than?

jan-ivar: We seem to be talking past each other. No-one's objected to implementation-defined behavior here. A SHOULD *is* implementation-defined behavior. A floor on implementation-defined behavior is still implementation-defined behavior.

I'm asking the reason we can't give implementer guidance, or put any boundaries on it.

fippo: well, then take the implementers feedback: such a number is not required and a specification constraining without a solid data basis and experimentation (such as done by the paper) is not how things are done in video encoding.

jan-ivar: Specs generally don't dictate quality or performance of media output. But that would be like if we put bounds on returned psnr values themselves. This is not what we're talking about.

The shape of `getStats` requires websites take two or more samples to divide by a the time in between. This creates a potential interoperability issue with obtaining valid measurements over short periods of time for this stat specifically (even if this is not the intended use) if the minimum measurement interval varies significantly by implementation. No expertise on psnr is needed to observe this.

Understanding that delivering measurements can be costly, I'd like feedback on how a generous 15 second upper cap on measurement interval would impose on implementers. The number is inherently arbitrary as it is picked to be outside the bounds of reasonable implementations.

If we can't agree on a number, some guidance to implementers and authors that would lead them to correctly guess measurement cadence as suggested in this thread (~every second, rarely ever more than every 5) would also do the job I think.

fippo: How do you propose to satisfy this arbitrary when switching from an encoder that exposes PSNR to one that does not?

fippo: Adding more unfounded numbers (unless you have run large scale experiments?) is not going to help.
1 second is quite likely to be an issue for high-resolution screen sharing.

Going once

jan-ivar: Edit: ^ s/every second/every few seconds/

jan-ivar: If situations arise where implementations fall outside these bounds we can always revisit.

fippo: Since a vague soft limit doesn't avoid the "numerator is zero" what problem does this solve?
What is the behavior when the track was removed via replaceTrack? Follow-up question in #619...

jan-ivar: > Since a vague soft limit doesn't avoid the "numerator is zero" what problem does this solve?

@youennf does it address your concern in https://github.com/w3c/webrtc-stats/pull/794#issuecomment-2721716590?

> What is the behavior when the track was removed via replaceTrack?

The encoding frame rate drops to zero, which is lower than 1/15, and no PSNR measurements happen.

youennf:
My understanding is that, if the encoder is providing PSNR, there is probably no perf issue in doing so for every frame.
For encoders that do not support PSNR, it is ok to not expose that value (modulo maybe a fingerprinting note).

Based on this, I would tend to remove `psnrMeasurements` and just keep `psnrSum`.

vr000m:
I think psnrSum/psnrMeasurements gives us the information even in periods it's not updated. Ie if an app is calling getstats regularly, ie every second or every 2 seconds no updates in psnrSum and measurements already gives us info, that no updates were made.

youennf:
> Getting the values are still valuable, but we don't want to enable it for every frame as that will come at a noticeable CPU penalty.

The fact that some encoders can expose or not PSNR would be new information exposed to the web and could be used for fingerprinting.

Also, this PR gives no implementor's guideline. My assumption is that a single measurement frequency would be used for all encoders, this frequency value would be fixed for a given UA instance, and probably for a given UA across all devices it runs on (say a specific version of Chrome). It would be good to clarify this, otherwise I could see potential additional threats. Maybe the PING WG should weigh in there.

sprangerik:
> The fact that some encoders can expose or not PSNR would be new information exposed to the web and could be used for fingerprinting.

This would map essentially 1:1 with the implementation used, and that can already pretty easily be inferred (e.g. via https://www.w3.org/TR/webrtc-stats/#dom-rtcoutboundrtpstreamstats-encoderimplementation or platform+https://www.w3.org/TR/webrtc-stats/#dom-rtcoutboundrtpstreamstats-powerefficientencoder, not to mention info from WebCodecs,  WebGPU, parsing data from encoded transform, etc etc). So while it might be a new "bit", it doesn't actually provide any new information imo.

> Also, this PR gives no implementor's guideline. My assumption is that a single measurement frequency would be used for all encoders, this frequency value would be fixed for a given UA instance, and probably for a given UA across all devices it runs on (say a specific version of Chrome). It would be good to clarify this, otherwise I could see potential additional threats. 

Can we let the implementor's guideline just be along what has been said above, e.g. "the frequency should be as high as possible as long as the performance impact can be kept negligible". I don't see a reason to change the frequency based on codec type, only by implementation performance overhead. For a given implementation though I don't see a reason to change the frequency - detailing that this should be fixed for a given UA seems fine to me.

youennf:
> This would map essentially 1:1 with the implementation used, and that can already pretty easily be inferred (e.g. via https://www.w3.org/TR/webrtc-stats/#dom-rtcoutboundrtpstreamstats-encoderimplementation or platform+https://www.w3.org/TR/webrtc-stats/#dom-rtcoutboundrtpstreamstats-powerefficientencoder, n

Both `encoderImplementation` and `powerEfficientEncoder` have an exposure note: `must not exist unless exposing hardware is allowed`.

A possibility is to restrict psnr in the same manner.
In that case, it seems ok to have per frame PSNR for hardware encoders and less frequent PSNR for software encoders.

> This should be fixed for a given UA seems fine to me.

I find it useful information.

jan-ivar:
> This would map essentially 1:1 with the implementation used, and that can already pretty easily be inferred (e.g. via https://www.w3.org/TR/webrtc-stats/#dom-rtcoutboundrtpstreamstats-encoderimplementation or platform+https://www.w3.org/TR/webrtc-stats/#dom-rtcoutboundrtpstreamstats-powerefficientencoder,

Those don't seem like great examples as they're blocked on [exposing hardware is allowed](https://www.w3.org/TR/webrtc-stats/#dfn-exposing-hardware-is-allowed), unless we're suggesting adding that requirement here? What are the other examples?

> My assumption is that a single measurement frequency would be used for all encoders, this frequency value would be fixed for a given UA instance, and probably for a given UA across all devices it runs on (say a specific version of Chrome). It would be good to clarify this, otherwise I could see potential additional threats.

Agreed. Clarifying these assumptions in the guidance can only help.

> Can we let the implementor's guideline just be along what has been said above, e.g. "the frequency should be as high as possible as long as the performance impact can be kept negligible". I don't see a reason to change the frequency based on codec type, only by implementation performance overhead. For a given implementation though I don't see a reason to change the frequency - detailing that this should be fixed for a given UA seems fine to me.

Doesn't tying it too tightly to performance make it another performance metric? I like the part that it should not vary by codec.

fippo:
> In that case, it seems ok to have per frame PSNR for hardware encoders and less frequent PSNR for software encoders.

Even on hardware encoders it has an impact on power consumption and the return on doing it on every frame is not there, see the parts in the paper that talk about subsampling.

I'm fine with gating on HW.

youennf:
> Even on hardware encoders it has an impact on power consumption

If so, PSNR gathering could be opt-in, something like:

```
dictionary StatsGatheringOptions {
    float psnrPreferredFrequency = 0;
};
partial interface RTCPeerConnection {
    attribute StatsGatheringOptions statsOptions;
};
```

Is it overkill?

fippo:
> Is it overkill?

Quoting https://w3c.github.io/webrtc-stats/#guidelines-for-design-of-stats-objects:
There is, by design, no control surface for the application to influence how stats are generated.

youennf:
Until now, there was no concern about stats being potentially computer intensive.

sprangerik:
> Those don't seem like great examples as they're blocked on [exposing hardware is allowed](https://www.w3.org/TR/webrtc-stats/#dfn-exposing-hardware-is-allowed), unless we're suggesting adding that requirement here? What are the other examples?

Other examples... If it's "powerEfficient = false" then it's essentially a SW implementation and the implementation will follow from browser type and version. If "powerEfficient = true" it's going to be a hardware encoder, and then in 99.99% of cases map exactly to the GPU used. The GPU can be found in a number of different ways, e.g. using https://developer.mozilla.org/en-US/docs/Web/API/GPUAdapterInfo

Another way I mentioned was using WebCodecs. Simplest example: take a known (non-zero) input image and encode it using a known bitrate, then take the resulting bitstream data array that came out of the encoder and create hash from it. That will basically give you a per-implementation unique signature.

Even if we try to forcefully add some noise to screw with that identification mechanism, there are other aspects that can be used to detected the implementation - e.g. which selection of coding tools that were used, filter parameter selection, cropping behavior, etc etc. You can of course do this with webrtc as well, capturing the bitstream data with an encoded transform instead, it's just a little more unwieldy. 

Once you know which encoder implementation the user has, a bunch of data can inferred (e.g. feature capabilities, performance characteristics, etc) - but that data doesn't add any more fingerprinting surface compared to just knowing the implementation. Behavior when it comes to PSNR implementation would fall into this category as well.

fippo:
@jan-ivar: please summarize your concerns (I am happy to request a standards position if required) and provide alternatives *with implementations*.

I am fine having this in -provisional-stats. This was a *courtesy*, it was quite apparently not even reciprocated with reading the published paper on the subject.

fippo:
gating on HW was already been added 2e3124033bba5b71dd81f6165bfb3db9a99bca3a

Your continued question for "guidance" despite a published paper is surprising, in particular given folks working on encoders are typically very familiar with PSNR and given that Mozilla is effectively a mere consumer of libWebRTC where the implementation of the logic deciding on which frames to request the encoder to calculate PSNR would live.

jan-ivar:
> ...detailing that this should be fixed for a given UA seems fine to me.

I interpreted this to mean keeping the same frequency *across* encoders in the same UA was desirable (modulo HW/SW). Which is it?

webrtc-stats.html
jan-ivar: How's this?
```suggestion
                  The frequency of PSNR measurements is [=implementation-defined=].
                </p>
                <p class="note">
                  User agents are encouraged to measure PSNR at a fixed frequency
                  for each encoder implementation, and as frequently as possibly as
                  long as it does not introduce a noticeable performance penalty.
```

fippo:
> which could be used to add an informal note

Neither of these saves you from talking to a subject matter expert. Or reading the paper. Or doing power consumption measurements.

> keeping the same frequency across encoders in the same UA was desirable (modulo HW/SW)

Where is the value in specifying an arbitrary restriction?

webrtc-stats.html
youennf: LGTM

jan-ivar (approved):
LGTM with the guidance! This should help with interop. Since PSNR measurements can be costly, the lack of a floor seems necessary to not impose arbitrary limits on the UAs ability to remain performant.

fippo:
I look forward to your *independent implementations*!

PR #802, opened 2025-03-27
Update RTP lifetime description: outbound-rtp to be created before first packet is sent

### LINKED ISSUE #797: RTCOutboundRtpStreamStats should be created "early"
> The lifetime of all RTP [monitored object](https://w3c.github.io/webrtc-stats/#dfn-monitored-object)s starts when the [RTP stream](https://w3c.github.io/webrtc-stats/#dfn-rtp-stream) is first used: When the first RTP packet is sent or received on the [SSRC](https://w3c.github.io/webrtc-stats/#dfn-ssrc) it represents, or when the first RTCP packet is sent or received that refers to the [SSRC](https://w3c.github.io/webrtc-stats/#dfn-ssrc) of the [RTP stream](https://w3c.github.io/webrtc-stats/#dfn-rtp-stream).

I assume SSRCs are NOT signalled in the SDP in this post, even though they typically are with existing implementaitons.

While inbound-rtps don't have a known SSRC until the first packet is received, this is NOT true for outbound-rtps where the SSRC is picked by the WebRTC engine as soon as the RTCRtpSender has been "configured" which happens by doing O/A.

LibWebRTC already does what I think makes sense: If 3 encodings are configured, we get 3 outbound-rtps and the SSRCs are known whether or not a packet has been sent yet.

If we implement the spec literally, a simulcast stream would look like a singlecast stream during ramp-up when only the first encoding has started sending yet. Existing web apps wants to know the SSRCs of all RTP streams whether or not they have ramped up yet, or even if they're not going to because the RTCRtpSender's track is null ("mute" use case).

I think this is one of those "easier to align spec with impl than to align impl with spec".

henbos:
Fixes #797.

[Original description] The RTP sender is created at setLocalDescription time, i.e. has-local-offer or stable. This is consistent with the transceiver's MID value also being assigned at this point in time.
- I have verified that this description is consistent with Chrome Stable: [offerer fiddle](https://jsfiddle.net/gy8us4Lk/) and [answerer fiddle](https://jsfiddle.net/5ytoqzum/).

[Update based on review] The RTP sender is configured when we go back to stable signaling state (SDP answer is set), this means both Chrome and Firefox needs to update its behavior

WPT test coverage is being added [here](https://chromium-review.googlesource.com/c/chromium/src/+/6402253).
- Chrome currently passes the first test and fails the second but I'm happy to make Chrome pass both tests.

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-stats/pull/802.html" title="Last updated on Mar 28, 2025, 9:39 AM UTC (ebb579a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/802/d3baab4...henbos:ebb579a.html" title="Last updated on Mar 28, 2025, 9:39 AM UTC (ebb579a)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -620,9 +620,12 @@ enum RTCStatsType {
           RTCP SR or RR block.
         </p>
         <p>
-          The lifetime of all RTP [= monitored object =]s starts when the <a>RTP stream</a> is first
-          used: When the first RTP packet is sent or received on the <a>SSRC</a> it represents, or when
-          the first RTCP packet is sent or received that refers to the <a>SSRC</a> of the <a>RTP stream</a>.
+          The lifetime of all RTP [= monitored object =]s are tied to <a>SSRC</a>s.
+          The {{RTCOutboundRtpStreamStats}} is created when the RTP sender is configured by
+          <code>setLocalDescription()</code>, the {{RTCInboundRtpStreamStats}} is created when
+          the first RTP packet for this SSRC is received and the remote RTP stream stats
+          objects ({{RTCRemoteInboundRtpStreamStats}} and {{RTCRemoteOutboundRtpStreamStats}})
+          are created when the corresponding RTCP packet is first received.
         </p>
         <p>
           RTP monitored objects are deleted when the corresponding RTP sender or
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -622,7 +622,9 @@ enum RTCStatsType {
         <p>
           The lifetime of all RTP [= monitored object =]s are tied to <a>SSRC</a>s.
           The {{RTCOutboundRtpStreamStats}} is created when the RTP sender is configured by
-          <code>setLocalDescription()</code>, the {{RTCInboundRtpStreamStats}} is created when
+          setting a local or remote SDP answer (via <code>setLocalDescription()</code>
+          or <code>setRemoteDescription()</code> and the signaling state returning to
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -620,9 +620,12 @@ <h3>
           RTCP SR or RR block.
         </p>
         <p>
-          The lifetime of all RTP [= monitored object =]s starts when the <a>RTP stream</a> is first
-          used: When the first RTP packet is sent or received on the <a>SSRC</a> it represents, or when
-          the first RTCP packet is sent or received that refers to the <a>SSRC</a> of the <a>RTP stream</a>.
+          The lifetime of all RTP [= monitored object =]s are tied to <a>SSRC</a>s.
+          The {{RTCOutboundRtpStreamStats}} is created when the RTP sender is configured by
+          <code>setLocalDescription()</code>, the {{RTCInboundRtpStreamStats}} is created when
+          the first RTP packet for this SSRC is received and the remote RTP stream stats
```
jan-ivar, line 626: I think in the meeting I agreed with "early" being as early as coming back to `"stable"`, but this is even earlier.

Editors can integrate if we change PR to say that.

jan-ivar (approved):

webrtc-stats.html
henbos: PR updated. I will also update the CL that adds the corresponding web platform tests to do the same.

PR #803, opened 2025-03-27
Add outbound-rtp.encodingIndex.

### LINKED ISSUE #801: RTCOutboundRtpStreamStats.encodingIndex
Apps that don't use RIDs have no good way of knowing which encoding has which SSRC.

Heuristics for this (looking at resolution or frame rate) is not bullet proof because of adaptation or ramp up and is also not very ergonomic. It also remains a mystery for inactive encodings.

Adding an integer for this is cheap for apps that want to know which encoding has which SSRC ahead of time.

henbos:
Fixes #801.

Sorting requires it to be an index. Sorting or not, we need to know which encoding has which SSRC. and different streams may use different RIDs for different semantics, e.g. {q,h,f} = {quarter, half, full} may be used for camera streams and {l,h} = {lowFps, highFps} may be used for screenshare streams. Or something else, we don't want to have coupling between the stats pipeline and arbitrary app naming conventions somewhere else which are not set in stone or consistent.

The stats pipeline is typically a separate part of the code base or even an entirely external tool like https://fippo.github.io/webrtc-dump-importer/ which does not have access to JS APIs like getParameters().

The stats pipeline often executes after the call has ended, so index needs to be stored inside the report for later analysis. Why force the app to have to modify the RTCStatsReport even if it could do it?

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/webrtc-stats/pull/803.html" title="Last updated on Mar 28, 2025, 9:59 AM UTC (4a9229c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webrtc-stats/803/62bad06...henbos:4a9229c.html" title="Last updated on Mar 28, 2025, 9:59 AM UTC (4a9229c)">Diff</a>

### DIFF
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -2098,14 +2098,9 @@ enum RTCStatsType {
               <dd>
                 <p>
                   MUST NOT [= map/exist =] for audio.
-                  Only [= map/exist =]s if this <a>RTP stream</a> belongs to an
-                  RTP sender that has been configured for simulcast, in other
-                  words the {{RTCRtpSender}} has multiple
-                  {{RTCRtpSendParameters/encodings}} and
-                  {{RTCOutboundRtpStreamStats}} objects. The {{encodingIndex}}
-                  is the index of the encoding that represents this <a>RTP
-                  stream</a>. MUST NOT be [= map/exist | present =] for
-                  singlecast senders.
+                  This is the index of the encoding that represents this
+                  <a>RTP stream</a> in the RTP sender's list of
+                  {{RTCRtpSendParameters/encodings}}.
                 </p>
               </dd>
               <dt>
--- a/webrtc-stats.html
+++ b/webrtc-stats.html
@@ -1952,6 +1952,7 @@ enum RTCStatsType {
              DOMString            mediaSourceId;
              DOMString            remoteId;
              DOMString            rid;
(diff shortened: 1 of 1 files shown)


webrtc-stats.html
```diff
@@ -2029,6 +2030,23 @@ <h2>
...
+                <p>
+                  MUST NOT [= map/exist =] for audio.
+                  Only [= map/exist =]s if this <a>RTP stream</a> belongs to an
+                  RTP sender that has been configured for simulcast, in other
+                  words the {{RTCRtpSender}} has multiple
+                  {{RTCRtpSendParameters/encodings}} and
+                  {{RTCOutboundRtpStreamStats}} objects. The {{encodingIndex}}
+                  is the index of the encoding that represents this <a>RTP
+                  stream</a>. MUST NOT be [= map/exist | present =] for
+                  singlecast senders.
```
jan-ivar, line 2047: Maybe add zero for simulcast?

jan-ivar (approved):

webrtc-stats.html
henbos: 0 for singlecast*, yes will do

henbos: Updated to say that this is the index without any talk about simulcast or singlecast, the implication is that this always has a value regardless

PR #813, opened 2025-08-12
rfc8888 feedback: use [= map/exist =]s instead of reported

fippo:
"reported" is a bit more vague and refers "stats being reported" and values being reported in RTCP.
For this we want exist I think.

The ECT1/CE can be measured on inbound-rtp without negotiating the RTCP so should not have that tag.


fippo:
@alvestrand ^

PR #824, opened 2026-09-22
Delete totalEncodedBytesTarget

### LINKED ISSUE #653: remove outbound-rtp totalEncodedBytesTarget
https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-totalencodedbytestarget
is implemented by libwebrtc but returns 0. So far nobody has complained (but me but I complain about *everything*)

This doesn't offer much value over comparing targetBitrate to 8*bytesSent/second
The difference is RTP packetization overhead which is a few bytes per RTP packet (or frame) which is depending on the codec and not modifiable in any way.

Additionally it poses some extra problems for encoded-transform. Since that operates on frames, adding encryption bytes to it should be accounted for in the totalEncodedBytesTarget. Which is a major headache given how this is currently implemented in libwebrtc at least.

Please vote with 👎 to remove or 👍 to keep

henbos:
Fixes #653.

As per issue description, this metric was implemented in 2019, broke in 2020 without anyone noticing, and ever since it has just returned 0 meaning there should be no backwards compat risk with just deleting and unshipping


henbos:
As suggested by @fippo

fippo (approved):

jan-ivar:
Firefox [supports](https://jsfiddle.net/jib1/uzneds0p/12/) (hit Go! and scroll down) the [totalEncodedBytesTarget stat](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-totalencodedbytestarget). Looks like we added it [4 years ago](https://bugzilla.mozilla.org/show_bug.cgi?id=1767170).

It seems to be producing ever-increasing numbers as one would expect, which seems to conflict with the statement that libwebrtc returns 0.

Will check with our team on this.

fippo:
Firefox is reading the wrong C++ field for this, the one Chrome moved away from. On [simulcast](https://fippo.github.io/simulcast-playground/rid-as-mid) Firefox shows identical values for the three streams (but is missing the per-stream targetbitrate?):
```
s = await pc1.getStats();
[...s.values()].filter(r => r.type === 'outbound-rtp').map(r => r.totalEncodedBytesTarget)
```
So 0 or wrong and nobody complained.

jan-ivar (approved):

guidou (approved):

jan-ivar:
@youennf SGTY?

henbos:
shiiiip it

